PR #2, opened 2026-01-29
Start localstack pro container

carole-lavillonniere:
Implements the `lstk start` command that pulls and runs the `localstack/localstack-pro` container using Docker.

- Adds a `Runtime` interface to abstract container operations, enabling future support for Podman, Kubernetes, etc.
- Implements `DockerRuntime` using the official Docker SDK
- Uses `docker/go-connections/nat` for port mapping helpers (standard way to define port bindings with the Docker SDK)
- Basic progress logging during image pull

## Known Limitations

This is a minimal first implementation:

- **Progress logging is noisy**: Each layer update prints a new line. Ideally each update should replace the previous one (will be addressed with bubbletea).
- **No container cleanup**: Fails if container already exists:
  ```
  Error: failed to start localstack-main: Error response from daemon: Conflict.
  The container name "/localstack-main" is already in use by container "05f923...".
  You have to remove (or rename) that container to be able to reuse that name.
  ```
- **No auth token handling**: Container fails to start without `LOCALSTACK_AUTH_TOKEN`, but no error is logged and CLI exits 0 (health check not implemented yet)
- **No integration tests yet**: Coming in a subsequent PR


cmd/start.go
```diff
@@ -11,10 +14,63 @@ var startCmd = &cobra.Command{
...
+		}
+
+		if !running {
+			return fmt.Errorf("container %s failed to start", config.Name)
+		}
+
+		fmt.Printf("%s running (container: %s)\n", config.Name, containerID[:12])
+	}
+
+	return nil
```
silv-io, line 75: nit: I would put this section in the internals as well, or in a UI package. Then we can improve it to use [bubbletea components](https://github.com/charmbracelet/bubbletea/blob/main/examples/progress-animated/main.go) for the progress indicator. But we can also do this when we get there.

internal/runtime/runtime.go
```diff
@@ -0,0 +1,23 @@
...
+
+type PullProgress struct {
+	LayerID string
+	Status  string
+	Current int64
+	Total   int64
+}
+
+// Runtime abstracts container runtime operations (Docker, Podman, Kubernetes, etc.)
+type Runtime interface {
```
silv-io, line 19: q: I guess we will extend this in the future with things like StreamLogs, Stop, etc.?

internal/runtime/docker.go
```diff
@@ -0,0 +1,103 @@
...
+	resp, err := d.client.ContainerCreate(ctx,
+		&container.Config{
+			Image:        config.Image,
+			ExposedPorts: exposedPorts,
+		},
+		&container.HostConfig{
+			PortBindings: portBindings,
+		},
+		nil, nil, config.Name,
+	)
```
silv-io, line 85: What's missing here besides the auth token are the other envvars as well. But we can focus on that when having the config set up.

silv-io (approved):
LGTM!

internal/runtime/runtime.go
carole-lavillonniere: Yes, my idea was to take it step by step with small, easy-to-review PRs without unused code.

PR #3, opened 2026-01-29
Better context handling

carole-lavillonniere:
This allows gracefully cancel of ongoing runtime (docker) operations instead of killing the process abruptly.

- Create root context with `signal.NotifyContext`. Cancel when signals `SIGINT/Ctrl+C` and `SIGTERM` are received.

- Pass context to cobra and to commands and runtime operations.

towards DRG-352

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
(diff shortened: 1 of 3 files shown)


silv-io (approved):
LGTM!

PR #4, opened 2026-01-29
Integration test for start command

carole-lavillonniere:
Adds integration tests for the `lstk start` command, written in go.

Considered [bats](https://github.com/bats-core/bats-core) to write the integration tests suite, but:
- bats is not type-safe
- bats it more complicated to write components with, to avoid code redundancy
- bats is harder to debug
- bats would bring more complexity by adding a 2nd language in the project


silv-io (approved):
LGTM! Glad we're not using `bats` here :P

PR #5, opened 2026-01-30
Handle container startup failures

carole-lavillonniere:
- Retrieve container logs when startup fails, and output the actual error instead of generic "container exited unexpectedly". We should make it look prettier in the future with bubbletea. For now we just output the raw docker logs.
- Moved business logic from `cmd/` to `internal/container/` 
- Transformed the one integration test to a failure case since auth token handling is not yet implemented. It will come together with a success test case in a subsequent PR.

### How to test
- `make test-integration` passes
- `./bin/lstk start` shows license error

Sample output:

 ```
❯ make build && ./bin/lstk
go build -o bin/lstk .
Pulling localstack/localstack-pro:latest...
  latest: Pulling from localstack/localstack-pro
  : Digest: sha256:47b0373bc10996b5333e5e116dd5d788f58d6035c75fb73376833abe62d2b16a
  : Status: Image is up to date for localstack/localstack-pro:latest
Starting localstack-aws...
Waiting for localstack-aws to be ready...
Error: localstack-aws exited unexpectedly:

 LocalStack version: 4.13.1.dev3
"LocalStack build date: 2026-01-30
%LocalStack build git hash: 6cb59848d

1Localstack returning with exit code 55. Reason:
0===============================================
#License activation failed! 🔑❌

�Reason: No credentials were found in the environment. Please make sure to either set the LOCALSTACK_AUTH_TOKEN variable to a valid auth token. If you are using the CLI, you can also run `localstack auth set-token`.

gDue to this error, Localstack has quit. LocalStack pro features can only be used with a valid license.

_- Please check that your credentials are set up correctly and that you have an active license.
O  You can find your credentials in our webapp at https://app.localstack.cloud.
^- If you want to continue using LocalStack without pro features you can set `ACTIVATE_PRO=0`.
```

toward DRG-352

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,4 +10,4 @@ clean:
 	rm -rf $(BUILD_DIR)
 
 test-integration: build
-	cd test/integration && go test -v .
+	cd test/integration && go test -count=1 -v .
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/localstack/lstk/internal/container"
+	"github.com/localstack/lstk/internal/runtime"
 	"github.com/spf13/cobra"
 )
 
@@ -13,7 +15,17 @@ var rootCmd = &cobra.Command{
 	Short: "LocalStack CLI",
 	Long:  "lstk is the command-line interface for LocalStack.",
 	Run: func(cmd *cobra.Command, args []string) {
-		if err := runStart(cmd.Context()); err != nil {
+		rt, err := runtime.NewDockerRuntime()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			os.Exit(1)
+		}
+
+		onProgress := func(msg string) {
+			fmt.Println(msg)
+		}
+
+		if err := container.Start(cmd.Context(), rt, onProgress); err != nil {
 			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
 			os.Exit(1)
 		}
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -3,9 +3,10 @@ package runtime
 import "context"
 
 type ContainerConfig struct {
-	Image string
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: internal/runtime/runtime.go [hunk 1]
```diff
@@ -3,9 +3,10 @@ package runtime
 import "context"
 
 type ContainerConfig struct {
-	Image string
-	Name  string
-	Ports map[string]string
+	Image      string
+	Name       string
+	Port       string
+	HealthPath string
```
carole-lavillonniere, line 9: We don't need to configure multiple ports yet, so simplified and made `Port` a `string` instead of a `map` to keep things simple

Makefile
```diff
@@ -10,4 +10,4 @@ clean:
 	rm -rf $(BUILD_DIR)
 
 test-integration: build
-	cd test/integration && go test -v .
+	cd test/integration && go test -count=1 -v .
```
carole-lavillonniere, line 13: This disables test caching.

```
 ❯ go help testflag | grep "disable test caching" -C 2
When 'go test' runs in package list mode, 'go test' caches successful
package test results to avoid unnecessary repeated running of tests. To
disable test caching, use any test flag or argument other than the
cacheable flags. The idiomatic way to disable test caching explicitly
is to use -count=1.
```

cmd/root.go
```diff
@@ -13,7 +15,17 @@ var rootCmd = &cobra.Command{
...
-		if err := runStart(cmd.Context()); err != nil {
+		rt, err := runtime.NewDockerRuntime()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+			os.Exit(1)
+		}
+
+		onProgress := func(msg string) {
+			fmt.Println(msg)
+		}
```
carole-lavillonniere, line 26: The onProgress callback decouples the business logic from how output is displayed.  Without it, the `internal/container` package would need to import and depend on UI (bubbletea).

silv-io (approved):
LGTM!

PR #7, opened 2026-01-30
Read auth token from env vars

carole-lavillonniere:
Read `LOCALSTACK_AUTH_TOKEN` from env vars and pass value to localstack-pro container.
 
We now have 3 test cases:

- Test: TestStartCommandSucceedsWithValidToken
Condition: Valid token
Expected: Container running

- Test: TestStartCommandFailsWithoutToken
Condition: No token
Expected: error "LOCALSTACK_AUTH_TOKEN is not set"

- Test: TestStartCommandFailsWithInvalidToken
Condition: Invalid token
Expected: error"License activation failed"

## Testing
`LOCALSTACK_AUTH_TOKEN=ls-xxx make test-integration`
or 
`cp env.example .env`, set token and `make test-integration`


silv-io (approved):
LGTM!

PR #9, opened 2026-02-02
Browser authentication flow

carole-lavillonniere:
### Changes
- Implements browser-based login when no auth token is saved or set in env var (CLI opens browser to LocalStack webapp, receives token via local callback server)
- Starts an HTTP server in order to receive the callback from the webapp (the call will be implemented in the webapp in a subsequent PR)
- Falls back gracefully by printing the URL if browser cannot be opened
**Dependencies**
- Added `github.com/pkg/browser` for cross-platform browser opening

### Integration test
- `TestStartCommandTriggersLoginWithoutToken`: simulates browser callback to verify that the login flow works  (`LOCALSTACK_WEB_APP_URL` env var allows overriding webapp URL for testing)

### TODO
- handle `/redirect?name=CLI` in webapp and execute CLI callback
- save the received auth token

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -1,16 +1,25 @@
 package auth
 
 import (
-	"errors"
+	"context"
+	"log"
 	"os"
 )
 
-// GetToken returns the auth token from keyring or environment variable.
-func GetToken() (string, error) {
-	// TODO: try keyring first
+// GetToken tries in order: 1) keychain 2) LOCALSTACK_AUTH_TOKEN env var 3) browser login
+func GetToken(ctx context.Context) (string, error) {
+	// TODO: try keychain first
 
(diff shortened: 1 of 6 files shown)


internal/auth/auth.go
```diff
@@ -1,16 +1,25 @@
...
 	}
-	return "", errors.New("auth token not found, please set LOCALSTACK_AUTH_TOKEN")
+
+	log.Println("Authentication required. Opening browser...")
+	token, err := Login(ctx)
+	if err != nil {
+		log.Println("Authentication failed.")
+		return "", err
+	}
+	log.Println("Login successful.")
```
carole-lavillonniere, line 23: Print directly for now but we will need to agree on a communication mechanism between business logic and TUI (callback functions/bubbletea messages pattern/channels etc)

silv-io (approved):
Nice, looks very clean

PR #10, opened 2026-02-03
Store token in keychain

carole-lavillonniere:
- Store auth token in system keyring using go-keyring
- Refactor auth package to use dependency injection via `Auth` struct for better testability
- Add https://github.com/uber-go/mock to generate mocks for unit testing
- Add unit test for keyring failure case (cannot be easily tested with integration test)
- Add integration tests verifying token storage and retrieval from keyring


internal/auth/login.go
```diff
@@ -19,12 +32,16 @@ func getWebAppURL() string {
...
+func login(ctx context.Context) (string, error) {
 	listener, err := net.Listen("tcp", "127.0.0.1:45678")
 	if err != nil {
 		return "", fmt.Errorf("failed to start callback server: %w", err)
 	}
-	defer listener.Close()
+	defer func() {
+		if err := listener.Close(); err != nil {
+			log.Printf("failed to close listener: %v", err)
+		}
```
carole-lavillonniere, line 30: This is to make golangcilint happy (we need to check for returned error)

internal/auth/login.go
```diff
@@ -41,8 +58,16 @@ func Login(ctx context.Context) (string, error) {
...
+	go func() {
+		if err := server.Serve(listener); err != nil && err != http.ErrServerClosed {
+			errCh <- fmt.Errorf("callback server error: %w", err)
+		}
+	}()
+	defer func() {
+		if err := server.Shutdown(ctx); err != nil {
+			log.Printf("failed to shutdown server: %v", err)
+		}
+	}()
```
carole-lavillonniere, line 57: Same here, this is for golangcilint

test/integration/start_test.go
```diff
@@ -85,6 +121,7 @@ func TestStartCommandFailsWithInvalidToken(t *testing.T) {
 
 func cleanup() {
 	ctx := context.Background()
-	dockerClient.ContainerStop(ctx, containerName, container.StopOptions{})
-	dockerClient.ContainerRemove(ctx, containerName, container.RemoveOptions{Force: true})
+	_ = dockerClient.ContainerStop(ctx, containerName, container.StopOptions{})
+	_ = dockerClient.ContainerRemove(ctx, containerName, container.RemoveOptions{Force: true})
+	_ = keyring.Delete(keyringService, keyringUser)
```
carole-lavillonniere, line 126: This is also to remove golangcilint errors

internal/auth/login.go
```diff
@@ -19,12 +31,16 @@ func getWebAppURL() string {
 	return "https://app.localstack.cloud"
 }
 
-func Login(ctx context.Context) (string, error) {
+func login(ctx context.Context) (string, error) {
```
silv-io: q: why not have this directly as the method of browserLogin?

go.mod
```diff
@@ -7,19 +7,26 @@ require (
 	github.com/docker/go-connections v0.5.0
 	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c
 	github.com/spf13/cobra v1.10.2
+	github.com/stretchr/testify v1.11.1
```
silv-io, line 11: q: is there really no way to not have test dependencies in the main dependencies in go :(

internal/auth/auth.go
```diff
@@ -6,20 +6,39 @@ import (
...
+		return token, nil
+	}
 
 	if token := os.Getenv("LOCALSTACK_AUTH_TOKEN"); token != "" {
 		return token, nil
 	}
 
 	log.Println("Authentication required. Opening browser...")
-	token, err := Login(ctx)
+	token, err := a.login.Login(ctx)
```
silv-io: nit: `a.login.Login` looks a bit funny, but it's alright :)

silv-io (approved):
LGTM! just some minor questions and nitpicks

go.mod
carole-lavillonniere: Not that I know of.
The testing modules (or any code not reachable from main)  won't be included in the final binary though, but I guess you know that.

internal/auth/auth.go
carole-lavillonniere: Renamed

internal/auth/login.go
carole-lavillonniere: Much better 👍

go.mod
silv-io: Ah yes, actually didn't think about that :D

PR #12, opened 2026-02-03
Add golangci-lint locally and in CI

silv-io:
Adds:
* golangci-lint in default settings to the CI
* make target to run it (and to check that the same version is used as in CI to avoid confusion)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,4 +26,4 @@ jobs:
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v7
         with:
-          version: v2.8.0
+          version-file: .tool-versions
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,29 @@
+name: CI
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,29 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
```
carole-lavillonniere, line 6: question: why only on `main`?

silv-io: This is for pushes, so when there is a commit that triggers it. 

For PRs there is the separate `pull_request` trigger below.

We don't really have any branch besides `main` and pull request branches.

carole-lavillonniere: You're right.
I thought that `pull_request` was only upon PR open, but I checked the doc and it says:
> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened.

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request
So all good :+1:

carole-lavillonniere (approved):

PR #16, opened 2026-02-04
Logout command

carole-lavillonniere:
- Add `lstk logout` command to remove stored auth token from system keyring
- Logout succeeds silently if no token stored
- Add integration tests for logout

closes DRG-451


silv-io (approved):
LGTM, as mentioned it could be good to have a general `reset` command for caches, logins, configs etc.

PR #17, opened 2026-02-05
Fix binary path in integration tests for windows

silv-io:
- Change to build a file ending with .exe for windows
- Add a `binaryPath` helper to the integration tests which returns a different path for windows.

This should fix the config test on windows.

Also:
- Adapt the Makefile so that `make integration` depends on the build artifact itself, not the build target

What's missing still is that `docker` apparently is available on windows but it doesn't work properly because it can't pull the image? Should be a follow-up fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,15 +82,7 @@ jobs:
             go.sum
             test/integration/go.sum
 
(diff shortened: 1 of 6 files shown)


carole-lavillonniere (approved):
👏

PR #18, opened 2026-02-05
Add device flow authentication

carole-lavillonniere:
This PR adds a device code flow as a fallback option when the browser-based authentication doesn't work.

When running `lstk start`, the CLI now supports two authentication paths:
1. Browser Flow (primary, introduced here https://github.com/localstack/lstk/pull/9): Opens browser and listens for callback on localhost:45678
2. Device Flow (fallback, new in this PR): User presses ENTER to manually complete authentication using a verification code

## User Experience

When running `lstk`:
```
Browser didn't open? Open https://app.localstack.cloud/auth/request/abc123 to authorize device.
Verification code: XYZ789
Waiting for authentication... (Press ENTER when complete)
```

If user presses ENTER after confirming in browser:
```
Checking if auth request is confirmed...
Auth request confirmed, exchanging for token...
Fetching license token...
Login successful!
```

## Mock Server Approach

This PR introduces a platform API client.
The device flow tests use `httptest.NewServer` to create a real HTTP server that mimics the platform API. 
The production code remains unchanged, we just point it to a different URL via `LOCALSTACK_PLATFORM_URL`.
This allows `TestDeviceFlowSuccess` to verify the complete end-to-end flow including container startup with a valid token.


internal/auth/login.go
```diff
@@ -20,18 +22,15 @@ type LoginProvider interface {
...
+	// Browser flow
 	listener, err := net.Listen("tcp", "127.0.0.1:45678")
 	if err != nil {
 		return "", fmt.Errorf("failed to start callback server: %w", err)
 	}
-	defer func() {
-		if err := listener.Close(); err != nil {
-			log.Printf("failed to close listener: %v", err)
-		}
-	}()
```
carole-lavillonniere, line 31: ` server.Shutdown()` closes the listener, so no explicit close needed

internal/auth/login.go
```diff
@@ -44,37 +51,93 @@ func (browserLogin) Login(ctx context.Context) (string, error) {
...
+		return b.completeDeviceFlow(ctx, authReq)
 	case <-ctx.Done():
 		return "", ctx.Err()
 	}
 }
 
-func getWebAppURL() string {
-	// allows overriding the URL for testing
-	if url := os.Getenv("LOCALSTACK_WEB_APP_URL"); url != "" {
-		return url
```
carole-lavillonniere, line 77: Not sure why I added this in the first place, but we never overwrite it at the moment, so I removed it.

test/integration/start_test.go
```diff
@@ -55,44 +42,6 @@ func TestStartCommandSucceedsWithValidToken(t *testing.T) {
 	assert.True(t, inspect.State.Running, "container should be running")
 }
 
-func TestStartCommandTriggersLoginWithoutToken(t *testing.T) {
```
carole-lavillonniere, line 58: This test was just moved to `test/integration/login_browser_flow_test.go`

internal/api/client.go
```diff
@@ -0,0 +1,175 @@
...
+	Token string `json:"token"`
+}
+
+type PlatformClient struct {
+	baseURL    string
+	httpClient *http.Client
+}
+
+func NewPlatformClient() *PlatformClient {
+	baseURL := os.Getenv("LOCALSTACK_PLATFORM_URL")
```
silv-io: nit: We already have this in the original CLI as `API_ENDPOINT` (which is set to `https://api.localstack.cloud/v1`). Maybe we can keep this with that name here for consistency?

internal/auth/login.go
```diff
@@ -3,32 +3,39 @@ package auth
...
+
+func newBrowserLogin() *browserLogin {
+	return &browserLogin{
+		platformClient: api.NewPlatformClient(),
+	}
+}
 
-func (browserLogin) Login(ctx context.Context) (string, error) {
+func startCallbackServer() (*http.Server, chan string, chan error, error) {
 	listener, err := net.Listen("tcp", "127.0.0.1:45678")
```
silv-io: nit: I know it was in already before, but maybe we can have this address defined as a constant somewhere?

internal/auth/login.go
```diff
@@ -44,37 +51,93 @@ func (browserLogin) Login(ctx context.Context) (string, error) {
...
+	}
+	fmt.Printf("Verification code: %s\n", authReq.Code)
+	fmt.Println("Waiting for authentication... (Press ENTER when complete)")
+
+	// Listen for ENTER key in background
+	go func() {
+		reader := bufio.NewReader(os.Stdin)
+		_, _ = reader.ReadString('\n')
+		enterCh <- struct{}{}
+	}()
```
silv-io, line 105: note: here we'll have to check as well how to integrate it well with UI components

internal/auth/login.go
silv-io: Theoretically we could use it to test against staging or locally started webapp. 

I wonder though if we should maybe look into handling these envvars with the config package? I'll refactor it slightly later maybe

test/integration/login_browser_flow_test.go
```diff
@@ -0,0 +1,56 @@
...
+	go func() {
+		out, _ := cmd.CombinedOutput()
+		output <- out
+	}()
+
+	// Wait for callback server to be ready
+	time.Sleep(1 * time.Second)
+
+	// Simulate browser callback with mock token
+	resp, err := http.Get("http://127.0.0.1:45678/auth/success?token=mock-token")
```
silv-io, line 42: q: what happens if we use `https://localhost.localstack.cloud` instead of `http://127.0.0.1:45678`? could we still handle it then? I imagine in the browser we'd prefer to send to https

silv-io (commented):
Just some minor nits and questions :) But looks nice ✅

test/integration/login_browser_flow_test.go
carole-lavillonniere: The callback is local (browser → 127.0.0.1) and never goes on the public internet so no TLS should be fine. Also the `localhost.localstack.cloud` DNS is not yet set up during login, only when the container starts and it would route to the container, not the CLI.

internal/api/client.go
carole-lavillonniere: Makes sense! Renamed to `LOCALSTACK_API_ENDPOINT`

internal/auth/login.go
carole-lavillonniere: 👍  Added a constant

internal/auth/login.go
carole-lavillonniere: 👍 Makes sense. Put it back in.
I agree it would be good to have a central place to read env vars and not do it all over the codebase.
I used this in the past https://github.com/caarlos0/env

test/integration/login_browser_flow_test.go
silv-io: `localhost.localstack.cloud` is always set up to redirect to `127.0.0.1`, that's how we've registered that domain. We only need the container for the subdomains afaik


I was just wondering if the browser would give any warnings if we didn't use https, but since it's local it should be fine, yes :)

carole-lavillonniere: Oh nice, I did not know it was a public DNS. So it should work as well :shrug:

silv-io (approved):
LGTM, but looks like the build is failing now :)

PR #19, opened 2026-02-06
Replace zalando/go-keyring with 99designs/keyring

carole-lavillonniere:
Switches to `99designs/keyring` for  automatic fallback to encrypted file storage when system keyring is unavailable. 
The previous implementation using `zalando/go-keyring` would fail without a keyring service. This would have required a custom fallback implementation (more code to maintain).
`99designs/keyring` has low maintenance but is stable (keyring APIs rarely change).

##  Fallback Strategy

```go
// Try system keyrings first (most secure)
ring, err := keyring.Open(config)
if err != nil {
    // Fall back to encrypted file storage
    config.AllowedBackends = []keyring.BackendType{keyring.FileBackend}
    ring, err = keyring.Open(config)
}
```

1. First attempt tries default backends (system keyrings preferred)
2. If system keyring fails or is broken, retry with FileBackend -> handles both "not available" and "partially broken" scenarios

## Storage Locations

Using `os.UserConfigDir()` for OS-dependant paths:
- **Linux**: `~/.config/localstack/`
- **macOS**: `~/Library/Application Support/localstack/`
- **Windows**: `%APPDATA%\localstack\`


internal/auth/keyring.go
```diff
@@ -15,16 +22,62 @@ type Keyring interface {
...
+
+func newSystemKeyring() (*systemKeyring, error) {
+	configDir, err := os.UserConfigDir()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get config directory: %w", err)
+	}
+
+	config := keyring.Config{
+		ServiceName: keyringService,
+		FileDir:     filepath.Join(configDir, "localstack"),
```
silv-io: can use `configDir` from `internal/config.go`
(guess will need to rename to ConfigDir)

test/integration/main_test.go
```diff
@@ -27,6 +31,22 @@ func TestMain(m *testing.M) {
 		_, err = dockerClient.Ping(context.Background())
 		dockerAvailable = err == nil
 	}
+
+	configDir, _ := os.UserConfigDir()
+	config := keyring.Config{
+		ServiceName: "localstack",
+		FileDir:     filepath.Join(configDir, "localstack"),
```
silv-io: same here with the config path

test/integration/main_test.go
```diff
@@ -46,3 +66,29 @@ func envWithoutAuthToken() []string {
...
+}
+
+func keyringDelete(service, user string) error {
+	key := fmt.Sprintf("%s/%s", service, user)
+	err := testKeyring.Remove(key)
+	if err == keyring.ErrKeyNotFound || os.IsNotExist(err) {
+		return nil
+	}
+	return err
+}
```
silv-io, line 104: nit: with these it was a bit confusing to me in the other locations why it's suddenly `keyringGet` vs `keyring.Get`. Maybe we can call it `testKeyringGet` etc.?

silv-io (commented):
Looks good to me in general, just some minor comments.

Nice that we can use something with integrated file fallback 🚀

internal/auth/keyring.go
carole-lavillonniere: :ok_hand: addressed

test/integration/main_test.go
carole-lavillonniere: I added another `configDir()` func just for integration tests, to not reuse production code in the integration tests suite.

test/integration/main_test.go
carole-lavillonniere: I don't fully get the concern here. 
We now have to wrap the calls to `keyring.Get` etc because the API changed in the new sdk, and we cannot use the sdk methods directly in the test cases.
`testKeyringGet` sounds confusing to me because it would imply that this function is testing something, or that it's a mock keyring implementation, but both are not true.

silv-io: it's alright, 
I think my confusion mainly came because of how GitHub presents PRs so I didn't see the declarations in test_main first :) 

We can keep it like that 👍🏼

silv-io (approved):

PR #20, opened 2026-02-10
Lint test files

carole-lavillonniere:
- Lint test files as well as part of the `lint` makefile target
- Fix all linting errors detected


silv-io (approved):
makes sense to lint those as well :)

PR #21, opened 2026-02-11
Add stop command

gtsiolis:
Fix DES-108

Screenshot 2026-02-11 at 17 43 13 (2)

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -50,6 +50,14 @@ func Start(ctx context.Context, rt runtime.Runtime, onProgress func(string)) err
 	}
 
 	for _, config := range containers {
+		if running, _ := rt.IsRunning(ctx, config.Name); running {
+			onProgress(fmt.Sprintf("LocalStack is already running"))
+			return nil
+		}
+
+		// Remove any existing stopped container with the same name
+		_ = rt.Remove(ctx, config.Name)
+
 		onProgress(fmt.Sprintf("Pulling %s...", config.Image))
 		progress := make(chan runtime.PullProgress)
 		go func() {
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/docker/docker/errdefs"
 	"github.com/localstack/lstk/internal/auth"
 	"github.com/localstack/lstk/internal/config"
 	"github.com/localstack/lstk/internal/runtime"
@@ -56,7 +57,9 @@ func Start(ctx context.Context, rt runtime.Runtime, onProgress func(string)) err
 		}
 
 		// Remove any existing stopped container with the same name
-		_ = rt.Remove(ctx, config.Name)
+		if err := rt.Remove(ctx, config.Name); err != nil && !errdefs.IsNotFound(err) {
+			return fmt.Errorf("failed to remove existing container %s: %w", config.Name, err)
+		}
 
 		onProgress(fmt.Sprintf("Pulling %s...", config.Image))
 		progress := make(chan runtime.PullProgress)
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -7,7 +7,7 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/docker/docker/errdefs"
+	"github.com/containerd/errdefs"
 	"github.com/localstack/lstk/internal/auth"
 	"github.com/localstack/lstk/internal/config"
 	"github.com/localstack/lstk/internal/runtime"
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -51,11 +51,6 @@ func Start(ctx context.Context, rt runtime.Runtime, onProgress func(string)) err
 	}
 
 	for _, config := range containers {
-		if running, _ := rt.IsRunning(ctx, config.Name); running {
-			onProgress(fmt.Sprintf("LocalStack is already running"))
-			return nil
-		}
-
 		// Remove any existing stopped container with the same name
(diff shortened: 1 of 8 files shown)


internal/container/start.go
```diff
@@ -50,6 +50,14 @@ func Start(ctx context.Context, rt runtime.Runtime, onProgress func(string)) err
 	}
 
 	for _, config := range containers {
+		if running, _ := rt.IsRunning(ctx, config.Name); running {
+			onProgress(fmt.Sprintf("LocalStack is already running"))
+			return nil
+		}
+
+		// Remove any existing stopped container with the same name
+		_ = rt.Remove(ctx, config.Name)
```
carole-lavillonniere: issue: It would be better to check the returned error in production code (It's ok to ignore them in a test file).
I assume the docker sdk returns a specific error for "container not running" which we can check for and ignore here, but all other errors should not be ignored.

test/integration/start_test.go
```diff
@@ -67,6 +67,33 @@ func TestStartCommandSucceedsWithKeyringToken(t *testing.T) {
...
+	cleanup()
+	t.Cleanup(cleanup)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	cmd := exec.CommandContext(ctx, binaryPath(), "start")
+	cmd.Env = append(os.Environ(), "LOCALSTACK_AUTH_TOKEN="+authToken)
+	output, err := cmd.CombinedOutput()
+	require.NoError(t, err, "first lstk start failed: %s", output)
```
carole-lavillonniere: Suggestion: can we use the docker client to run the container the first time, instead of the start command?
This way we'd be testing "what happens when container is already running", not "what happens when running lstk twice", and we would isolate the specific behavior we want to test and not run twice authentication, license validation etc (irrelevant in this scenario).
(but maybe this is too much effort, I am not sure about this one)

internal/container/stop.go
```diff
@@ -0,0 +1,43 @@
...
+type StopError struct {
+	Name string
+	Err  error
+}
+
+func (e *StopError) Error() string {
+	msg := e.Err.Error()
+
+	if strings.Contains(msg, "No such container") || strings.Contains(msg, "not found") {
+		return fmt.Sprintf("%s is not running", e.Name)
```
carole-lavillonniere: suggestion: It seems like the docker sdk is defining functions that can be used to check the error instead of doing string comparison: https://pkg.go.dev/github.com/containerd/errdefs#IsNotFound

internal/container/start.go
```diff
@@ -50,6 +50,14 @@ func Start(ctx context.Context, rt runtime.Runtime, onProgress func(string)) err
...
+
```
carole-lavillonniere, line 58: nit: I would have implemented "start succeeds when already running" in another PR and have this one only implement the stop command, in order to keep PRs small and doing only one thing.

test/integration/stop_test.go
```diff
@@ -0,0 +1,87 @@
...
+	require.NotEmpty(t, authToken, "LOCALSTACK_AUTH_TOKEN must be set to run this test")
+
+	cleanup()
+	t.Cleanup(cleanup)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	defer cancel()
+
+	startCmd := exec.CommandContext(ctx, binaryPath(), "start")
+	startCmd.Env = append(os.Environ(), "LOCALSTACK_AUTH_TOKEN="+authToken)
```
carole-lavillonniere: suggestion: same here, we could try to use the sdk instead of start command

internal/container/start.go
gtsiolis: Changed in 0408af1.

internal/container/stop.go
gtsiolis: Changed in 1f8e3eb.

internal/container/start.go
gtsiolis: Changed in 0601e91.

test/integration/stop_test.go
gtsiolis: Changed in 5f61b8c.

test/integration/start_test.go
gtsiolis: **thought**: This should be no longer relevant after removing the "already running" check from this PR, right?

gtsiolis:
Some of the checks also fail in `main`.  👀 

@carole-lavillonniere, could you take another look when you get a chance? 😁

internal/container/stop.go
carole-lavillonniere: With this change I don't think there is a need to define a custom error `StopError` anymore

carole-lavillonniere:
> Some of the checks also fail in `main`. 👀
> 
> @carole-lavillonniere, could you take another look when you get a chance? 😁

We are aware of the integration tests failing on windows/main, just created an issue: DRG-505

However there are more tests failing in this PR, namely:
```
=== RUN   TestStopCommandSucceeds
    stop_test.go:33: 
        	Error Trace:	/home/runner/work/lstk/lstk/test/integration/stop_test.go:20
        	            				/home/runner/work/lstk/lstk/test/integration/stop_test.go:33
        	Error:      	Received unexpected error:
        	            	Error response from daemon: No such image: alpine:latest
        	Test:       	TestStopCommandSucceeds
        	Messages:   	failed to create test container
--- FAIL: TestStopCommandSucceeds (0.00s)
```

internal/container/stop.go
gtsiolis: Nice catch! Changed in a55d50a.

carole-lavillonniere (approved):

PR #22, opened 2026-02-12
Validate license before starting container

carole-lavillonniere:
The CLI now validates licenses upfront via the platform API, providing clear error messages if validation fails, rather than waiting for the container to fail during startup.
This should be further improved in the future by resolving the version for the `latest` tag without having to pull the images (DRG-504, DRG-508).

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -14,6 +14,8 @@ const (
 	EmulatorAWS       EmulatorType = "aws"
 	EmulatorSnowflake EmulatorType = "snowflake"
 	EmulatorAzure     EmulatorType = "azure"
+
+	dockerRegistry = "localstack"
 )
 
 var emulatorImages = map[EmulatorType]string{
@@ -44,7 +46,7 @@ func (c *ContainerConfig) Image() (string, error) {
 	if tag == "" {
 		tag = "latest"
 	}
-	return fmt.Sprintf("localstack/%s:%s", productName, tag), nil
+	return fmt.Sprintf("%s/%s:%s", dockerRegistry, productName, tag), nil
 }
 
 // Name returns the container name: "localstack-{type}" or "localstack-{type}-{tag}" if tag != latest
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -17,7 +17,7 @@ const (
 )
 
 var emulatorImages = map[EmulatorType]string{
-	EmulatorAWS: "localstack/localstack-pro",
+	EmulatorAWS: "localstack-pro",
 }
 
 var emulatorHealthPaths = map[EmulatorType]string{
@@ -36,15 +36,15 @@ type ContainerConfig struct {
 }
 
 func (c *ContainerConfig) Image() (string, error) {
-	baseImage, ok := emulatorImages[c.Type]
-	if !ok {
-		return "", fmt.Errorf("%s emulator not supported yet by lstk", c.Type)
+	productName, err := c.ProductName()
+	if err != nil {
+		return "", err
 	}
 	tag := c.Tag
 	if tag == "" {
 		tag = "latest"
 	}
-	return fmt.Sprintf("%s:%s", baseImage, tag), nil
+	return fmt.Sprintf("localstack/%s:%s", productName, tag), nil
 }
 
 // Name returns the container name: "localstack-{type}" or "localstack-{type}-{tag}" if tag != latest
@@ -62,6 +62,14 @@ func (c *ContainerConfig) HealthPath() (string, error) {
 		return "", fmt.Errorf("%s emulator not supported yet by lstk", c.Type)
 	}
 	return path, nil
+}
+
+func (c *ContainerConfig) ProductName() (string, error) {
+	productName, ok := emulatorImages[c.Type]
+	if !ok {
+		return "", fmt.Errorf("%s emulator not supported yet by lstk", c.Type)
+	}
+	return productName, nil
 }
 
 func ConfigDir() (string, error) {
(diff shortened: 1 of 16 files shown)


internal/config/config.go
```diff
@@ -36,15 +36,15 @@ type ContainerConfig struct {
...
+	productName, err := c.ProductName()
+	if err != nil {
+		return "", err
 	}
 	tag := c.Tag
 	if tag == "" {
 		tag = "latest"
 	}
-	return fmt.Sprintf("%s:%s", baseImage, tag), nil
+	return fmt.Sprintf("localstack/%s:%s", productName, tag), nil
```
silv-io: nit: maybe lets put the "localstack" somewhere as a const named `vendor` or just `localstack` to avoid typos

internal/config/config.go
```diff
@@ -64,6 +64,14 @@ func (c *ContainerConfig) HealthPath() (string, error) {
 	return path, nil
 }
 
+func (c *ContainerConfig) ProductName() (string, error) {
```
silv-io, line 69: q: did we check that the other products are actually binding their product name to the image name? I think for snowflake it might not be just `snowflake`

internal/container/start.go
```diff
@@ -83,6 +98,50 @@ func Start(ctx context.Context, rt runtime.Runtime, onProgress func(string)) err
...
 }
 
+func validateLicense(ctx context.Context, rt runtime.Runtime, platformClient api.PlatformAPI, containerConfig runtime.ContainerConfig, cfgContainer *config.ContainerConfig, token string, onProgress func(string)) error {
+	version := cfgContainer.Tag
+	if version == "" || version == "latest" {
+		actualVersion, err := rt.GetImageVersion(ctx, containerConfig.Image)
+		if err != nil {
+			log.Printf("Could not resolve version from image %s: %v", containerConfig.Image, err)
+			if version == "" {
+				version = "latest"
```
silv-io: q: does this now work after all or is this just a placeholder until it does?

test/integration/license_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const licenseContainerName = "localstack-aws"
+
+func TestLicenseValidationSuccess(t *testing.T) {
+	requireDocker(t)
+	authToken := os.Getenv("LOCALSTACK_AUTH_TOKEN")
+	require.NotEmpty(t, authToken, "LOCALSTACK_AUTH_TOKEN must be set to run this test")
```
silv-io, line 24: nit: would swap those two lines to have all the preconditions together at the start :D

test/integration/license_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	cleanupLicense()
+	t.Cleanup(cleanupLicense)
+
+	validationErrors := make(chan error, 1)
+
+	// Mock platform API that returns success
+	mockServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
silv-io, line 32: it's verbose but I like it

test/integration/main_test.go
```diff
@@ -67,10 +69,17 @@ func requireDocker(t *testing.T) {
 	}
 }
 
-func envWithoutAuthToken() []string {
+func envWithout(keys ...string) []string {
```
silv-io, line 81: praise: nice

silv-io (commented):
Looks good, just some minor nits :)

test/integration/license_test.go
carole-lavillonniere: They cannot be swapped since line 24 relies on line 23, or did you mean something else?

silv-io: 🤦🏼 sorry, misread it

Anyways, we could use `os.LookupEnv` here instead because it doesn't swallow away the result of the syscall. Would be more accurate I guess. I'll leave it up to you

```
// Getenv retrieves the value of the environment variable named by the key.
// It returns the value, which will be empty if the variable is not present.
// To distinguish between an empty value and an unset value, use [LookupEnv].
func Getenv(key string) string {
	testlog.Getenv(key)
	v, _ := syscall.Getenv(key)
	return v
}

// LookupEnv retrieves the value of the environment variable named
// by the key. If the variable is present in the environment the
// value (which may be empty) is returned and the boolean is true.
// Otherwise the returned value will be empty and the boolean will
// be false.
func LookupEnv(key string) (string, bool) {
	testlog.Getenv(key)
	return syscall.Getenv(key)
}
```

internal/config/config.go
carole-lavillonniere: How does this look? 9e84b5dbe86bd27f8866fbaad5b003c4fbe7caca

internal/config/config.go
carole-lavillonniere: How about we do this tiny refactoring later when we need to?

internal/container/start.go
carole-lavillonniere: No that's just me committing old code :see_no_evil: 9e84b5dbe86bd27f8866fbaad5b003c4fbe7caca

test/integration/license_test.go
carole-lavillonniere: What would you suggest we do with the bool returned?

silv-io: I think with the bool we know specifically if it's not set, with getting it and checking it could either be not set or just set to empty. probably irrelevant

silv-io (approved):
LGTM! Just need to address the merge conflicts now 😬

PR #23, opened 2026-02-12
Add dependabot config to update go and dependencies

silv-io:
Go 1.26 is out. Would be nice to have an action just upgrading it for us and also our dependencies via dependabot while we're at it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    groups:
+      go-version:
+        patterns:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    groups:
+      go-version:
+        patterns:
+          - "go"
```
carole-lavillonniere: issue: as far as I know this does not upgrade the project's go version. We could use Renovate instead which supports it.

silv-io: you're right, I've changed it to a github action now

carole-lavillonniere (approved):

PR #24, opened 2026-02-13
Fix failing integration test on windows

carole-lavillonniere:
## Problem
GitHub Actions Windows runners don't support nested virtualization, which is required to run Linux containers on Windows. From [GitHub discussion](https://github.com/orgs/community/discussions/25491):

> "The Windows VMs are not enabled for nested virtualization so you won't be able to run a linux container on the Windows VM."

## Solution
- New `login` command that performs authentication without starting containers. Use this new command in authentication integration tests
- Docker-based tests are now skipped on windows on CI, while login/logout tests run successfully (Login/authentication doesn't need containers)
- Added an env var `KEYRING=file` to force saving token to file instead of keychain. Only enable when running tests on macOS. With the changes of this PR we now run the login command on macOS, but it would without the var stay stuck because of user interaction expected with the keychain pop-up.

Integration tests now run on all platforms:
- **Linux/macOS**: All tests run (login + docker tests). Keychain is disabled for macOS.
- **Windows CI**: Login tests run, docker tests skip automatically
- **Windows local**: All tests run if docker desktop is available (with linux containers)

## Additional changes
- Makefile `test-integration` target now depends on `build` instead of just the binary, in order to always use latest code
- Fixes docker pull errors on windows where "no matching manifest for windows/amd64" wasn't being caught, causing tests to fail with "No such image" errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ test:
 	@JUNIT=""; [ -n "$$CREATE_JUNIT_REPORT" ] && JUNIT="--junitfile test-results.xml"; \
 	go run gotest.tools/gotestsum@latest --format testdox $$JUNIT -- ./cmd/... ./internal/...
 
-test-integration: $(BUILD_DIR)/$(BINARY_NAME)
+test-integration: build
 	@JUNIT=""; [ -n "$$CREATE_JUNIT_REPORT" ] && JUNIT="--junitfile ../../test-integration-results.xml"; \
 	cd test/integration && go run gotest.tools/gotestsum@latest --format testdox $$JUNIT -- -count=1 ./...
 
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ test:
 	@JUNIT=""; [ -n "$$CREATE_JUNIT_REPORT" ] && JUNIT="--junitfile test-results.xml"; \
 	go run gotest.tools/gotestsum@latest --format testdox $$JUNIT -- ./cmd/... ./internal/...
 
-test-integration: build
+test-integration: $(BUILD_DIR)/$(BINARY_NAME)
 	@JUNIT=""; [ -n "$$CREATE_JUNIT_REPORT" ] && JUNIT="--junitfile ../../test-integration-results.xml"; \
 	if [ "$$(uname)" = "Darwin" ]; then \
 		cd test/integration && KEYRING=file go run gotest.tools/gotestsum@latest --format testdox $$JUNIT -- -count=1 ./...; \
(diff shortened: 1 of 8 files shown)


internal/auth/keyring.go
```diff
@@ -41,6 +41,11 @@ func newSystemKeyring() (*systemKeyring, error) {
 		},
 	}
 
+	// Force file backend if KEYRING env var is set to "file"
+	if os.Getenv("KEYRING") == "file" {
+		keyringConfig.AllowedBackends = []keyring.BackendType{keyring.FileBackend}
+	}
```
silv-io, line 47: nit: would be good to add a todo here to wire it into the config system. Ideally I think we shouldn't have os.GetEnv across the code but only read it through the config.

test/integration/login_browser_flow_test.go
```diff
@@ -12,14 +12,13 @@ import (
 )
 
 func TestBrowserFlowStoresToken(t *testing.T) {
-	requireDocker(t)
```
silv-io, line 15: praise: smart to remove that dependency there!

test/integration/main_test.go
```diff
@@ -65,6 +70,10 @@ func requireDocker(t *testing.T) {
 	if !dockerAvailable {
 		t.Skip("Docker is not available")
 	}
+	// Skip Docker tests on Windows (GitHub Actions doesn't support Linux containers)
```
silv-io, line 73: so weird that they make the `docker` command available then

Makefile
```diff
@@ -18,9 +18,13 @@ test:
 	@JUNIT=""; [ -n "$$CREATE_JUNIT_REPORT" ] && JUNIT="--junitfile test-results.xml"; \
 	go run gotest.tools/gotestsum@latest --format testdox $$JUNIT -- ./cmd/... ./internal/...
 
-test-integration: $(BUILD_DIR)/$(BINARY_NAME)
+test-integration: build
```
silv-io: nit: this would now always force a rebuild then. The reason why I've added it like this is so that if we do a `make build` and afterwards a `make test-integration` that we won't build twice.

silv-io (commented):
Looks good, just some minor nits! 🚀

internal/auth/keyring.go
carole-lavillonniere: Created a ticket instead DRG-529

test/integration/main_test.go
carole-lavillonniere: It is available but can only run windows containers.

Makefile
carole-lavillonniere: Ok, removed. I removed it because I tend to forget to rebuild which leads to wrong test results + go build is very quick, but fine :)

test/integration/main_test.go
silv-io: TIL about windows containers :)

Makefile
silv-io: you can just do `make clean test-integration` instead

silv-io (approved):
Thanks for addressing the comments! 😃

PR #25, opened 2026-02-13
Add event sink system instead of plain CLI printing

silv-io:
## Motivation

Currently, the CLI prints directly to stdout throughout the codebase. This becomes problematic when integrating with bubbletea for TUI rendering, since bubbletea owns stdout and you can't print to it while the TUI is running.

By routing all output through a `Sink` interface, we can:
- Use `PlainSink` for non-interactive CLI mode (writes to stdout)
- Swap in a future `TUISink` that sends events to bubbletea's message loop via `Program.Send()`

The event types (`LogEvent`, `ProgressEvent`, etc.) map naturally to bubbletea's `tea.Msg` types, making the future integration straightforward.

## Changes

- Added `internal/output` package with typed events (`LogEvent`, `WarningEvent`, `ContainerStatusEvent`, `ProgressEvent`) and a `Sink` interface
- Used Go generics with union type constraints for compile-time event type safety
- Replaced direct `fmt.Print`/`log.Printf` calls in `container/start.go`, `auth/auth.go`, `auth/login.go`, and command handlers with sink-based emission
- Updated all commands (`start`, `login`, `logout`) to create and pass a `PlainSink`

## Tests

- Added unit tests for `PlainSink` covering all event types and formatting variations
- Updated existing `auth_test.go` to use `SinkFunc` for capturing emitted events and asserting on them
- All existing tests pass; the sink abstraction is transparent to test behavior

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 bin/
 *.test
+
+.idea
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,6 +25,7 @@ Note: Integration tests require `LOCALSTACK_AUTH_TOKEN` environment variable for
   - `container/` - Handling different emulator containers
   - `runtime/` - Abstraction for container runtimes (Docker, Kubernetes, etc.) - currently only Docker implemented
   - `auth/` - Authentication (env var token or browser-based login)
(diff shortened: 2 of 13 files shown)


carole-lavillonniere (approved):
Lovely

PR #31, opened 2026-02-13
Fix upgrade workflow

silv-io:
The new upgrade workflow installed the current go version to upgrade the go version, which doesn't work in CI by the looks of it (works locally though).

Should've done a test run before merging. Fixed it now and confirmed with test runs that it will work in #32 and #33

### DIFF
--- a/.github/workflows/weekly-go-upgrade.yml
+++ b/.github/workflows/weekly-go-upgrade.yml
@@ -43,4 +43,3 @@ jobs:
             - Runs `go mod tidy` to keep module files consistent.
           labels: dependencies
           token: ${{ secrets.PRO_ACCESS_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly-go-upgrade.yml
```diff
@@ -39,4 +42,5 @@ jobs:
             - Updates the `go` directive to the latest released Go version.
             - Runs `go mod tidy` to keep module files consistent.
           labels: dependencies
-          delete-branch: true
+          token: ${{ secrets.PRO_ACCESS_TOKEN }}
+          reviewers: silv-io,carole-lavillonniere
```
carole-lavillonniere: nit: can we pick up the users from codeowners file instead?

carole-lavillonniere (approved):

PR #34, opened 2026-02-16
Fix: add label to keyring entry 

carole-lavillonniere:
With this change, macOS keychain users won't see an empty string in the pop-up:
image

## Other changes

- Simplified the keyring API: the generic `Keyring` interface is replaces with a purpose-specific `AuthTokenStorage` interface. This removes the need for passing redundant key parameters.
- Also simplified the API of the keyring used in integration tests.


carole-lavillonniere:
@coderabbitai review

internal/auth/token_storage.go
```diff
@@ -69,21 +72,18 @@ func (k *systemKeyring) Get(service, user string) (string, error) {
...
 
-func (k *systemKeyring) Set(service, user, password string) error {
-	return k.ring.Set(keyring.Item{
-		Key:  k.makeKey(service, user),
-		Data: []byte(password),
+func (s *authTokenStorage) SetAuthToken(token string) error {
+	return s.ring.Set(keyring.Item{
+		Key:   keyringAuthTokenKey,
+		Data:  []byte(token),
+		Label: keyringAuthTokenLabel,
```
carole-lavillonniere, line 79: The added `Label` is the fix.

silv-io (approved):
LGTM! Got nothing besides the nitpicks that the rabbit already found

PR #36, opened 2026-02-16
Remove browser flow as auth method

carole-lavillonniere:
- Remove browser flow auth method, keep device flow only
- Improve the UI for authentication

Reasons for removing browser flow:
- It feels less modern and secure to users
- It is less secure
- It's hard to know when to fallback to device flow 
- Handling 2 login methods makes the code more complicated

Output:
image


carole-lavillonniere:
@coderabbitai review

internal/auth/auth.go
```diff
@@ -42,7 +42,7 @@ func (a *Auth) GetToken(ctx context.Context) (string, error) {
 	}
 
 	output.EmitLog(a.sink, "Authentication required. Opening browser...")
```
silv-io: nit: Maybe we can rename this log now, because we're asking later if we want to open it

internal/auth/login.go
```diff
@@ -15,106 +13,69 @@ import (
...
-	mux.HandleFunc("/auth/success", func(w http.ResponseWriter, r *http.Request) {
-		token := r.URL.Query().Get("token")
-		if token == "" {
-			errCh <- fmt.Errorf("no token in callback")
-			http.Error(w, "No token received", http.StatusBadRequest)
-			return
-		}
-		w.WriteHeader(http.StatusOK)
-		tokenCh <- token
+	// Ask whether to open the browser; ENTER or Y accepts (default yes), N skips
```
silv-io, line 43: thought: we might want to change the type a bit so that we can codify the default option more clearly. But for now this is fine

internal/ui/app.go
```diff
@@ -62,15 +62,27 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			}
+			// A single character key press selects the matching option
+			for _, opt := range a.pendingInput.Options {
+				if msg.String() == opt.Key {
+					responseCmd := sendInputResponseCmd(a.pendingInput.ResponseCh, output.InputResponse{SelectedKey: opt.Key})
+					a.pendingInput = nil
+					a.inputPrompt = a.inputPrompt.Hide()
+					return a, responseCmd
+				}
 			}
```
silv-io, line 85: nice that this works so well :)

silv-io (approved):
Love a good cleanup!

Just a thought on expanding the userinputevent functionality. The output nit we can handle later. We'll freshen it up either way now that we default to the TUI. 

Also, for posterity, this is how it looks with TUI now: 
```

    ▟████▖    LocalStack (lstk)
   ▟██▙█▙█▟   dev
     ▀▛▀▛▀

  Authentication required. Opening browser...
  Visit: https://app.localstack.cloud/auth/request/...
  Verification code: 3FE3
  Checking if auth request is confirmed...
  Auth request confirmed, exchanging for token...
  Fetching license token...
  Login successful.
  Pulling localstack/localstack-pro:latest...
    latest: Pulling from localstack/localstack-pro
    : Digest: sha256:dab91774aa81fc565ff8019962749810d2184f28f4c7acbea514c9754235de9f
    : Status: Image is up to date for localstack/localstack-pro:latest
  localstack-aws: validating license (4.13.2.dev137)
  Starting localstack-aws...
  Waiting for localstack-aws to be ready...
  localstack-aws ready (containerId: c5326e33aa84)
```

internal/auth/auth.go
carole-lavillonniere: Good catch, I lost this when resolving conflicts

internal/auth/login.go
carole-lavillonniere: Agreed.

PR #37, opened 2026-02-16
Add styled CLI output with spinner for logout command

gtsiolis:
This will add TUI for the logout command.

This will ...
 - Add interactive TUI with loading spinner for the logout command
  - Introduce styled terminal output with colored prefixes for success/note/warning messages
  - Show appropriate feedback: "Success: Logged out successfully." or "Note: Not currently  logged in."

| Loading | Success | Note |
|--------|--------|--------|
| Frame 480974612 | Frame 480974614  | Frame 480974613 |

Fix DES-113

### DIFF
--- a/cmd/logout.go
+++ b/cmd/logout.go
@@ -1,12 +1,14 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
 	"github.com/localstack/lstk/internal/api"
 	"github.com/localstack/lstk/internal/auth"
 	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/ui"
 	"github.com/spf13/cobra"
 )
 
@@ -15,6 +17,10 @@ var logoutCmd = &cobra.Command{
 	Short:   "Remove stored authentication credentials",
 	PreRunE: initConfig,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		if ui.IsInteractive() {
+			return ui.RunLogout(cmd.Context())
+		}
+
 		sink := output.NewPlainSink(os.Stdout)
 		platformClient := api.NewPlatformClient()
 		tokenStorage, err := auth.NewTokenStorage()
@@ -23,9 +29,11 @@ var logoutCmd = &cobra.Command{
 		}
 		a := auth.New(sink, platformClient, tokenStorage, false)
 		if err := a.Logout(); err != nil {
+			if errors.Is(err, auth.ErrNotLoggedIn) {
+				return nil
+			}
 			return fmt.Errorf("failed to logout: %w", err)
 		}
-		fmt.Println("Logged out successfully.")
 		return nil
 	},
 }
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -5,11 +5,12 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/99designs/keyring"
 	"github.com/localstack/lstk/internal/api"
 	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/output"
 )
+
+var ErrNotLoggedIn = errors.New("not logged in")
 
 type Auth struct {
 	tokenStorage AuthTokenStorage
@@ -58,9 +59,25 @@ func (a *Auth) GetToken(ctx context.Context) (string, error) {
 
 // Logout removes the stored auth token from the keyring
 func (a *Auth) Logout() error {
-	err := a.tokenStorage.DeleteAuthToken()
-	if errors.Is(err, keyring.ErrKeyNotFound) {
-		return nil
+	output.EmitSpinnerStart(a.sink, "Logging out...")
+
+	_, err := a.tokenStorage.GetAuthToken()
+	if errors.Is(err, ErrTokenNotFound) {
+		output.EmitSpinnerStop(a.sink)
+		output.EmitNote(a.sink, "Not currently logged in")
+		return ErrNotLoggedIn
 	}
-	return err
+	if err != nil {
+		output.EmitSpinnerStop(a.sink)
+		return fmt.Errorf("failed to read auth token: %w", err)
+	}
+
+	if err := a.tokenStorage.DeleteAuthToken(); err != nil {
+		output.EmitSpinnerStop(a.sink)
+		return fmt.Errorf("failed to delete auth token: %w", err)
+	}
+
+	output.EmitSpinnerStop(a.sink)
+	output.EmitSuccess(a.sink, "Logged out successfully")
+	return nil
 }
--- a/internal/output/events.go
+++ b/internal/output/events.go
@@ -15,6 +15,8 @@
 //   - Use for errors that need more than a single line
 package output
 
+import "time"
+
 type MessageSeverity int
 
 const (
@@ -30,8 +32,9 @@ type MessageEvent struct {
 }
 
 type SpinnerEvent struct {
-	Active bool
-	Text   string
+	Active      bool
+	Text        string
+	MinDuration time.Duration // Minimum time spinner should display (0 = use default)
 }
 
 type ErrorAction struct {
@@ -144,8 +147,16 @@ func EmitContainerLogLine(sink Sink, line string) {
 	Emit(sink, ContainerLogLineEvent{Line: line})
 }
 
+const DefaultSpinnerMinDuration = 400 * time.Millisecond
+
+// EmitSpinnerStart starts spinner with default min duration (400ms)
 func EmitSpinnerStart(sink Sink, text string) {
-	Emit(sink, SpinnerEvent{Active: true, Text: text})
+	Emit(sink, SpinnerEvent{Active: true, Text: text, MinDuration: DefaultSpinnerMinDuration})
+}
+
+// EmitSpinnerStartWithDuration starts spinner with custom min duration (0 = no minimum)
+func EmitSpinnerStartWithDuration(sink Sink, text string, minDuration time.Duration) {
+	Emit(sink, SpinnerEvent{Active: true, Text: text, MinDuration: minDuration})
 }
 
 func EmitSpinnerStop(sink Sink) {
--- a/internal/ui/run_logout.go
+++ b/internal/ui/run_logout.go
@@ -1,0 +1,57 @@
+package ui
+
+import (
+	"context"
+	"errors"
+	"os"
+
+	tea "github.com/charmbracelet/bubbletea"
+	"github.com/localstack/lstk/internal/api"
+	"github.com/localstack/lstk/internal/auth"
+	"github.com/localstack/lstk/internal/output"
+)
+
+func RunLogout(parentCtx context.Context) error {
+	_, cancel := context.WithCancel(parentCtx)
+	defer cancel()
+
+	app := NewApp("", cancel)
+	p := tea.NewProgram(app, tea.WithInput(os.Stdin), tea.WithOutput(os.Stdout))
+	runErrCh := make(chan error, 1)
+
+	go func() {
+		tokenStorage, err := auth.NewTokenStorage()
+		if err != nil {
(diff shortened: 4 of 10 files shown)


silv-io:
I'd prefer to add this after we have the actual TUI in there with #38, but looks good so far :)

gtsiolis:
> after we have the actual TUI in there with https://github.com/localstack/lstk/pull/38

Sounds good, putting this on hold. Please, ping me when I can get back to this. 🏓

silv-io:
@gtsiolis the TUI landed on `main` now, so we can continue with this. If you need help with the concepts we introduced there now, let me know :)

gtsiolis:
Awesome, will make another attempt after deleting everything in the branch, but will ping you if this needs help. 😭

gtsiolis:
I've rebased and re-applied the logic on top of TUI. Ready to review.

gtsiolis:
The suggested fix for the last bot review can be the following but leaving this to a human.

```diff
diff --git a/internal/output/plain_sink.go b/internal/output/plain_sink.go
index 07c9f35..2342f77 100644
--- a/internal/output/plain_sink.go
+++ b/internal/output/plain_sink.go
@@ -6,20 +6,26 @@ import (
        "os"

        "github.com/charmbracelet/lipgloss"
+       "golang.org/x/term"
 )

 var secondaryStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("241"))

 type PlainSink struct {
-       out io.Writer
-       err error
+       out   io.Writer
+       err   error
+       isTTY bool
 }

 func NewPlainSink(out io.Writer) *PlainSink {
        if out == nil {
                out = os.Stdout
        }
-       return &PlainSink{out: out}
+       isTTY := false
+       if f, ok := out.(*os.File); ok {
+               isTTY = term.IsTerminal(int(f.Fd()))
+       }
+       return &PlainSink{out: out, isTTY: isTTY}
 }

 // Err returns the first write error encountered, if any.
@@ -41,7 +47,11 @@ func (s *PlainSink) emit(event any) {

        switch event.(type) {
        case SuccessEvent, NoteEvent, WarningEvent:
-               line = secondaryStyle.Render("> ") + line
+               if s.isTTY {
+                       line = secondaryStyle.Render("> ") + line
+               } else {
+                       line = "> " + line
+               }
        }

        _, err := fmt.Fprintln(s.out, line)
```

internal/output/events.go
```diff
@@ -1,7 +1,7 @@
 package output
 
 type Event interface {
-	LogEvent | WarningEvent | ContainerStatusEvent | ProgressEvent | UserInputRequestEvent | ContainerLogLineEvent
+	LogEvent | WarningEvent | SuccessEvent | NoteEvent | ContainerStatusEvent | ProgressEvent | UserInputRequestEvent | ContainerLogLineEvent
```
silv-io: thought: I think we'll need to document a bit which event should be used for which kind of message, and what the difference is in how they're going to be displayed.

internal/output/plain_sink.go
```diff
@@ -4,8 +4,12 @@ import (
 	"fmt"
 	"io"
 	"os"
+
+	"github.com/charmbracelet/lipgloss"
```
silv-io: blocker: we shouldn't put styling in the plain sink. this one is for non-terminal outputs and it doesn't make sense to have color or anything like that here

internal/auth/auth.go
```diff
@@ -56,11 +56,20 @@ func (a *Auth) GetToken(ctx context.Context) (string, error) {
...
 	err := a.tokenStorage.DeleteAuthToken()
 	if errors.Is(err, keyring.ErrKeyNotFound) {
-		return nil
+		return LogoutResult{TokenDeleted: false}, nil
+	}
+	if err != nil {
+		return LogoutResult{}, err
 	}
-	return err
+	return LogoutResult{TokenDeleted: true}, nil
```
carole-lavillonniere: suggestion:
It would be more idiomatic go to return a sentinel error rather than a boolean here. We'd have to define an error in this file, something like 
`var ErrNotLoggedIn = errors.New("not currently logged in")`
and we could return 
`return ErrNotLoggedIn`
instead of 
`return LogoutResult{}, err`
and we would not need the struct anymore:
 `type LogoutResult struct {
	TokenDeleted bool
}`

internal/ui/run_logout.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func (a LogoutApp) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
+	switch msg := msg.(type) {
+	case tea.KeyMsg:
+		if msg.String() == "ctrl+c" || msg.String() == "q" {
+			return a, tea.Quit
+		}
+
+	case logoutDoneMsg:
```
carole-lavillonniere: suggestion:
we could have 2 different messages here `logoutDoneMsg` and something like `logoutNotLoggedInMsg`, as far as I understand that would be more idiomatic bubble tea: messages represent an event and are not meant to contain payload.

internal/ui/run_logout.go
```diff
@@ -0,0 +1,135 @@
...
+		a.spinner = a.spinner.Hide()
+		if msg.result.TokenDeleted {
+			a.result = &logoutResultDisplay{success: true, message: "Logged out successfully."}
+		} else {
+			a.result = &logoutResultDisplay{success: false, message: "Not currently logged in."}
```
carole-lavillonniere: Funnily at the moment I always get 
> `> Success: Logged out successfully.`

even when running the logout command multiple times.

internal/output/plain_sink.go
```diff
@@ -4,8 +4,12 @@ import (
...
 )
 
+var secondaryStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("241"))
```
carole-lavillonniere: issue: 
Don't we already do this in` internal/ui/app.go `line 165?
I think we can remove all changes here in file` internal/output/plain_sink.go`
We should not do any styling here and we should not import lipgloss in this file.
This would also solve the rabbit's report.
It's about separating concerns: the plain sink should to be an abstraction that does not know about rendering.

cmd/logout.go
```diff
@@ -1,30 +1,46 @@
...
+		if err != nil {
+			return err
+		}
+
+		if result.TokenDeleted {
+			output.EmitSuccess(sink, "Logged out successfully.")
+		} else {
+			output.EmitNote(sink, "Not currently logged in.")
+		}
+
```
silv-io: blocker: This should all happen inside the Logout internal package IMO. The command should just pass different sinks based on TUI or non-TUI mode.

As we can see further up you're not passing a sink to the TUI because you're handling all the output centrally here.

We need to do this distinction though to keep plain and TUI mode completely separate.

silv-io (changes requested):
The base components that we'll need are there and look reasonable IMO.

However, to have a nice separation of concerns and better switching between plain and TUI mode, we should move the output inside of the Logout command and pass a TUI sink to the model.

This way we won't have to deal with lipgloss styling in the plain sink.

gtsiolis:
Thanks both, tried to address all comments in 19a1d51 and e5943ab. 

Could you take another look when you get a chance, or pick this up from here?

cmd/logout.go
```diff
@@ -1,31 +1,40 @@
...
 		platformClient := api.NewPlatformClient()
 		tokenStorage, err := auth.NewTokenStorage()
 		if err != nil {
-			return fmt.Errorf("failed to initialize token storage: %w", err)
+			return err
+		}
+
+		if ui.IsInteractive() {
+			a := auth.New(nil, platformClient, tokenStorage, false)
+			return ui.RunLogout(cmd.Context(), version, a)
```
carole-lavillonniere: Thought: It looks a bit funny that we're passing a nil sink in interactive mode, and that we basically have to define the output in 2 different places:
1. TUI mode (`ui/run_logout.go`): strings "Logging out", "Not currently logged in.", "Logged out successfully."
2. Non-interactive mode (`cmd/logout.go`): the same strings in output.Emit*() calls

What do you think @silv-io ? Not sure how this could be improved...

Maybe in the future we can implement something like a FileSink to capture everything in a log file when running in interactive mode.

carole-lavillonniere:
> Thanks both, tried to address all comments in [19a1d51](https://github.com/localstack/lstk/commit/19a1d511f7d4cb159c7376c15ed8f2394635b143) and [e5943ab](https://github.com/localstack/lstk/commit/e5943ab414498b2cff347c5c262b50db331f5780).
> 
> Could you take another look when you get a chance, or pick this up from here?

This looks good to me! Just left one comment mostly for Silvio. Let's wait for his review.

cmd/logout.go
gtsiolis: Good point, rebased to keep fresh, and waiting for @silv-io to take a second look.

silv-io: It looks funny because it's using the sinks wrong :P 

The model defines own messages for success and failure even though we already emit those in the internal logout package. 

What we need to do here is use a TUISink and handle the actual Logout success messages in the model

internal/ui/run_logout.go
```diff
@@ -0,0 +1,150 @@
...
+type logoutSuccessMsg struct{}
+
+type logoutNotLoggedInMsg struct{}
+
+type logoutErrMsg struct {
+	err error
+}
+
+type logoutState int
+
```
silv-io: blocker: we don't need these. Instead we should use the success message sent from the logout package. What we're doing here is basically inventing state in the UI vs. using the actual state emitted by auth.Logout

internal/ui/run_logout.go
```diff
@@ -0,0 +1,150 @@
...
+		a.spinner = a.spinner.Hide()
+		a.err = msg.err
+		return a, tea.Quit
+
+	default:
+		var cmd tea.Cmd
+		a.spinner, cmd = a.spinner.Update(msg)
+		return a, cmd
+	}
+
```
silv-io: blocker: here we should handle the actual success messages emitted by auth.Logout by passing a TUI sink into it.

internal/auth/auth.go
```diff
@@ -56,11 +56,23 @@ func (a *Auth) GetToken(ctx context.Context) (string, error) {
...
-		return nil
+		output.EmitNote(a.sink, "Not currently logged in.")
+		return ErrNotLoggedIn
 	}
-	return err
+	if err != nil {
+		return err
+	}
+
+	output.EmitSuccess(a.sink, "Logged out successfully.")
```
silv-io: note: We should do these events in a way so that the TUI can handle them in it's update flow. The success looks fine here. Not sure about the other ones yet :D

silv-io (changes requested):
Looks much better already, but it circumvents the defined communication flow in the TUI and also is blocking the UI I think. what should happen is:

1. In the Run for the TUI we start the logout flow in a goroutine with a tui sink.
2. Then we start the TUI program which will receive the events from auth.Logout
3. The TUI model should then handle the actual events from auth.Logout

You can look at the startup TUI for an example how it works.

If you want I can also just adapt it a bit here to make it fit

gtsiolis:
@silv-io I pushed cefce0e as an attempt to resolve the review comments, what do you think? You are also welcome to pick this up from here if it can help iterate faster. 🏁

silv-io:
@gtsiolis Fixed up this one as well now 👍🏼

silv-io (approved):
Fixed it up but would like @carole-lavillonniere to have another look as well :)

carole-lavillonniere:
It shows this after logging out successfully (I was definitely logged in). Probably not intended? @gtsiolis @silv-io 
image

silv-io:
hm weird I just double checked locally and for me it shows a success after log out

carole-lavillonniere (approved):

carole-lavillonniere:
> hm weird I just double checked locally and for me it shows a success after log out
> 
> 

That's because I have the auth token var set, will solve this in another PR.

PR #38, opened 2026-02-16
Add TUI architecture with a header

silv-io:
## Motivation

We’ve already merged the generic event/sink foundation, so this PR builds a minimal TUI on top of that architecture. The goal is to keep output routing consistent across TTY and non-TTY modes, have the Nimbo header branding, and make future feature emitters plug into the same typed event pipeline.

## Changes

- Added a dedicated TUI sink in internal/output (`TUISink`) that forwards events via `Program.Send()`.
- Introduced a minimal Bubble Tea UI in internal/ui with a simple static Nimbo header and chronological message feed below it. This will change in the future to be more dynamic.
- Wired output mode selection at the command boundary
  - interactive terminals use BubbleTea + `TUISink`,
  - non-interactive mode uses `PlainSink`.
- Remove all waiting for user input from non-ui components. Instead introduce a `UserInputRequestEvent` with included response channel for the UI to interact with
- Updated Claude.md to enforce Bubble Tea/TUI best-practices using the event/sink architecture for future emitters.

This is how it looks like now:
image

We'll continuously update this to contain all the content we want in the header and also to not just print out line after line after the header. This is a first iteration :) 

## Tests

Added new tests for TUISink, styles, header component, and app (login and general startup)

Also, I've changed the Device Flow and Browser Flow tests to use a PTY to be able to interact with BubbleTea properly


internal/auth/auth.go
carole-lavillonniere: nit: This is not used anymore, let's remove it?

internal/auth/login.go
```diff
@@ -76,7 +78,12 @@ func (b *browserLogin) Login(ctx context.Context) (string, error) {
 		}
 	}()
 
-	enterCh := make(chan struct{}, 1)
+	enterCh := b.enterSignal
+	if enterCh == nil {
+		enterCh = stdinEnterSignal()
+	} else {
+		drainEnterSignal(enterCh)
```
carole-lavillonniere: nit: can we add a comment to explain why `drainEnterSignal` is needed?

internal/ui/components/header.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (h Header) View() string {
+	title := styles.Title.Render("LocalStack (lstk)")
+	version := styles.Version.Render(h.version)
+
+	return "\n" + nimboLine1() + " " + title + "\n" +
+		nimboLine2() + " " + version + "\n" +
+		nimboLine3() + "\n"
+}
```
carole-lavillonniere, line 56: suggestion: can we use `lipgloss.JoinHorizontal` to join the logo and text columns side by side, instead of building each line manually? Something like:
```
func (h Header) View() string {
	logo := lipgloss.JoinVertical(lipgloss.Left,
		nimboLine1(),
		nimboLine2(),
		nimboLine3(),
	)

	text := lipgloss.JoinVertical(lipgloss.Left,
		styles.Title.Render("LocalStack (lstk)"),
		styles.Version.Render(h.version),
		"",
	)

	return "\n" + lipgloss.JoinHorizontal(lipgloss.Top, logo, " ", text) + "\n"
}
```

internal/ui/styles/styles_test.go
```diff
@@ -0,0 +1,17 @@
...
+	if NimboDarkColor != "#3F51C7" {
+		t.Fatalf("unexpected NimboDarkColor: %s", NimboDarkColor)
+	}
+	if NimboMidColor != "#5E6AD2" {
+		t.Fatalf("unexpected NimboMidColor: %s", NimboMidColor)
+	}
+	if NimboLightColor != "#7E88EC" {
+		t.Fatalf("unexpected NimboLightColor: %s", NimboLightColor)
+	}
+}
```
carole-lavillonniere, line 17: suggestion: what about testing the exact value of the rendered header (strings and colors) instead of tests `TestNimboColorConstants` and `TestHeaderViewIncludesNimboAndVersion`?
We could use https://pkg.go.dev/gotest.tools/v3/golden to write it to a golden file.
A failing test would show an exact diff and it would cover the full rendered layout not just substring presence.

internal/auth/login.go
silv-io: I think I'll remove that whole part so that we don't take inputs in the internal code anymore but have this all just work through public function calls

internal/ui/components/header.go
silv-io: you're right! I had it like this before even before I redid it to work with the event sink system

silv-io: Done, also used the built in padding now :)

internal/ui/styles/styles_test.go
silv-io: Done, not sure how the diffs will look like with the ANSI colors enabled though :D

internal/auth/auth.go
```diff
@@ -11,25 +11,38 @@ import (
...
 	browserLogin LoginProvider
 	sink         output.Sink
+	allowLogin   bool
 }
 
-func New(sink output.Sink, platformClient api.PlatformAPI) (*Auth, error) {
-	storage, err := newAuthTokenStorage()
-	if err != nil {
-		return nil, err
+func New(cfg AuthConfig) (*Auth, error) {
```
carole-lavillonniere: question: Why do we introduce a new `AuthConfig` here instead of passing each arguments?
Why don't we always pass the storage and create it wherever `New` is called?

internal/ui/run_login_test.go
```diff
@@ -0,0 +1,267 @@
...
+func (m *mockTokenStorage) SetAuthToken(token string) error {
+	m.token = token
+	return nil
+}
+
+func (m *mockTokenStorage) DeleteAuthToken() error {
+	m.token = ""
+	return nil
+}
+
```
carole-lavillonniere: question: it's not clear to me why we re-implement a mock token storage here when we have a generated one at `internal/auth/mock_token_storage_test.go`

internal/ui/run_login_test.go
```diff
@@ -0,0 +1,267 @@
...
+	case <-time.After(10 * time.Second):
+		t.Fatal("timeout waiting for login")
+	}
+
+	tm.Send(tea.QuitMsg{})
+	tm.WaitFinished(t, teatest.WithFinalTimeout(time.Second))
+
+	out := readOutput(tm.FinalOutput(t))
+	assert.Contains(t, out, "Login successful")
+}
```
carole-lavillonniere, line 236: suggestion: I understand why the integration tests were replaced but I think it's worth keeping them alongside the teatest ones. They catch different things: teatest bypasses the binary, whereas the integration tests do actual env vars handling, actual keyring access, etc.

I did a little bit of research and we could subprocess through a PTY instead of a pipe using github.com/creack/pty. When the bin is started with a PTY bubbletea should run in interactive mode.

The assertions can stay simple,  we don't need to assert on exact rendering just that key strings appear.

WDYT?

silv-io: good find! I'll try to bring them back with a PTY

internal/auth/auth.go
silv-io: It was just to make the parameter optional since other call sites didn't set it. But yeah it would actually fit the rest of the style better if we put the auth token creation to the caller (like we do with sinks, platform clients etc.). Will change it.

internal/ui/run_login_test.go
silv-io: We can't import that one I think because it's in a `_test.go` file. I can remove the `_test` from there. If we only import it in tests it won't be added to the final binary.

carole-lavillonniere (approved):
🥳

PR #39, opened 2026-02-16
Centralize env vars with Viper

carole-lavillonniere:
- Introduces centralized environment variable management using Viper, replacing scattered `os.Getenv()` calls throughout the codebase, eliminating raw string env var names.
- Introduces `test/integration/env` to do the same for integration tests.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -8,6 +8,7 @@ import (
 	"github.com/localstack/lstk/internal/api"
 	"github.com/localstack/lstk/internal/config"
 	"github.com/localstack/lstk/internal/container"
+	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/output"
 	"github.com/localstack/lstk/internal/runtime"
 	"github.com/localstack/lstk/internal/ui"
@@ -27,6 +28,7 @@ var rootCmd = &cobra.Command{
 		if cmd.Name() == "version" {
 			return nil
 		}
+		env.Init()
 		return config.Init()
 	},
 	Run: func(cmd *cobra.Command, args []string) {
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -13,7 +13,7 @@ type Env struct {
 	Keyring     string
 }
 
-var Vars *Env
+var Vars = &Env{}
 
 // Init initializes environment variable configuration
 func Init() {
--- a/internal/env/env.go
+++ b/internal/env/env.go
@@ -1,0 +1,33 @@
+package env
+
+import (
+	"strings"
+
+	"github.com/spf13/viper"
+)
+
+type Env struct {
+	AuthToken   string
+	APIEndpoint string
+	WebAppURL   string
+	Keyring     string
+}
+
+var Vars *Env
+
+// Init initializes environment variable configuration
+func Init() {
+	viper.SetEnvPrefix("LOCALSTACK")
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
+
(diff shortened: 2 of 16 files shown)


cmd/root.go
```diff
@@ -18,6 +18,7 @@ var rootCmd = &cobra.Command{
 	Short: "LocalStack CLI",
 	Long:  "lstk is the command-line interface for LocalStack.",
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		config.InitEnv()
 		return config.Init()
```
carole-lavillonniere: todo: merge `InitEnv` and `Init` in one method

carole-lavillonniere:
@coderabbitai review

carole-lavillonniere:
@coderabbitai review

internal/env/env.go
```diff
@@ -0,0 +1,33 @@
...
+	APIEndpoint string
+	WebAppURL   string
+	Keyring     string
+}
+
+var Vars = &Env{}
+
+// Init initializes environment variable configuration
+func Init() {
+	viper.SetEnvPrefix("LOCALSTACK")
```
silv-io: When I did this for the jam project I was super annoyed that you can't set two prefixes. I'd love to have a LSTK prefix for the CLI and a LOCALSTACK prefix for the container. You'd then need to handle both manually.

Then the LOCALSTACK prefix would just contain things that configure the container and are defined in [here](https://docs.localstack.cloud/aws/capabilities/config/configuration/) and the LSTK prefix would just be for things like the Keyring. 

Maybe it's worth it to investigate how to do that best in a future iteration. Could be that it is possible but I just missed it. WDYT?

test/integration/env/env.go
```diff
@@ -0,0 +1,68 @@
...
+type Key string
+
+const (
+	AuthToken   Key = "LOCALSTACK_AUTH_TOKEN"
+	APIEndpoint Key = "LOCALSTACK_API_ENDPOINT"
+	Keyring     Key = "LOCALSTACK_KEYRING"
+	CI          Key = "CI"
+)
+
+// Get returns the value of the given environment variable.
```
silv-io: nit: some maybe not necessary comments here. But I guess for hover mode in vim it's nice to see these

test/integration/login_device_flow_test.go
```diff
@@ -137,10 +133,7 @@ func TestDeviceFlowFailure_RequestNotConfirmed(t *testing.T) {
 	defer cancel()
 
 	cmd := exec.CommandContext(ctx, binaryPath(), "login")
-	cmd.Env = append(
-		envWithout("LOCALSTACK_AUTH_TOKEN"),
-		"LOCALSTACK_API_ENDPOINT="+mockServer.URL,
-	)
+	cmd.Env = env.Without(env.AuthToken).With(env.APIEndpoint, mockServer.URL)
```
silv-io: praise: looks so clean 🤩

silv-io (approved):
LGTM! Just some ideas about how to handle cli env vars vs. container env vars

internal/env/env.go
carole-lavillonniere: I'm thinking all container-specific env vars should probably rather go in the config, wdyt? Otherwise it could get messy when having multiple containers.
But in any case for now I could rename our env vars to be prefixed with `LSTK_` rather than `LOCALSTACK_` since they are CLI-specific, makes sense?

test/integration/env/env.go
carole-lavillonniere: That was way too much comments indeed. Removed.

internal/env/env.go
silv-io: yeah let's do LSTK instead, faster to type as well :P

PR #40, opened 2026-02-17
Handle already-running containers and port conflicts

carole-lavillonniere:
Before pulling images or validating licenses, `lstk start` now checks the state of each configured container:

- If a container is already running, it logs `"<name> is already running"` and skips it. If all containers are already running, the command exits successfully without doing any work.
- If the configured port is already bound by another process, it returns an error `"port <N> already in use"` with a hint pointing to the config file.

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -3,6 +3,7 @@ package container
 import (
 	"context"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	stdruntime "runtime"
@@ -54,6 +55,15 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, platformCl
 		}
 	}
 
+	// Check state before proceeding: skip already-running containers, reject port conflicts.
+	containers, cfg.Containers, err = selectContainersToStart(ctx, rt, sink, containers, cfg.Containers)
+	if err != nil {
+		return err
+	}
+	if len(containers) == 0 {
+		return nil
+	}
+
 	// Pull all images first
 	for _, config := range containers {
 		// Remove any existing stopped container with the same name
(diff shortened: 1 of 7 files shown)


carole-lavillonniere:
@coderabbitai review

internal/container/start.go
```diff
@@ -54,53 +55,113 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, platformCl
...
 		}
 
-		output.EmitStatus(sink, "ready", config.Name, fmt.Sprintf("containerId: %s", containerID[:12]))
+		output.EmitStatus(sink, "ready", c.Name, fmt.Sprintf("containerId: %s", containerID[:12]))
 	}
-
 	return nil
 }
 
+func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink output.Sink, containers []runtime.ContainerConfig, cfgContainers []config.ContainerConfig) ([]runtime.ContainerConfig, []config.ContainerConfig, error) {
```
silv-io: thought: that's a long line. Wild that `gofmt` doesn't split it up :) Maybe `gofumpt` would?

internal/container/start.go
```diff
@@ -54,53 +55,113 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, platformCl
...
+	var filtered []runtime.ContainerConfig
+	var filteredCfg []config.ContainerConfig
```
silv-io: thought: I wonder if we could somehow compose those together so that we don't have to handle them separately, like having runtime.ContainerConfig contain config.ContainerConfig, somewhat like this:

```
type ContainerConfig struct {
	Image      string
	Name       string
	Port       string
	HealthPath string
	Env        []string // e.g., ["KEY=value", "FOO=bar"]
	*config.ContainerConfig
}
```

Not sure if it wouldn't make things annoying somewhere else though :)

internal/container/start.go
```diff
@@ -54,53 +55,113 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, platformCl
...
+	conn, err := net.DialTimeout("tcp", "localhost:"+port, time.Second)
+	if err != nil {
+		return nil
+	}
+	err = conn.Close()
+	if err != nil {
+		return nil
+	}
+	return fmt.Errorf("port %s already in use", port)
+}
```
silv-io: thought: I wonder if it makes sense to have that somewhere in a util/x, but we might only need it here.

silv-io (approved):
LGTM! Just some thoughts that popped up

internal/container/start.go
carole-lavillonniere: Yeah gofmt does not care about line length and leave it up to the coder to wrap.
So if we want it we'd need to introduce another tool.

internal/container/start.go
carole-lavillonniere: That's a fair point, it looks quite complex at the moment.
That suggestion would require` internal/runtime` to import `internal/config` though. It could cause import cycle.
What about this instead?  02272426ebf76f9291e7da87e8eca06a447cf1dc
I added `Tag` and `ProductName` to `runtime.ContainerConfig `

internal/container/start.go
carole-lavillonniere: Makes sense, moved it to `internal/ports` f4f420e4f11c2b0ef781681041080e233e447392

silv-io (approved):
LGTM!

PR #41, opened 2026-02-17
Add calendar-versioned release automation

silv-io:
## Motivation
We need a reliable, tag-driven release pipeline for `lstk` using goreleasesr.
For this we'll need to embed build metadata into the binaries.
Also, we're opting for SemVer-compatible CalVer here to align with the general product decision of using calendar versioning going forward.

## Changes
- build metadata vars in cmd (version, commit, buildDate)
- `version` output commands (`lstk version` and `lstk --version`)
- `.goreleaser.yaml` for linux, macOS and Windows on ARM and AMD
- Release workflow triggered on tag push, that validates that the tag is in CalVer format and then publishes on GitHub Releases
- Manual workflow to push the next tag
- Update README to explain how versioning works

## Tests
```
goreleaser check
goreleaser release --snapshot --clean --skip=publish
```

## Related

FLC-325

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -22,11 +22,8 @@ var rootCmd = &cobra.Command{
 	Short: "LocalStack CLI",
 	Long:  "lstk is the command-line interface for LocalStack.",
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		// Version should be side-effect free and must not create/read user config.
 		if cmd.Name() == "version" {
-			return nil
-		}
-
-		if showVersion, err := cmd.Flags().GetBool("version"); err == nil && showVersion {
 			return nil
 		}
 		return config.Init()
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -13,11 +13,22 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var version = "dev"
+var commit = "none"
+var buildDate = "unknown"
+
 var rootCmd = &cobra.Command{
 	Use:   "lstk",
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -13,11 +13,22 @@ import (
...
 	Short: "LocalStack CLI",
 	Long:  "lstk is the command-line interface for LocalStack.",
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		if cmd.Name() == "version" {
+			return nil
+		}
+
+		if showVersion, err := cmd.Flags().GetBool("version"); err == nil && showVersion {
+			return nil
+		}
```
carole-lavillonniere: suggestion: maybe worth adding a comment here because it took me a while to understand that we want to skip initializing the config when running the `version` command.
Another option might be to define `PersistentPreRunE` one level down, per command, so we simply don't call it for `version`?
Also from what I understood, `--version` never reaches `PersistentPreRunE` since it is a built-in flag that prints the template set with`SetVersionTemplate`, so we could remove the second `if`?

carole-lavillonniere (approved):
Left one suggestion, but it looks 🌟

cmd/root.go
silv-io: I think if we remove the config init from the root command it might introduce more noise to the other commands. I don't see us not needing it anywhere else. I'll add a comment though for why that's needed.

I'll check with the `--version` tag and remove the other `if`. Thanks for pointing it out :)

PR #42, opened 2026-02-18
Add `lstk logs` command

carole-lavillonniere:
`lstk logs` streams container logs to the terminal until the user cancels with Ctrl+C.

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -1,0 +1,27 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/localstack/lstk/internal/container"
+	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/runtime"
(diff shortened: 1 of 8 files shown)


carole-lavillonniere:
@coderabbitai review

silv-io (approved):
LGTM! We can then do the TUI integration in a next iteration. 

Also what the rabbit says seems legit, so might want to look into it :D Never would've caught that one

carole-lavillonniere:
> LGTM! We can then do the TUI integration in a next iteration.
> 
> Also what the rabbit says seems legit, so might want to look into it :D Never would've caught that one

Apparently it's deprecated in docker :shrug: 
https://github.com/moby/moby/blob/89c5e8fd66634b6128fc4c0e6f1236e2540e46e0/errdefs/is.go#L12
Claude: 1, Rabbit: 0

PR #44, opened 2026-02-20
Rename auth token env var

carole-lavillonniere:
Rename `LSTK_AUTH_TOKEN` -> `LOCALSTACK_AUTH_TOKEN`

`LOCALSTACK_AUTH_TOKEN` is not prefixed with `LSTK_` in order to be shared seamlessly with other LocalStack tools


silv-io (approved):
LGTM!

PR #49, opened 2026-02-25
Add payload to auth request

carole-lavillonniere:
Add `{"actor": "lstk", "version": "0.1.0"}` as payload of `POST /v1/auth/request`.
Related PR https://github.com/localstack/localstack-platform/pull/1695
closes DRG-548
cc @Pive01

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/localstack/lstk/internal/env"
+	"github.com/localstack/lstk/internal/version"
 )
 
 type PlatformAPI interface {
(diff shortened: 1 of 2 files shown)


internal/api/client.go
```diff
@@ -73,10 +74,20 @@ func NewPlatformClient() *PlatformClient {
 }
 
 func (c *PlatformClient) CreateAuthRequest(ctx context.Context) (*AuthRequest, error) {
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.baseURL+"/v1/auth/request", nil)
+	payload := map[string]string{
+		"actor":   "cli-v2",
+		"version": version.Version(),
```
silv-io, line 81: q: While developing this will have the value `dev`. Will the endpoint be able to handle that?

internal/api/client.go
```diff
@@ -73,10 +74,20 @@ func NewPlatformClient() *PlatformClient {
 }
 
 func (c *PlatformClient) CreateAuthRequest(ctx context.Context) (*AuthRequest, error) {
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.baseURL+"/v1/auth/request", nil)
+	payload := map[string]string{
+		"actor":   "cli-v2",
```
silv-io: q: why not `lstk`?

silv-io (approved):
LGTM!

Just some clarifying questions

internal/api/client.go
carole-lavillonniere: Question for @Pive01

internal/api/client.go
carole-lavillonniere: f9099cd4e57b56f96cf139d2aeac53a67b3a57db

internal/api/client.go
Pive01: Yeah currently we do not do anything with this data, but in the future we could say "you need version x.y.z of this product to authenticate"....also we get some stats of how much is this endpoint used by who so what you will put there matters but not too much as long as there is something.....So to answer the question, dev versions are fine :smile:

silv-io: just to make clear, the string will literally be `dev` so it wouldn't parse as `x.y.z`. 
But if that causes any issues in the future we can just simply change it up when the time arises so it's fine :D

carole-lavillonniere: it will only be dev for local builds though, right @silv-io?

silv-io: yes, but as discussed we might want to adapt local builds to do goreleaser snapshot builds instead, so that we always have some kind of version in there.

PR #46, opened 2026-02-20
Better error for missing container runtime

gtsiolis:
Here's an attempt to resolve DES-114 and DRG-538. Early draft having fun with CLI UX, anything missing?

| CLIv1 | CLIv2 BEFORE | CLIv2 AFTER |
|--------|--------|--------|
| Screenshot 2026-02-20 at 21 43 34 | Screenshot 2026-02-20 at 21 43 24 | Screenshot 2026-02-20 at 21 42 30 |

PR #50, opened 2026-02-25
Improve event system structure and add reusable TUI components

silv-io:
## Motivation

The current event system has fragmented types (`LogEvent`, `WarningEvent`) that don't scale well as we add more UI feedback patterns. We also lack reusable TUI components for common patterns like spinners and structured error display.

This PR introduces a unified event system that makes it easier to:
- Distinguish between informational messages, successes, warnings, and notes
- Show/hide loading spinners with explicit control
- Display rich, structured errors with recovery actions

## Changes

**New event types in `internal/output/`:**
- `MessageEvent` with severity levels (Info, Success, Note, Warning) — replaces `LogEvent`/`WarningEvent`
- `SpinnerEvent` for explicit spinner start/stop control
- `ErrorEvent` for structured errors with title, summary, detail, and suggested actions

**New TUI components in `internal/ui/components/`:**
- `Spinner` — wraps Bubble Tea's spinner with simple Start/Stop API
- `RenderMessage()` — styled message rendering by severity
- `ErrorDisplay` — rich error panel with actions

**Backward compatible:** `EmitLog()` and `EmitWarning()` still work (marked deprecated, forward to `MessageEvent`).

## Tests

- Unit tests for all new event formatting
- Component tests for Spinner and ErrorDisplay
- App integration tests for event handling
- Updated existing tests to use new event types

```bash
go test ./internal/output/... ./internal/ui/...
```

## Follow-ups

- Use `EmitSpinnerStart`/`EmitSpinnerStop` in `start` command during image pull (or introduce a loading bar -> we'll see what's better)
- Use `EmitError` for Docker connection failures and auth errors

## Related

- FLC-369
- #37 & #46 will both benefit from the changes here to use common components


internal/ui/components/message.go
```diff
@@ -0,0 +1,20 @@
...
+	switch e.Severity {
+	case output.SeveritySuccess:
+		return prefix + styles.Success.Render("Success:") + " " + styles.Message.Render(e.Text)
+	case output.SeverityNote:
+		return prefix + styles.Note.Render("Note:") + " " + styles.Message.Render(e.Text)
+	case output.SeverityWarning:
+		return prefix + styles.Warning.Render("Warning:") + " " + styles.Message.Render(e.Text)
+	default:
+		return styles.Message.Render(e.Text)
+	}
```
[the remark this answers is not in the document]
silv-io: This is ok actually. TUI is allowed to render these differently. I think they will look even more different than this in the future.

internal/output/events.go
```diff
@@ -68,12 +106,25 @@ func Emit[E Event](sink Sink, event E) {
...
+}
+
+func EmitNote(sink Sink, text string) {
+	Emit(sink, MessageEvent{Severity: SeverityNote, Text: text})
+}
+
+// Deprecated: Use EmitInfo instead
 func EmitLog(sink Sink, message string) {
-	Emit(sink, LogEvent{Message: message})
+	Emit(sink, MessageEvent{Severity: SeverityInfo, Text: message})
```
carole-lavillonniere: question: Why keeping it? Isn't it deadcode?

internal/output/events.go
```diff
@@ -97,3 +148,15 @@ func EmitUserInputRequest(sink Sink, event UserInputRequestEvent) {
 func EmitContainerLogLine(sink Sink, line string) {
 	Emit(sink, ContainerLogLineEvent{Line: line})
 }
+
+func EmitSpinnerStart(sink Sink, text string) {
+	Emit(sink, SpinnerEvent{Active: true, Text: text})
+}
+
+func EmitSpinnerStop(sink Sink) {
+	Emit(sink, SpinnerEvent{Active: false})
```
carole-lavillonniere, line 152: nit: I feel it'd be clearer to have 2 events: `SpinnerStartEvent{Text: text}` and `SpinnerStopEvent{}`

internal/output/plain_format.go
```diff
@@ -71,3 +76,37 @@ func formatUserInputRequest(e UserInputRequestEvent) string {
...
+		sb.WriteString(e.Detail)
+	}
+	for _, action := range e.Actions {
+		sb.WriteString("\n  → ")
+		sb.WriteString(action.Label)
+		sb.WriteString(" ")
+		sb.WriteString(action.Value)
+	}
+	return sb.String()
+}
```
carole-lavillonniere, line 112: suggestion: the name of this file is a bit confusing/too generic. What about `plain_format.go` or `text_format.go`?

carole-lavillonniere: suggestion: the output package is also dealing with a lot of things at the moment, what about restructuring like this:
- internal/events/: event types, Sink interface, Emit* helpers
- internal/output/: PlainSink
- internal/ui/: TUISink 

You might have a better idea on how to split.

carole-lavillonniere (approved):
Added a few minor questions/suggestions but overall I approve the direction!

internal/output/events.go
silv-io: I was first thinking about not changing all the usages, but then the linter complained either way about using deprecated code, so I changed it already. Will remove those

internal/output/events.go
silv-io: I'm afraid of increasing the amount of events too much so that they become undiscoverable.
Wanted to keep it manageable by grouping the events a bit more like with the message event.

internal/output/plain_format.go
silv-io: you're right. I'll restructure it a bit. It first existed as an idea to keep the formatting of both outputs the same, but I noticed this wouldn't make sense because the TUI has more capabilities.

silv-io: Actually, moving away the implementations of the sinks from the interface is not possible because the interface intentionally has an unexported function. I'll do the renaming now and rethink the structure in a follow-up

PR #51, opened 2026-02-25
Implement `lstk logs --follow` and exit by default

anisaoshafi:
Implement `lstk logs --follow`, which should stream the logs. Exit by default on `lstk logs`.

Resolves DRG-537

PR #52, opened 2026-02-25
Improve help command layout

gtsiolis:
This will improves `help` command output readability and design.

What changed?
1.  Updated root help usage to a single line (Usage: lstk [options] [command])
1. Added a concise descriptor line under usage (LSTK - LocalStack command-line interace)
1. Renamed section headers in help output (Commands, Options)
1. Updated root flag descriptions (Show help, Show version)
1. Removed default Cobra footer
1. Updated version command short text

| BEFORE | AFTER |
|--------|--------|
| Frame 514780677 | Frame 514780677-1 |

### DIFF
--- a/cmd/help_test.go
+++ b/cmd/help_test.go
@@ -1,0 +1,71 @@
+package cmd
+
+import (
+	"bytes"
+	"context"
+	"strings"
+	"testing"
+)
+
+func executeWithArgs(t *testing.T, args ...string) (string, error) {
+	t.Helper()
+
+	origOut := rootCmd.OutOrStdout()
+	origErr := rootCmd.ErrOrStderr()
+
+	buf := new(bytes.Buffer)
+	rootCmd.SetOut(buf)
+	rootCmd.SetErr(buf)
+	rootCmd.SetArgs(args)
+
+	err := rootCmd.ExecuteContext(context.Background())
+	out := buf.String()
+
+	rootCmd.SetArgs(nil)
+	rootCmd.SetOut(origOut)
+	rootCmd.SetErr(origErr)
+
+	return out, err
+}
+
+func TestRootHelpOutputTemplate(t *testing.T) {
+	out, err := executeWithArgs(t, "--help")
+	if err != nil {
+		t.Fatalf("expected no error, got %v", err)
+	}
+
+	assertContains(t, out, "Usage:")
+	assertContains(t, out, "lstk [options] [command]")
+	assertContains(t, out, "LSTK - LocalStack command-line interface")
+	assertContains(t, out, "Commands:")
+	assertContains(t, out, "Options:")
+	assertNotContains(t, out, "Available Commands:")
+	assertNotContains(t, out, `Use "lstk [command] --help" for more information about a command.`)
+}
+
+func TestSubcommandHelpUsesSubcommandUsageLine(t *testing.T) {
+	out, err := executeWithArgs(t, "start", "--help")
+	if err != nil {
+		t.Fatalf("expected no error, got %v", err)
+	}
+
+	assertContains(t, out, "Start emulator and services.")
+	assertContains(t, out, "Usage:")
+	assertContains(t, out, "lstk start")
+	assertContains(t, out, "Options:")
+	assertNotContains(t, out, "LSTK - LocalStack command-line interface")
+}
+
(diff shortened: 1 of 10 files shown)


cmd/root.go
```diff
@@ -38,6 +39,33 @@ var rootCmd = &cobra.Command{
...
+	usageTemplate = strings.Replace(
+		usageTemplate,
+		`Usage:{{if .Runnable}}
+  {{.UseLine}}{{end}}{{if .HasAvailableSubCommands}}
+  {{.CommandPath}} [command]{{end}}`,
+		`Usage: {{if not .HasParent}}lstk [options] [command]{{else}}{{.UseLine}}{{end}}{{if not .HasParent}}
+
+LSTK - LocalStack command-line interface{{end}}`,
+		1,
+	)
```
carole-lavillonniere: issue: it's a bit fragile to replace certain strings in the help template (if we upgrade cobra, it could change and break). However I see why it makes sense: if we were instead writing the template from scratch, new commands would not be automatically added and we could forget to add them. 
So I would suggest to add a unit test to make sure the help output is what we expect.
Also I would move the code to `cmd/help.go`

silv-io: Actually, there is a range over the commands in the [original template](https://github.com/spf13/cobra/blob/61968e893eee2f27696c2fbc8e34fa5c4afaf7c4/command.go#L1942-L1971) which we also could just do.

I think writing our own template might actually be a better solution in this case

silv-io: The templating is using the built-in https://pkg.go.dev/text/template, so we're quite free in what we can do there.

gtsiolis: Thanks, both! Whoa I didn't even notice that template in the file. 

Moved in d1f2fdd. Let me know if this looks better now. 🏓 🏓

cmd/help_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	err := rootCmd.ExecuteContext(context.Background())
+	out := buf.String()
+
+	rootCmd.SetArgs(nil)
+	rootCmd.SetOut(origOut)
+	rootCmd.SetErr(origErr)
+
+	return out, err
+}
```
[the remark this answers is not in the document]
gtsiolis: Appended a commit for this in 55b2658. Let me know if this does not look good.

cmd/logs.go
```diff
@@ -11,8 +11,8 @@ import (
 
 var logsCmd = &cobra.Command{
 	Use:     "logs",
-	Short:   "Show container logs",
-	Long:    "Show logs from the LocalStack container. Use --follow to stream in real-time.",
+	Short:   "Show emulator logs",
+	Long:    "Show logs from the emulator. Use --follow to stream in real-time.",
```
gtsiolis, line 15: Updated also the terminology from https://github.com/localstack/lstk/pull/51.

https://github.com/localstack/lstk/blob/a90fc63654e1732f2e8f52f88d7d3c3350c47c05/cmd/logs.go#L14-L15

 Cc @anisaoshafi @silv-io @carole-lavillonniere for visibility and because of [relevant discussion](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1772094161931569).

cmd/help_test.go
```diff
@@ -24,10 +28,27 @@ func executeWithArgs(t *testing.T, args ...string) (string, error) {
...
+	}
+}
+
+func resetFlagSet(flags *pflag.FlagSet) {
+	flags.VisitAll(func(flag *pflag.Flag) {
+		_ = flag.Value.Set(flag.DefValue)
+		flag.Changed = false
+	})
+}
+
```
carole-lavillonniere, line 51: thought: it's a little strange we have to do that. ATM `rootCmd` is a package-level singleton and I wonder if we could construct a fresh one for each test. Anyways that would be a refactoring outside the scope of this PR.
cc @silv-io

carole-lavillonniere (approved):

cmd/help_test.go
gtsiolis: Thanks, @carole-lavillonniere! Merging this one. Opened [PRO-226](https://linear.app/localstack/issue/PRO-226/refactor-rootcmd-from-package-level-singleton-to-constructor-function) to keep track of this comment.

PR #53, opened 2026-02-26
Add AGENTS.md symlink and document emulator terminology

gtsiolis:
Add agent instructions to standardize user-facing terminology, and add `AGENTS.md` symlink for other AI tools.

See [relevant discussion](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1772094161931569) for more context.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CLAUDE.md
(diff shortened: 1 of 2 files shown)


AGENTS.md
silv-io, line 1: Is this an actual symbolic link created with `ln -s` or just the file referencing the other one?

I'm open to doing either but I'm not sure this is what you've wanted here. Maybe GitHub also shows it weird :)

silv-io (approved):
LGTM in general, but a bit confused if this actually a symlink

AGENTS.md
gtsiolis: Yes, it's a symbolic link, equivalent to `ln -s CLAUDE.md AGENTS.md`.

Thanks for taking a look, @silv-io!

PR #54, opened 2026-02-26
Combine browser and auth-wait prompts in auth flow

gtsiolis:
Fix DES-148, DES-149

| BEFORE 1/2 | BEFORE 2/2 | AFTER 1/1 |
|--------|--------|--------|
| Screenshot 2026-02-26 at 18 36 25 | Screenshot 2026-02-26 at 18 36 29 | Screenshot 2026-03-03 at 12 41 03 |

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -42,9 +42,11 @@ func (a *Auth) GetToken(ctx context.Context) (string, error) {
 		return "", fmt.Errorf("authentication required: set LOCALSTACK_AUTH_TOKEN or run in interactive mode")
 	}
 
-	output.EmitInfo(a.sink, "No existing credentials found. Please log in:")
 	token, err := a.login.Login(ctx)
 	if err != nil {
+		if errors.Is(err, context.Canceled) {
+			return "", err
+		}
 		output.EmitWarning(a.sink, "Authentication failed.")
 		return "", err
 	}
--- a/internal/auth/login.go
+++ b/internal/auth/login.go
@@ -35,46 +35,32 @@ func (l *loginProvider) Login(ctx context.Context) (string, error) {
 	}
 
 	authURL := fmt.Sprintf("%s/auth/request/%s", getWebAppURL(), authReq.ID)
-	output.EmitInfo(l.sink, fmt.Sprintf("Visit: %s", authURL))
-	output.EmitInfo(l.sink, fmt.Sprintf("Verification code: %s", authReq.Code))
 
-	// Ask whether to open the browser; ENTER or Y accepts (default yes), N skips
-	browserCh := make(chan output.InputResponse, 1)
+	output.EmitAuth(l.sink, output.AuthEvent{
+		Preamble: "Welcome to lstk, a command-line interface for LocalStack",
+		Code:     authReq.Code,
+		URL:      authURL,
+	})
+	_ = browser.OpenURL(authURL)
+
+	output.EmitSpinnerStart(l.sink, "Waiting for authorization...")
+
+	responseCh := make(chan output.InputResponse, 1)
 	output.EmitUserInputRequest(l.sink, output.UserInputRequestEvent{
-		Prompt:     "Open browser now?",
-		Options:    []output.InputOption{{Key: "y", Label: "Y"}, {Key: "n", Label: "n"}},
-		ResponseCh: browserCh,
+		Prompt:     "Waiting for authorization...",
+		Options:    []output.InputOption{{Key: "any", Label: "Press any key when complete"}},
+		ResponseCh: responseCh,
 	})
 
 	select {
-	case resp := <-browserCh:
-		if resp.Cancelled {
-			return "", context.Canceled
-		}
-		if resp.SelectedKey != "n" {
-			if err := browser.OpenURL(authURL); err != nil {
-				output.EmitWarning(l.sink, fmt.Sprintf("Failed to open browser: %v", err))
-			}
-		}
-	case <-ctx.Done():
-		return "", ctx.Err()
-	}
-
-	// Wait for the user to complete authentication in the browser
-	enterCh := make(chan output.InputResponse, 1)
-	output.EmitUserInputRequest(l.sink, output.UserInputRequestEvent{
-		Prompt:     "Waiting for authentication",
-		Options:    []output.InputOption{{Key: "enter", Label: "Press ENTER when complete"}},
-		ResponseCh: enterCh,
-	})
-
-	select {
-	case resp := <-enterCh:
+	case resp := <-responseCh:
+		output.EmitSpinnerStop(l.sink)
 		if resp.Cancelled {
 			return "", context.Canceled
 		}
 		return l.completeAuth(ctx, authReq)
 	case <-ctx.Done():
+		output.EmitSpinnerStop(l.sink)
 		return "", ctx.Err()
 	}
 }
--- a/internal/output/events.go
+++ b/internal/output/events.go
@@ -49,8 +49,14 @@ type ErrorEvent struct {
 	Actions []ErrorAction
 }
 
+type AuthEvent struct {
+	Preamble string
+	Code     string
+	URL      string
+}
+
 type Event interface {
-	MessageEvent | SpinnerEvent | ErrorEvent | ContainerStatusEvent | ProgressEvent | UserInputRequestEvent | ContainerLogLineEvent
+	MessageEvent | AuthEvent | SpinnerEvent | ErrorEvent | ContainerStatusEvent | ProgressEvent | UserInputRequestEvent | ContainerLogLineEvent
 }
 
 type Sink interface {
@@ -140,6 +146,10 @@ func EmitProgress(sink Sink, container, layerID, status string, current, total i
 }
 
 func EmitUserInputRequest(sink Sink, event UserInputRequestEvent) {
+	Emit(sink, event)
+}
+
+func EmitAuth(sink Sink, event AuthEvent) {
 	Emit(sink, event)
 }
 
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -2,12 +2,14 @@ package ui
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/charmbracelet/bubbles/spinner"
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/localstack/lstk/internal/output"
 	"github.com/localstack/lstk/internal/ui/components"
+	"github.com/localstack/lstk/internal/ui/styles"
 )
 
 const maxLines = 200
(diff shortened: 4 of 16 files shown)


internal/auth/login.go
```diff
@@ -38,11 +38,10 @@ func (l *loginProvider) Login(ctx context.Context) (string, error) {
 	output.EmitLog(l.sink, fmt.Sprintf("Visit: %s", authURL))
```
gtsiolis: **issue(non-blocking)**: Not part of this PR, but we need to wrap the URL, otherwise clicking on the URL from a terminal that hides the URL because of smaller width breaks the URL it grabs to open. Cc @carole-lavillonniere 

Click only opens what's visible. 
| Default window width | Resized window width |
|--------|--------|
| Screenshot 2026-02-26 at 17 21 20 | Screenshot 2026-02-26 at 17 21 42 |

internal/auth/login.go
```diff
@@ -38,11 +38,10 @@ func (l *loginProvider) Login(ctx context.Context) (string, error) {
...
 	output.EmitLog(l.sink, fmt.Sprintf("Verification code: %s", authReq.Code))
 
-	// Ask whether to open the browser; ENTER or Y accepts (default yes), N skips
 	browserCh := make(chan output.InputResponse, 1)
 	output.EmitUserInputRequest(l.sink, output.UserInputRequestEvent{
-		Prompt:     "Open browser now?",
-		Options:    []output.InputOption{{Key: "y", Label: "Y"}, {Key: "n", Label: "n"}},
+		Prompt:     "Open browser?\nWaiting for authentication... (Press ENTER when complete)",
```
gtsiolis: **thought**: I will add some styles once https://github.com/localstack/lstk/pull/50 lands to improve information hierarchy.

internal/auth/login.go
gtsiolis: Keeping track in [DES-149](https://linear.app/localstack/issue/DES-149/text-wrap-url-on-browser-prompt).

gtsiolis: Fixed in ba04bd6.

internal/ui/app.go
```diff
@@ -62,39 +80,61 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 			for _, opt := range a.pendingInput.Options {
 				if msg.String() == opt.Key {
+					a.lines = appendLine(a.lines, styledLine{text: formatResolvedInput(*a.pendingInput, opt.Key)})
 					responseCmd := sendInputResponseCmd(a.pendingInput.ResponseCh, output.InputResponse{SelectedKey: opt.Key})
 					a.pendingInput = nil
 					a.inputPrompt = a.inputPrompt.Hide()
 					return a, responseCmd
 				}
 			}
 		}
```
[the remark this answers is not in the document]
gtsiolis: Fixed in 596ac30.

gtsiolis:
Rebased to keep fresh and address review comment in https://github.com/localstack/lstk/pull/54#pullrequestreview-3862482153.

gtsiolis:
@copilot review, please. 🤖

internal/auth/auth.go
```diff
@@ -41,9 +41,13 @@ func (a *Auth) GetToken(ctx context.Context) (string, error) {
 		return "", fmt.Errorf("authentication required: set LOCALSTACK_AUTH_TOKEN or run in interactive mode")
 	}
 
-	output.EmitLog(a.sink, "No existing credentials found. Please log in:")
+	output.EmitSecondaryLog(a.sink, "> Welcome to LSTK, a command-line interface for LocalStack")
```
carole-lavillonniere: nit: should we stick to lowercase for lstk? LSTK makes it look like an acronym

internal/auth/auth.go
```diff
@@ -41,9 +41,13 @@ func (a *Auth) GetToken(ctx context.Context) (string, error) {
...
+	output.EmitLog(a.sink, "")
```
carole-lavillonniere: issue: I don't think we should emit empty logs in order to solve a rendering problem (adding an empty line). I wonder if we can have `FormatEventLine` add spacing after certain events.
cc @silv-io

internal/ui/app.go
```diff
@@ -114,21 +155,167 @@ func sendInputResponseCmd(responseCh chan<- output.InputResponse, response outpu
...
+		}
+		return runClipboardCandidates(text, candidates)
+	case "windows":
+		return runClipboardCandidates(text, [][]string{
+			{"cmd", "/c", "clip"},
+		})
+	default:
+		return fmt.Errorf("clipboard copy not supported on %s", runtime.GOOS)
+	}
+}
```
carole-lavillonniere: issue: these functions are pretty out-of-place in this file. Also generally I think it'd be better to split the work in smaller chunks. We could have one PR that just does the auth flow change, and another one adding the ability to copy the URL, how does this sound? It would make reviewing easier.

internal/output/events.go
```diff
@@ -1,7 +1,7 @@
 package output
 
 type Event interface {
-	LogEvent | WarningEvent | ContainerStatusEvent | ProgressEvent | UserInputRequestEvent | ContainerLogLineEvent
+	LogEvent | HighlightLogEvent | SecondaryLogEvent | WarningEvent | ContainerStatusEvent | ProgressEvent | UserInputRequestEvent | ContainerLogLineEvent
```
carole-lavillonniere: issue: 
`HighlightLogEvent` and `SecondaryLogEvent` are leaking rendering information in the event layer (should be only domain specific, not rendering specific).
I would suggest instead of these 2 a `AuthInstructionsEvent`:
```
type AuthInstructionsEvent struct {
    URL  string
    Code string
}
```

Then we only emit one event instead of ~7 that are meaningless without each other.
Curious to hear what @silv-io thinks.

internal/ui/components/input_prompt.go
```diff
@@ -37,16 +35,14 @@ func (p InputPrompt) View() string {
...
-		labels := make([]string, len(p.options))
-		for i, opt := range p.options {
-			labels[i] = opt.Label
-		}
-		text += " [" + strings.Join(labels, "/") + "]"
+
+	formatted, ok := output.FormatEventLine(output.UserInputRequestEvent{
+		Prompt:  p.prompt,
+		Options: p.options,
+	})
```
carole-lavillonniere: issue: here we are creating a "fake" event in order to reuse the prompt formatting logic from `formatUserInputRequest`.
`FormatEventLine` is meant to be called in response to a received event, not like this with a "fake" event.
Instead we can extract the prompt formatting into an exported helper that both can call:
```
 // internal/output/format.go
  func FormatPrompt(prompt string, options []InputOption) string {
      // existing logic from formatUserInputRequest
  }
```

internal/ui/app.go
```diff
@@ -114,21 +155,167 @@ func sendInputResponseCmd(responseCh chan<- output.InputResponse, response outpu
...
+		if len(labels) > 1 {
+			firstLine = fmt.Sprintf("%s [%s]", firstLine, strings.Join(labels, "/"))
+		}
+	}
+
+	if selected == "" || len(labels) == 0 || (len(labels) == 1 && selected == labels[0]) {
+		return firstLine
+	}
+	return fmt.Sprintf("%s %s", firstLine, selected)
+}
```
carole-lavillonniere, line 239: issue: this duplicates `formatUserInputRequest` from format.go. Both format a prompt with its options. They differ slightly (one shows the selected key after), but structurally they're doing the same thing. If you extract `FormatPrompt` as discussed [here](https://github.com/localstack/lstk/pull/54/changes#r2871314849) you can just call it and append the selection.

internal/ui/app.go
```diff
@@ -114,21 +155,167 @@ func sendInputResponseCmd(responseCh chan<- output.InputResponse, response outpu
...
+	return fmt.Errorf("copy to clipboard failed: %s", strings.Join(errs, "; "))
+}
+
+func nextIsURL(lines []styledLine, i int) bool {
+	if i+1 < len(lines) {
+		next := lines[i+1]
+		return next.highlight && isURL(next.text)
+	}
+	return false
+}
```
carole-lavillonniere: issue: this is fragile, it relies on line ordering, if we add anything (a new line) it will break. if we introduce a auth event as discussed [here](https://github.com/localstack/lstk/pull/54/changes#r2871249573), we don't rely on line ordering but on semantic meaning

internal/auth/login.go
```diff
@@ -35,41 +34,23 @@ func (l *loginProvider) Login(ctx context.Context) (string, error) {
...
+	output.EmitLog(l.sink, "")
 
-	// Wait for the user to complete authentication in the browser
-	enterCh := make(chan output.InputResponse, 1)
+	responseCh := make(chan output.InputResponse, 1)
 	output.EmitUserInputRequest(l.sink, output.UserInputRequestEvent{
-		Prompt:     "Waiting for authentication",
-		Options:    []output.InputOption{{Key: "enter", Label: "Press ENTER when complete"}},
-		ResponseCh: enterCh,
+		Prompt:     "Waiting for authentication... (Press ENTER when complete)",
```
carole-lavillonniere: question: why did we move `(Press ENTER when complete)` to the prompt instead of using the dedicated label for it?

carole-lavillonniere (changes requested):

gtsiolis:
Thanks for taking a look, @carole-lavillonniere!

Marking this as blocked per our discussion, and will revisit once #50 lands.

internal/ui/app.go
gtsiolis: Totally, will extract this in a separate PR.

internal/auth/auth.go
silv-io: it could be if we wanted it to be `LocalStack ToolKit (for CLI)` :) 

But I like better how it looks in lower case :D

internal/auth/auth.go
silv-io: yep. those events should just send content. Any formatting should either happen in the Event line format for the plain sink or handled by the UI model for the TUI sink

internal/output/events.go
silv-io: Maybe we can even further generalize it into an `InstructionsEvent` for things that are not only auth?

gtsiolis:
Opted for opening the browser for now to limit step friction until we get to PRO-229, DES-110, etc. Let me know what you think, also tried to address the review comments above.

internal/auth/login.go
```diff
@@ -35,46 +35,32 @@ func (l *loginProvider) Login(ctx context.Context) (string, error) {
...
-	output.EmitUserInputRequest(l.sink, output.UserInputRequestEvent{
-		Prompt:     "Open browser now?",
-		Options:    []output.InputOption{{Key: "y", Label: "Y"}, {Key: "n", Label: "n"}},
-		ResponseCh: browserCh,
+	output.EmitInstructions(l.sink, output.InstructionsEvent{
+		Preamble: "Welcome to lstk, a command-line interface for LocalStack",
+		Code:     authReq.Code,
+		URL:      authURL,
 	})
+	_ = browser.OpenURL(authURL)
```
carole-lavillonniere, line 44: question to @silv-io: what did you have in mind exactly when you suggested an `InstructionsEvent`? https://github.com/localstack/lstk/pull/54#discussion_r2872884477
Here it has a a code and URL so it really looks more like a specific `AuthEvent`

silv-io: Maybe something in the direction "problem/task":"this", "guidance":"that".

But that might be too generic to create a nice looking UI in this case? So I think that's actually fine?

Maybe my "doctrine" to keep all event definitions in one place was misplaced as well because it turns out we need quite a bit of different ones. I'll think about it if it wouldn't be better solved by having something more "duck-typed" with an interface. I'll think about that, but so far this can just go in like that IMO 👍🏼

carole-lavillonniere (approved):
Let's just rename `InstructionsEvent` to `AuthEvent` for now and we're good to go! 👏

internal/auth/login.go
gtsiolis: Thanks, both! Renamed in d36e7f6.

PR #58, opened 2026-03-02
Telemetry client

carole-lavillonniere:
Adds a telemetry client that fires a single `cli_cmd` event when `lstk start` runs (this is merely to not have deadcode. Defining telemetry is still WIP PRO-210) 

- Enriches every event with `version`, `os`, `arch`, `is_ci`, `machine_id` (persistent uuid stored in the lstk config dir),  a per-session UUID (ultimately one session ID per command), a `X-Client` header with value `lstk/v2`
- Can be disabled when `LOCALSTACK_DISABLE_EVENTS=1`
- Sent in a background goroutine with a 3s timeout so a slow/unreachable endpoint never blocks
- Flushes on exit so that the event is sent even when the command returns early

## :heavy_check_mark: Tested

Ran analytics-backend locally and observed:
```{"timestamp": "2026-03-03T12:08:38.280", "level": "INFO", "threadName": "Dummy-8", "process": 1312488, "name": "analytics.core", "message": "record_events: Event(name='cli_cmd', metadata=EventMetadata(session_id='6ec424b8-0196-4815-83b7-2ae45e2d6146', client_time='2026-03-03 11:08:28.019701'), payload={'arch': 'amd64', 'cmd': 'lstk start', 'is_ci': False, 'machine_id': '5df95f6a-6dc0-4449-9ea8-f0f1b23b8f7c', 'os': 'linux', 'params': [], 'version': 'dev'})"}```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -56,7 +56,7 @@ func NewRootCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
 func Execute(ctx context.Context) error {
 	cfg := env.Init()
 	tel := telemetry.New(cfg.AnalyticsEndpoint, cfg.DisableEvents)
-	defer tel.Flush()
+	defer tel.Close()
 
 	root := NewRootCmd(cfg, tel)
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -17,25 +17,18 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func NewRootCmd(cfg *env.Env) *cobra.Command {
+func NewRootCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
 	root := &cobra.Command{
 		Use:     "lstk",
 		Short:   "LocalStack CLI",
 		Long:    "lstk is the command-line interface for LocalStack.",
 		PreRunE: initConfig,
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			rt, err := runtime.NewDockerRuntime()
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "Error: %v\n", err)
-				os.Exit(1)
+				return err
 			}
-
-			if err := runStart(cmd.Context(), rt, cfg); err != nil {
-				if !output.IsSilent(err) {
-					fmt.Fprintf(os.Stderr, "Error: %v\n", err)
-				}
-				os.Exit(1)
-			}
+			return runStart(cmd.Context(), rt, cfg, tel)
 		},
 	}
 
@@ -48,7 +41,7 @@ func NewRootCmd(cfg *env.Env) *cobra.Command {
 	root.Flags().Lookup("version").Usage = "Show version"
 
 	root.AddCommand(
-		newStartCmd(cfg),
+		newStartCmd(cfg, tel),
 		newStopCmd(),
 		newLoginCmd(cfg),
 		newLogoutCmd(cfg),
@@ -62,12 +55,21 @@ func NewRootCmd(cfg *env.Env) *cobra.Command {
 
 func Execute(ctx context.Context) error {
 	cfg := env.Init()
-	return NewRootCmd(cfg).ExecuteContext(ctx)
-}
-
-func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env) error {
 	tel := telemetry.New(cfg.AnalyticsEndpoint, cfg.DisableEvents)
 	defer tel.Flush()
+
+	root := NewRootCmd(cfg, tel)
+
+	if err := root.ExecuteContext(ctx); err != nil {
+		if !output.IsSilent(err) {
+			fmt.Fprintf(os.Stderr, "Error: %v\n", err)
+		}
+		return err
+	}
+	return nil
+}
+
+func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *telemetry.Client) error {
 	// TODO: replace map with a typed payload struct once event schema is finalised
 	tel.Track("cli_cmd", map[string]any{"cmd": "lstk start", "params": []string{}})
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -11,6 +11,7 @@ import (
 	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/output"
 	"github.com/localstack/lstk/internal/runtime"
+	"github.com/localstack/lstk/internal/telemetry"
 	"github.com/localstack/lstk/internal/ui"
 	"github.com/localstack/lstk/internal/version"
 	"github.com/spf13/cobra"
@@ -65,6 +66,11 @@ func Execute(ctx context.Context) error {
 }
 
 func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env) error {
+	tel := telemetry.New(cfg.AnalyticsEndpoint, cfg.DisableEvents)
+	defer tel.Flush()
+	// TODO: replace map with a typed payload struct once event schema is finalised
+	tel.Track("cli_cmd", map[string]any{"cmd": "lstk start", "params": []string{}})
+
 	platformClient := api.NewPlatformClient(cfg.APIEndpoint)
 	if ui.IsInteractive() {
 		return ui.Run(ctx, rt, version.Version(), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -71,7 +71,7 @@ func Execute(ctx context.Context) error {
 
 func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *telemetry.Client) error {
 	// TODO: replace map with a typed payload struct once event schema is finalised
-	tel.Track("cli_cmd", map[string]any{"cmd": "lstk start", "params": []string{}})
+	tel.Emit("cli_cmd", map[string]any{"cmd": "lstk start", "params": []string{}})
 
 	platformClient := api.NewPlatformClient(cfg.APIEndpoint)
 	if ui.IsInteractive() {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -71,7 +71,7 @@ func Execute(ctx context.Context) error {
 
 func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *telemetry.Client) error {
 	// TODO: replace map with a typed payload struct once event schema is finalised
-	tel.Emit("cli_cmd", map[string]any{"cmd": "lstk start", "params": []string{}})
+	tel.Emit(ctx, "cli_cmd", map[string]any{"cmd": "lstk start", "params": []string{}})
 
 	platformClient := api.NewPlatformClient(cfg.APIEndpoint)
 	if ui.IsInteractive() {
--- a/internal/telemetry/client.go
+++ b/internal/telemetry/client.go
@@ -3,6 +3,7 @@ package telemetry
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"os"
 	"runtime"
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: cmd/root.go [hunk 1]
```diff
@@ -53,6 +54,11 @@ func Execute(ctx context.Context) error {
 }
 
 func runStart(ctx context.Context, rt runtime.Runtime) error {
+	tel := telemetry.New()
+	defer tel.Flush()
```
carole-lavillonniere: TODO: create one telemetry client per run/command.
For this we need to re-work our dependency injection (coming in a subsequent PR).
Then we can easily flush after a command finishes, in `PersistentPostRun`.
However we need to investigate if/how to still send the event when calling `os.Exit(1)` (which we do on error)
since it kills the process before any post-run hooks fire.

Once we have this, we would also have a single `sessionID` per command which all events for that command would send.

internal/telemetry/client.go
```diff
@@ -0,0 +1,118 @@
...
+		if err != nil {
+			return
+		}
+
+		req, err := http.NewRequest(http.MethodPost, c.endpoint, bytes.NewReader(data))
+		if err != nil {
+			return
+		}
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("X-Client", clientHeader)
```
thrau: not sure what we would use this for at the moment. i like the idea of setting the user agent header though.

for example, we could build a typical user agent header:

```
localstack lstk/<version> <os>/<kernel>/<arch>

^^^^^^^^^^ ^^^^^^^^^^^^^^ ^^^^^^^^^^^^^^^^^^^^
namespace  software       system info
```

and send that. this looks future proof, standard conform, and i can see that being super useful in the analytics backend at some point :)
i would love to introduce this in the emulator also

internal/telemetry/client.go
```diff
@@ -0,0 +1,118 @@
...
+	Metadata eventMetadata `json:"metadata"`
+	Payload  any           `json:"payload"`
+}
+
+type eventMetadata struct {
+	ClientTime string `json:"client_time"`
+	SessionID  string `json:"session_id"`
+}
+
+func (c *Client) Track(name string, payload map[string]any) {
```
thrau: nit on naming: i would prefer something like `Log` or `Emit` as method name here. to me "Track" implies some sort of long-living process, a lifecycle, or a correlation of some sort. whereas this is really just a fire-and-forget logging method, and `Log` or `Emit` matches that mental model more.

thrau: i think we should pass a context object here, and hand it down to the http request, so we have lifecycle control. though not necessarily needed if we go with a producer/consumer pattern.

internal/telemetry/client.go
```diff
@@ -0,0 +1,118 @@
...
+		Name: name,
+		Metadata: eventMetadata{
+			ClientTime: time.Now().UTC().Format("2006-01-02 15:04:05.000000"),
+			SessionID:  c.sessionID,
+		},
+		Payload: enriched,
+	}
+
+	c.wg.Add(1)
+	go func() {
```
thrau: i understand that we don't want `Track` to block to avoid callers to manage concurrency, but i think there are better ways to solve this than creating a new go routine for each call and tracking their invocations using a WaitGroup.

i see a few drawbacks here:
* there's unbounded go routine creation. if we use this client to emit events in a loop, we may end up creating lots of go routines without any backpressure. nothing is slowing requests down, so this may end up creating resource spikes if used incorrectly.
* if memory serves, calling `WaitGroup.Add` with `WaitGroup.Wait` concurrently is not safe
* there's no cancellation or lifecycle control

my preference would be creating a producer/worker pattern:
* use a (optionally bounded) channel of events
* use a single go routine that reads from the channel and sends the data in a loop
* `Close` or `Shutdown` that stops acceptance of new events and drains the channel

now, `Track` becomes just a write into the channel. it gives us controlled concurrency, predictable resource usage, simpler shutdown mechanics (👀 `Flush`), option to create drop policies when queue is full, better encapsulation between concerns (marshalling data vs loading data)

internal/telemetry/machine_id.go
```diff
@@ -0,0 +1,49 @@
...
+	if err == nil {
+		id := strings.TrimSpace(string(data))
+		if id != "" {
+			return id
+		}
+	} else if !os.IsNotExist(err) {
+		return ""
+	}
+
+	id := uuid.NewString()
```
thrau: it would be great if we could attempt to use a salt hashed version of the machine id, or docker daemon id if available, so we have a predictable machine id. here's a reference in localstack: https://github.com/localstack/localstack/blob/ada377c2c63ac89f968cb31627a0d1a1dabb8284/localstack-core/localstack/utils/analytics/metadata.py#L172-L200

if we get it in the same way, we could ideally link the generated machine IDs the ones generated in localstack.

internal/telemetry/client.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+		_ = resp.Body.Close()
+	}()
+}
+
+// Flush blocks until all in-flight Track goroutines have completed. Call it
+// before process exit to avoid dropping telemetry events. It returns quickly
+// when no events are pending, and is bounded by the HTTP client's timeout in
+// the worst case.
+func (c *Client) Flush() {
```
thrau: nit on naming: as a caller, i would expect `Flush` to force buffered items to be emptied to the underlying IO writer. this method looks more like a `WaitForShutdown`? though we could avoid this if we go with the producer/consumer pattern.

thrau (changes requested):
Thanks for taking a first stab at this @carole-lavillonniere! Looks like we got the event schema down, so from an analytics-backend side this should work no problem (thanks for validating!).

It makes sense that we're filling the sessionId with a newly generated UUID. As we've spoken about, this is not used for anything else but localstack sessions at the moment, but at least we have clear semantics for the CLI for now (one command = new session id).  I don't have much to add there.

My main ask is around the machine ID, but I think it would be great if we could rework this using a producer/consumer pattern for reasons I explain inline. Happy to merge this as a first iteration though so we can get going.

thrau:
cc @vittoriopolverino in case you are interested

cmd/root.go
carole-lavillonniere: Rebased and addressed the TODO:
- We now create only one instance of telemetry client at root and pass only the required env vars
- We do not execute `os.Exit(1)` from the commands themselves but in one central place (main), after the "flushing" has happened: no more risk of exiting before the events have been sent. To achieve that we now use cobra's `RunE` instead of `Run`. `RunE` returns an error, which cobra propagates back.

internal/telemetry/client.go
carole-lavillonniere: Changed to the pattern you suggested. Skipped the kernel though since it's not directly available in go libraries, we would need to make some non-cross-platform system calls.

internal/telemetry/client.go
carole-lavillonniere: :heavy_check_mark: `Emit` it is.

carole-lavillonniere: :100: Added. I think it's good to pass the context regardless of the architecture. It's idiomatic go, it could be used in the future for propagation of values, trace contexts etc.
However I think it would be wrong behaviour to cancel the http requests for analytics. The events should go through even if:
- the user cancels the command with Ctrl+C
- the parent context had a timeout (not the case at the moment)
- the command exited early because of an error

That's why I made use of `context.WithoutCancel(ctx)`.

internal/telemetry/client.go
carole-lavillonniere: Renamed to `Close` with new producer/consumer pattern.

internal/telemetry/machine_id.go
carole-lavillonniere: Thanks for the reference!
Now we use those, in order:
- docker daemon ID (dkr_ prefix)
- /etc/machine-id (sys_ prefix)
- persisted random ID (gen_ prefix)

internal/telemetry/client.go
carole-lavillonniere: I was about to fight back with these arguments:

> there's unbounded go routine creation. if we use this client to emit events in a loop, we may end up creating lots of go routines without any backpressure. nothing is slowing requests down, so this may end up creating resource spikes if used incorrectly.

For now we mostly have discrete commands (only exception for now is `lstk logs --follow`), and likely each command emitting 1 event once we have analytics in place. I think this will become more relevant when we have a long-running interactive shell or REPL mode (which is planned), but even then I doubt we will be emitting so many events, even less so concurrently.
> if memory serves, calling WaitGroup.Add with WaitGroup.Wait concurrently is not safe

That's right, however we only `Wait` when calling `Flush`, which is deferred to when the command is finished running. So in practice at the moment we cannot call `Wait` concurrently to `Add`.


**BUT** the new producer/consumer pattern aka fire-and-forget channel does not really bring more complexity to the code, so I just went ahead and did it. It's a more solid design, I like it! Looking forward to your next round of review @thrau, thanks for taking the time to review this!

internal/telemetry/client.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	enriched := make(map[string]any, len(payload)+6)
+	for k, v := range payload {
+		enriched[k] = v
+	}
+	enriched["version"] = version.Version()
+	enriched["os"] = runtime.GOOS
+	enriched["arch"] = runtime.GOARCH
+	enriched["is_ci"] = os.Getenv("CI") != ""
```
silv-io: nit: small detail here. In the emulator and old CLI we actually test purely for the existence of this variable, not if it's not empty.

`os.LookupEnv` would be the option here then to maintain parity.

internal/telemetry/client.go
```diff
@@ -0,0 +1,139 @@
...
+	Metadata eventMetadata   `json:"metadata"`
+	Payload  any             `json:"payload"`
+}
+
+type eventMetadata struct {
+	ClientTime string `json:"client_time"`
+	SessionID  string `json:"session_id"`
+}
+
+func (c *Client) Emit(ctx context.Context, name string, payload map[string]any) {
```
silv-io, line 72: thought: I wonder if this naming won't be confusing with the other events we have. But I raised the idea of merging those into the same concept in the future either way so it makes sense.

silv-io (approved):
LGTM! Just some minor nitpicks

internal/telemetry/client.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// Close stops accepting new events, drains the event buffer, and blocks until
+// all pending HTTP requests have completed. Call it before process exit to
+// avoid dropping telemetry events.
+func (c *Client) Close() {
+	if !c.enabled {
+		return
+	}
+	close(c.events)
```
thrau: nit: maybe worth making Close idempotent. currently calling Close twice will panic because of this line

thrau (approved):
love it! i think this is much cleaner, more idiomatic, and the encapsulation of the worker will make it easier for us in the future to better control how events are sent.

happy to merge as is! 👍  but it's maybe worth testing end-to-end with the [analytics-backend integration test setup](https://github.com/localstack/analytics-backend/?tab=readme-ov-file#integration-setup-with-tinybird-local). we shoud make sure there are no errors when ingesting the event into tb local.

internal/telemetry/client.go
carole-lavillonniere: Good catch a0940f11cf6892321787a3a15249f84997c14810

internal/telemetry/client.go
carole-lavillonniere: Makes sense ec05e69f5dd9ce178dd2da6ca51b4414e9dc6d70

carole-lavillonniere:
> ut it's maybe worth testing end-to-end with the [analytics-backend integration test setup](https://github.com/localstack/analytics-backend/?tab=readme-ov-file#integration-setup-with-tinybird-local). we shoud make sure there are no errors when ingesting the event into tb local.


`docker compose -f docker-compose-itest.yml logs analytics-backend | grep lstk`
`analytics-backend-1  | {"timestamp": "2026-03-05T15:29:26.217", "level": "DEBUG", "threadName": "Dummy-2", "process": 33, "name": "analytics.web.resources.events", "message": "parsed event Event(name='cli_cmd', metadata=EventMetadata(session_id='81ad464b-4e8b-45a9-93ab-73531beee97a', client_time='2026-03-05 15:29:26.214175'), payload={'arch': 'amd64', 'cmd': 'lstk start', 'is_ci': False, 'machine_id': 'dkr_665f66e84e70', 'os': 'linux', 'params': [], 'version': 'dev'})"}`

PR #60, opened 2026-03-03
fix: logout message when auth token env var is set

carole-lavillonniere:
Fix message when the user is trying to logout but they are authenticated using the `LOCALSTACK_AUTH_TOKEN` env var:
logout


silv-io (approved):
Clean! :)

PR #61, opened 2026-03-03
Change tests output format to testname

carole-lavillonniere:
`testdox` format waits for each package to finish before flushing its output, which means the integration tests is not printing any progress during the ~100sec of its duration and often looks like it's hanging.

Switch to `--format testname`, which prints each test result line as soon as the event arrives.


silv-io (approved):
LGTM! Arguably this one is more useful either way in the logs :)

PR #63, opened 2026-03-03
Replace container name in command output

gtsiolis:
See DES-152 for more context.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-04 at 00 28 43 | Screenshot 2026-03-04 at 00 27 52 |

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -185,12 +185,12 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink, 
 		output.EmitStatus(sink, "starting", c.Name, "")
 		containerID, err := rt.Start(ctx, c)
 		if err != nil {
-			return fmt.Errorf("failed to start %s: %w", c.Name, err)
+			return fmt.Errorf("failed to start LocalStack: %w", err)
 		}
 
 		output.EmitStatus(sink, "waiting", c.Name, "")
 		healthURL := fmt.Sprintf("http://localhost:%s%s", c.Port, c.HealthPath)
-		if err := awaitStartup(ctx, rt, sink, containerID, c.Name, healthURL); err != nil {
+		if err := awaitStartup(ctx, rt, sink, containerID, "LocalStack", healthURL); err != nil {
 			return err
 		}
 
@@ -207,7 +207,7 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 			return nil, fmt.Errorf("failed to check container status: %w", err)
 		}
 		if running {
-			output.EmitInfo(sink, fmt.Sprintf("%s is already running", c.Name))
+			output.EmitInfo(sink, "LocalStack is already running")
 			continue
 		}
 		if err := ports.CheckAvailable(c.Port); err != nil {
--- a/internal/container/stop.go
+++ b/internal/container/stop.go
@@ -22,15 +22,15 @@ func Stop(ctx context.Context, rt runtime.Runtime, sink output.Sink) error {
 			return fmt.Errorf("checking %s running: %w", name, err)
 		}
 		if !running {
-			return fmt.Errorf("%s is not running", name)
+			return fmt.Errorf("LocalStack is not running")
 		}
-		output.EmitSpinnerStart(sink, fmt.Sprintf("Stopping %s", name))
+		output.EmitSpinnerStart(sink, "Stopping LocalStack...")
 		if err := rt.Stop(ctx, name); err != nil {
 			output.EmitSpinnerStop(sink)
-			return fmt.Errorf("failed to stop %s: %w", name, err)
+			return fmt.Errorf("failed to stop LocalStack: %w", err)
 		}
 		output.EmitSpinnerStop(sink)
-		output.EmitSuccess(sink, fmt.Sprintf("%s stopped", name))
+		output.EmitSuccess(sink, "LocalStack stopped")
 	}
 
 	return nil
--- a/internal/output/plain_format.go
+++ b/internal/output/plain_format.go
@@ -35,21 +35,21 @@ func FormatEventLine(event any) (string, bool) {
 func formatStatusLine(e ContainerStatusEvent) (string, bool) {
 	switch e.Phase {
 	case "pulling":
-		return "", false
+		return "Preparing LocalStack...", true
 	case "starting":
-		return fmt.Sprintf("Starting %s...", e.Container), true
+		return "Starting LocalStack...", true
 	case "waiting":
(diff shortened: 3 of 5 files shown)


internal/container/start.go
```diff
@@ -122,12 +122,12 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink,
 		output.EmitStatus(sink, "starting", c.Name, "")
 		containerID, err := rt.Start(ctx, c)
 		if err != nil {
-			return fmt.Errorf("failed to start %s: %w", c.Name, err)
+			return fmt.Errorf("failed to start LocalStack: %w", err)
```
carole-lavillonniere, line 188: nit: This code loops over all containers specified in the config. I know that we only support one container for launch, but the moment we want to support more the output will be confusing.
If someone adds a Snowflake or Azure container, all status messages will still say "LocalStack".
We could also use ProductName() which would read "localstack-pro" instead of "localstack-aws".
I am fine with ignoring this problem for now, just wanted to make you aware of it 🙂

carole-lavillonniere (approved):
Added one small concern, but it's good to go for me if you think we can ignore it!

internal/container/start.go
silv-io: We also have the Emulator Display name that we're using in the header. Could do that instead?

gtsiolis: Good insight. ⚾ 

I'd keep it as is.
1. The output here can stay consistent with the platform vision and avoid limiting the CLI in SKU-specific messaging, see [relevant discussion](https://linear.app/localstack/issue/SMF-426/legalbranding-evaluate-and-finalize-naming-for-azure-emulator#comment-746f87da) for more context. 
2. The emulation-specific information can still live as metadata or sub-information, regardless if one would start AWS, or SNO, both emulators, or side containers with tools (see DRG-364), etc.

gtsiolis: Jsut refreshed and saw your comment above in https://github.com/localstack/lstk/pull/63#discussion_r2883016278 after posting, @silv-io.

> We also have the Emulator Display name that we're using in the header. Could do that instead?

Do you mean something like this?


```diff
diff --git a/internal/container/start.go b/internal/container/start.go
index 8e55be3..85b587d 100644
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -122,12 +122,12 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink,
                output.EmitStatus(sink, "starting", c.Name, "")
                containerID, err := rt.Start(ctx, c)
                if err != nil {
-                       return fmt.Errorf("failed to start LocalStack: %w", err)
+                       return fmt.Errorf("failed to start %s: %w", c.ProductName, err)
                }

                output.EmitStatus(sink, "waiting", c.Name, "")
                healthURL := fmt.Sprintf("http://localhost:%s%s", c.Port, c.HealthPath)
-               if err := awaitStartup(ctx, rt, sink, containerID, "LocalStack", healthURL); err != nil {
+               if err := awaitStartup(ctx, rt, sink, containerID, c.ProductName, healthURL); err != nil {
                        return err
                }

@@ -144,7 +144,7 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
                        return nil, fmt.Errorf("failed to check container status: %w", err)
                }
                if running {
-                       output.EmitInfo(sink, "LocalStack is already running")
+                       output.EmitInfo(sink, fmt.Sprintf("%s is already running", c.ProductName))
                        continue
                }
                if err := ports.CheckAvailable(c.Port); err != nil {
diff --git a/internal/container/stop.go b/internal/container/stop.go
index 0e7af84..11a59dd 100644
--- a/internal/container/stop.go
+++ b/internal/container/stop.go
@@ -16,14 +16,18 @@ func Stop(ctx context.Context, rt runtime.Runtime, onProgress func(string)) erro
        }

        for _, c := range cfg.Containers {
-               onProgress("Stopping LocalStack...")
+               productName, err := c.ProductName()
+               if err != nil {
+                       productName = "LocalStack"
+               }
+               onProgress(fmt.Sprintf("Stopping %s...", productName))
                if err := rt.Stop(ctx, c.Name()); err != nil {
                        if errdefs.IsNotFound(err) {
-                               return fmt.Errorf("LocalStack is not running")
+                               return fmt.Errorf("%s is not running", productName)
                        }
-                       return fmt.Errorf("failed to stop LocalStack: %w", err)
+                       return fmt.Errorf("failed to stop %s: %w", productName, err)
                }
-               onProgress("LocalStack stopped")
+               onProgress(fmt.Sprintf("%s stopped", productName))
        }

        return nil
```

silv-io: yes!

gtsiolis: Coming back to this! Rebased, resolved conflicts and pushed.

@silv-io I've thought about this and think adding the emulator-specific name is something we will have to revert later in the output as LocalStack matures as a platform, to avoid legal nounnce of using the AWS or other emulator names in the output, etc.

Using the display name seems quite verbose and repeating (4 times repeating what's in the header).

### BEFORE (CURRENT)

```sh

    ▟████▖    lstk (dev)
   ▟██▙█▙█▟   LocalStack AWS Emulator
     ▀▛▀▛▀    localhost.localstack.cloud:4566

  > Success: Pulled localstack/localstack-pro:latest
  LocalStack: validating license (4.14.1.dev108)
  Starting LocalStack...
  Waiting for LocalStack to be ready...
  LocalStack ready (containerId: eadc8fd4f7e1)
```

### AFTER

```sh

    ▟████▖    lstk (dev)
   ▟██▙█▙█▟   LocalStack AWS Emulator
     ▀▛▀▛▀    localhost.localstack.cloud:4566

  > Success: Pulled localstack/localstack-pro:latest
  LocalStack AWS Emulator: validating license (4.14.1.dev108)
  Starting LocalStack AWS Emulator...
  Waiting for LocalStack AWS Emulator to be ready...
  LocalStack AWS Emulator ready (containerId: 49cb569fb0bc)
```


What do you think of going with LocalStack for now? Open to both if you think it's better for now.

silv-io: It's fine for now to keep as LocalStack, so let's merge it like this. 

However, consider the case of starting multiple emulators at once, how are we going to display that then? This is something we'll have to explore further.

silv-io (approved):

internal/container/start.go
gtsiolis: Good point. What I had in mind: Refer to all as LocalStack. Under the hood we can print container-specific information, and this will get more informed as our pricing and packaging evolves.

PR #65, opened 2026-03-03
Update header information

gtsiolis:
See DES-152 for more context.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-03 at 15 20 18 | Screenshot 2026-03-03 at 15 16 57 |


internal/ui/run.go
```diff
@@ -28,7 +30,14 @@ func Run(parentCtx context.Context, rt runtime.Runtime, version string, platform
...
 
-	app := NewApp(version, cancel)
+	emulatorName := "LocalStack Emulator"
+	endpoint := "localhost.localstack.cloud"
+	if cfg, err := config.Get(); err == nil && len(cfg.Containers) > 0 {
+		emulatorName = cfg.Containers[0].DisplayName()
+		endpoint = fmt.Sprintf("localhost.localstack.cloud:%s", cfg.Containers[0].Port)
+	}
+
+	app := NewApp(version, emulatorName, endpoint, cancel)
```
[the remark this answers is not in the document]
gtsiolis: Addressed in e8530bc.

internal/ui/components/header.go
```diff
@@ -45,12 +47,17 @@ func (h Header) View() string {
...
 	spacer := strings.Repeat(" ", headerPadding)
 
-	return "\n" + lipgloss.JoinHorizontal(lipgloss.Top, nimbo, spacer, text) + "\n"
+	joined := lipgloss.JoinHorizontal(lipgloss.Top, nimbo, spacer, text)
+	lines := strings.Split(joined, "\n")
+	for i, line := range lines {
+		lines[i] = strings.TrimRight(line, " ")
+	}
+	return "\n" + strings.Join(lines, "\n") + "\n"
 }
```
silv-io, line 63: nit: this looks a bit weird? What does it do exactly? Is this necessary?

internal/ui/run.go
```diff
@@ -28,7 +30,16 @@ func Run(parentCtx context.Context, rt runtime.Runtime, version string, platform
...
-	app := NewApp(version, cancel)
+	emulatorName := "LocalStack Emulator"
+	endpoint := "localhost.localstack.cloud"
+	if cfg, err := config.Get(); err == nil && len(cfg.Containers) > 0 {
+		emulatorName = cfg.Containers[0].DisplayName()
+		if cfg.Containers[0].Port != "" {
+			endpoint = fmt.Sprintf("localhost.localstack.cloud:%s", cfg.Containers[0].Port)
+		}
+	}
+
```
silv-io, line 42: nit: this looks a bit hacky but we can allow it for now I think. We'll have to revisit when we do proper multi-emulator support. 

Maybe add a "FIXME" comment? Don't know if the linter won't cry about that though :)

silv-io (approved):
LGTM! Just a bit of weirdness in the final header rendering and the extraction of the Container metadata

internal/ui/components/header.go
gtsiolis: It adds one empty line above the header; otherwise, the avatar would collide with the line above.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-03 at 18 33 46 | Screenshot 2026-03-03 at 18 34 33 |

silv-io: no I don't mean the last line but the thing with splitting the joined lines again and then trimming them on the right side and then joining them again. Seems a bit wasteful. Why do we need to trim them on the right?

internal/ui/run.go
gtsiolis: Good point, added a comment in a6fad93.

internal/ui/components/header.go
gtsiolis: Ah, good point. I remember seeing some extra trailing characters in the following file:

```sh
/internal/ui/components/testdata/header.golden
```

Plus, tests are failing without this. Any ideas?

```sh
FAIL internal/ui/components

=== Failed
=== FAIL: internal/ui/components TestHeaderView (0.00s)
    header_test.go:17: assertion failed:
        --- expected
        +++ actual
        @@ -1,5 +1,5 @@

        -····▟████▖····lstk·(v1.0.0)
        -···▟██▙█▙█▟···LocalStack·AWS·Emulator
        -·····▀▛▀▛▀····localhost.localstack.cloud:4566
        +····▟████▖·lstk·(v1.0.0)··················
        +···▟██▙█▙█▟LocalStack·AWS·Emulator········
        +·····▀▛▀▛▀·localhost.localstack.cloud:4566



        You can run 'go test . -update' to automatically update testdata/header.golden to the new expected value.'
```

PR #68, opened 2026-03-03
Remove left indent from messages in TUI

gtsiolis:
See DES-152 for more context.

** Ignore the missing information in the header, will fix in a follow up PR, logout does not need the header. 👾

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-04 at 00 38 21 | Screenshot 2026-03-04 at 00 37 50 |

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -26,6 +26,7 @@ type styledLine struct {
 	text      string
 	highlight bool
 	secondary bool
+	noIndent  bool
 }
 
 type App struct {
@@ -164,7 +165,8 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		a.spinner, _ = a.spinner.Stop()
 		return a, nil
 	case output.MessageEvent:
-		line := styledLine{text: components.RenderMessage(msg)}
+		noIndent := msg.Severity == output.SeveritySuccess || msg.Severity == output.SeverityNote || msg.Severity == output.SeverityWarning
+		line := styledLine{text: components.RenderMessage(msg), noIndent: noIndent}
 		if a.spinner.PendingStop() {
 			a.bufferedLines = append(a.bufferedLines, line)
 		} else {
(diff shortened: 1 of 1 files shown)


internal/ui/app.go
```diff
@@ -24,6 +24,7 @@ type styledLine struct {
 	text      string
 	highlight bool
 	secondary bool
+	noIndent  bool
```
carole-lavillonniere: suggestion: The change makes sense to me, but I would phrase the bool the other way around: there is a double-negative feel and the zero value `(false = indented`) is the minority case (only `SeverityInfo` gets indented). Since Go's default value for bool is false, it's cleaner to flip to indent bool so the default means "no indent".

carole-lavillonniere (approved):
Just one suggestion to make the code simpler and easier to read, otherwise it looks good!

silv-io:
I'm not sure why we had that indent in the first place 🤔

carole-lavillonniere:
> I'm not sure why we had that indent in the first place 🤔

Do we want to keep it for the Info severity @gtsiolis?

gtsiolis:
> I'm not sure why we had that indent in the first place 🤔

@silv-io I think we inherited this from Bubble Tea. 🧋 

> Do we want to keep it for the Info severity @gtsiolis?

@carole-lavillonniere I've removed the indent from _all_ messages for simplicity. Now messages with the "> " prefix or [other symbols](https://github.com/localstack/lstk/pull/72#discussion_r2882975250) we're going to use will stand out on demand. Initially the idea was to have left indent for all messages except for the ones with a symbol in front, but the complexity of keeping track of the cases for now is not worth it. I'd say it's remove left indent and let the symbols use colors and push text to the right to stand out. Something we can revise later. What do you think?

internal/ui/app.go
gtsiolis: This is no longer relevant, removd indent all together for now, see https://github.com/localstack/lstk/pull/68#issuecomment-4026117818.

silv-io:
> Initially the idea was to have left indent for all messages except for the ones with a symbol in front, but the complexity of keeping track of the cases for now is not worth it.

If I get to it I'll think about how we can do that nicely without thinking too much about it :)

silv-io (approved):

gtsiolis:

PR #69, opened 2026-03-03
Add option to skip header

gtsiolis:
See DES-152 for more context.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-04 at 00 52 54 | Screenshot 2026-03-04 at 00 46 54 |

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -38,10 +38,17 @@ type App struct {
 	pendingInput  *output.UserInputRequestEvent
 	err           error
 	quitting      bool
+	hideHeader    bool
 }
 
-func NewApp(version, emulatorName, endpoint string, cancel func()) App {
-	return App{
+type AppOption func(*App)
+
+var WithoutHeader AppOption = func(a *App) {
+	a.hideHeader = true
+}
+
+func NewApp(version, emulatorName, endpoint string, cancel func(), opts ...AppOption) App {
+	app := App{
 		header:       components.NewHeader(version, emulatorName, endpoint),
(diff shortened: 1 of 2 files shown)


internal/ui/app.go
```diff
@@ -38,17 +38,28 @@ type App struct {
...
+	hideHeader    bool
 }
 
-func NewApp(version, emulatorName, endpoint string, cancel func()) App {
-	return App{
+type AppOption func(*App)
+
+var WithoutHeader AppOption = func(a *App) {
+	a.hideHeader = true
+}
```
carole-lavillonniere:
```suggestion
func withoutHeader() AppOption {
     return func(a *App) { a.hideHeader = true }
}
```
We should avoid vars because they could be mutated, especially when exported.
Switched to lower case in order to not export it outside the package.

internal/ui/run_logout.go
```diff
@@ -15,7 +15,8 @@ func RunLogout(parentCtx context.Context) error {
 	_, cancel := context.WithCancel(parentCtx)
 	defer cancel()
 
-	app := NewApp("", "", "", cancel)
+	app := NewApp("", "", "", cancel, WithoutHeader)
```
carole-lavillonniere:
```suggestion
    app := NewApp("", "", "", cancel, withoutHeader())
```

carole-lavillonniere (approved):
Suggested a small clean-up, rest looks good!

gtsiolis:
Thanks for taking a look, @carole-lavillonniere! 👾

Applied changes from suggestions, will merge once checks pass.

PR #72, opened 2026-03-03
Improve container engine error

gtsiolis:
See DES-152 for more context.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-04 at 01 50 01 | Screenshot 2026-03-04 at 01 48 36 |

### DIFF
--- a/internal/ui/components/error_display.go
+++ b/internal/ui/components/error_display.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/localstack/lstk/internal/output"
 	"github.com/localstack/lstk/internal/ui/styles"
+	"github.com/localstack/lstk/internal/ui/wrap"
 )
 
 type ErrorDisplay struct {
@@ -39,7 +40,7 @@ func (e ErrorDisplay) View(maxWidth int) string {
 	if e.event.Summary != "" {
 		prefix := "> "
 		summaryWidth := maxWidth - len(prefix)
-		lines := softWrap(e.event.Summary, summaryWidth)
+		lines := wrap.SoftWrap(e.event.Summary, summaryWidth)
 		for i, line := range lines {
 			if i == 0 {
 				sb.WriteString(styles.Secondary.Render(prefix))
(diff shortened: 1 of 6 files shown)


internal/ui/components/error_display.go
```diff
@@ -50,12 +59,52 @@ func (e ErrorDisplay) View() string {
...
+			}
 			sb.WriteString("\n")
 		}
 	}
 
 	return sb.String()
 }
+
+// softWrap breaks text into lines at word boundaries, falling back to hard
+// breaks for words longer than maxWidth.
```
carole-lavillonniere: > falling back to hard breaks for words longer than maxWidth.

issue: Is this really implemented?

internal/ui/components/error_display.go
```diff
@@ -50,12 +59,52 @@ func (e ErrorDisplay) View() string {
...
+			current.WriteByte(' ')
+			current.WriteString(word)
+		}
+	}
+	if current.Len() > 0 {
+		lines = append(lines, current.String())
+	}
+
+	return lines
+}
```
carole-lavillonniere: suggestion: Can we move `softWrap` together with `hardWrap` in `internal/ui/app.go` to a dedicated `internal/ui/wrap.go` and unit-test them?

internal/ui/components/error_display.go
```diff
@@ -50,12 +59,52 @@ func (e ErrorDisplay) View() string {
 
 	if len(e.event.Actions) > 0 {
 		sb.WriteString("\n")
-		for _, action := range e.event.Actions {
-			sb.WriteString(styles.ErrorAction.Render("⇒ " + action.Label + " "))
-			sb.WriteString(styles.Message.Bold(true).Render(action.Value))
+		for i, action := range e.event.Actions {
+			if i > 0 {
+				sb.WriteString(styles.SecondaryMessage.Render("⇒ " + action.Label + " " + action.Value))
```
silv-io, line 65: nit: I know it was already in there like this, but I think it would be nice if we defined the symbols we're using somewhere central. Maybe as a const in ui/styles/symbols.go.

```go
const (
      ArrowRight = "→"
      ArrowBold  = "⇒"
      Check      = "✓"
      Cross      = "✗"
)
```

Something like this?

Could also just be something for a follow-up PR

silv-io: Could also allow us to add an option for using nerd fonts in the future :)

internal/ui/components/error_display.go
gtsiolis: Moved in 46f368e. ⏩

internal/ui/components/error_display.go
gtsiolis: Now it should work, added in 46f368e. 😅

internal/ui/components/error_display.go
gtsiolis: The current symbols were chosen for broad terminal compatibility. Just like with design systems, this is a living organism, and symbols may be added or changed later. But agreed it'd be a nice follow up, keeping track inm DES-156.

internal/ui/wrap/wrap.go
```diff
@@ -0,0 +1,80 @@
...
+				}
+			}
+			continue
+		}
+		if current.Len() == 0 {
+			current.WriteString(word)
+			continue
+		}
+		if current.Len()+1+len(runes) > maxWidth {
+			lines = append(lines, current.String())
```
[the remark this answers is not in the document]
gtsiolis: Appended 9723f80.

carole-lavillonniere (approved):

silv-io (approved):

gtsiolis:
Thanks @carole-lavillonniere for taking another look, and @silv-io for merging! 🏀

PR #74, opened 2026-03-04
Use output.Sink in stop command

carole-lavillonniere:
Replaces the `onProgress func(string)` callback in `container.Stop` with an `output.Sink`, consistent with the pattern used in `container.Start` and the rest of the codebase

image


carole-lavillonniere:
Closing since @gtsiolis did the same work here https://github.com/localstack/lstk/pull/71

PR #79, opened 2026-03-05
Support named environments in lstk config.toml

carole-lavillonniere:
Closes DRG-550.

- Adds top-level `[env.*]` sections to the config, where each named environment defines a set of `KEY=value` variables
- Adds `env` field to `[[containers]]` to reference which named environments to apply
- Resolves named environments into `KEY=value` pairs at start time, uppercasing keys to work around Viper's internal key lowercasing

## Example config:

```toml
[[containers]]
type = "aws"
tag = "latest"
port = "4566"
env = ["prod", "debug"]

[env.prod]
LOCALSTACK_HOST = "localstack.cloud"

[env.debug]
LS_LOG = "trace"
DEBUG = "1"
```
## TODO
Add an integration test once we are able to override the config path through a CLI flag (subsequent PR https://github.com/localstack/lstk/pull/80)

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -54,7 +54,11 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, platformCl
 			return err
 		}
 
-		env := append(c.Env, "LOCALSTACK_AUTH_TOKEN="+token)
+		resolvedEnv, err := c.ResolvedEnv(cfg.Env)
+		if err != nil {
+			return err
+		}
+		env := append(resolvedEnv, "LOCALSTACK_AUTH_TOKEN="+token)
 		containers[i] = runtime.ContainerConfig{
 			Image:       image,
 			Name:        c.Name(),
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -9,7 +9,8 @@ import (
 )
 
 type Config struct {
-	Containers []ContainerConfig `mapstructure:"containers"`
+	Containers []ContainerConfig            `mapstructure:"containers"`
+	Env        map[string]map[string]string `mapstructure:"env"`
 }
 
 func setDefaults() {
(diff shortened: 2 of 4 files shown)


internal/container/start.go
```diff
@@ -54,7 +54,11 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, platformCl
 			return err
 		}
 
-		env := append(c.Env, "LOCALSTACK_AUTH_TOKEN="+token)
+		resolvedEnv, err := c.ResolvedEnv(cfg.Env)
+		if err != nil {
+			return err
+		}
+		env := append(resolvedEnv, "LOCALSTACK_AUTH_TOKEN="+token)
```
anisaoshafi, line 61: This means that the stored token will have precedence instead of the one in the resolvedEnv, right? 
Could it error if it's also provided in the env config?

anisaoshafi (approved):
LGTM! 🚀

silv-io (approved):

internal/container/start.go
carole-lavillonniere: Sorry I merged too quick without addressing your comment!
> This means that the stored token will have precedence instead of the one in the resolvedEnv, right?

Yes! If the auth token is set in the config at the moment it would be silently ignored, it's not supported.
I don't think it makes too much sense to support it. The two standard use cases should be:
- login through web; auth token goes in keyring
- set auth token in env var on CI (not in config, otherwise it'd be version-controlled which is not secure)

What we could do though is add a warning if a user tries to set the auth token in the config, what do you think? Something like:
```
LOCALSTACK_AUTH_TOKEN set in env config is ignored;
use the LOCALSTACK_AUTH_TOKEN environment variable or run 'lstk login'

```

anisaoshafi: You're right, this shouldn't happen typically.
That warning doesn't hurt, but I don't think we need it.

PR #80, opened 2026-03-05
Add --config flag to override config file location

carole-lavillonniere:
- Adds a `--config` flag, allowing users to specify a custom config file path (e.g. `lstk --config /path/to/config.toml start`)
- This allows integration-testing the changes of PR https://github.com/localstack/lstk/pull/79

### DIFF
--- a/test/integration/license_test.go
+++ b/test/integration/license_test.go
@@ -6,9 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
-	"os/exec"
 	"testing"
-	"time"
 
 	"github.com/docker/docker/api/types/container"
 	"github.com/localstack/lstk/test/integration/env"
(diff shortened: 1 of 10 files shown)


test/integration/license_test.go
```diff
@@ -88,15 +85,11 @@ func TestLicenseValidationFailure(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
 
-	cmd := exec.CommandContext(ctx, binaryPath(), "start")
-	cmd.Env = env.With(env.APIEndpoint, mockServer.URL).With(env.AuthToken, "test-token-for-license-validation")
-	output, err := cmd.CombinedOutput()
-
+	_, stderr, err := runLstk(t, "", env.With(env.APIEndpoint, mockServer.URL).With(env.AuthToken, "test-token-for-license-validation"), "start")
```
carole-lavillonniere: We now use `runLstk` in each integration test. Previously it was only used in` config_test.go`

silv-io (approved):
LGTM! I wonder if we'll need that for the tests launched in the pty as well?

carole-lavillonniere:
> LGTM! I wonder if we'll need that for the tests launched in the pty as well?

Yes could be at some point

PR #81, opened 2026-03-05
Display error with TUI and suppress duplicate error

carole-lavillonniere:
Output of `lstk stop` for container not running:

### Before
image

### After
image

- Show errors in the TUI error display
- Suppress duplicate plain-text output
- Cobra's `SilenceErrors = true` stops from printing `RunE` errors itself.
- Cobras' `SilenceUsage = true` stops from dumping the full usage block on every runtime error (e.g. "container not running"). Flag parsing errors are caught by cobra before `RunE` is reached, so usage suppression does not affect those.


anisaoshafi (approved):
Nice! 🧹

PR #82, opened 2026-03-06
Collapse image pull progress into loading indicator

silv-io:
## Motivation

When pulling a Docker image, the CLI printed every individual layer progress line from Docker (downloading, extracting, etc. for each layer). This flooded the terminal with dozens of rapidly updating lines, making the output noisy and hard to follow.

## Changes

- Added a new `PullProgress` Bubble Tea component backed by `bubbles/progress` that aggregates pull progress across all Docker layers into a single animated progress bar. Shows the dominant phase (Downloading/Extracting), completed layer count, and overall byte progress with a Nimbo-themed gradient.
- The pull phase now shows a spinner ("Pulling image...") with the aggregated progress bar rendered below it, replacing the per-layer log spam.
- Pull errors now emit a styled `ErrorEvent` instead of a raw error string, consistent with how other failures (e.g., Docker not available) are displayed.
- Non-interactive output suppresses per-layer progress events entirely, showing only "Pulling..." and "Pulled" lines.

## Tests

- Unit tests for the `PullProgress` component covering aggregation, layer counting, phase detection, visibility, and edge cases.
- Unit tests for the TUI app verifying pull progress shows/hides on the correct status events and progress events update the component without appending lines.
- Updated existing plain sink tests to match the new suppression behavior.

## Screenshots

image

image

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -59,6 +59,8 @@ func Execute(ctx context.Context) error {
 	defer tel.Close()
 
 	root := NewRootCmd(cfg, tel)
+	root.SilenceErrors = true
+	root.SilenceUsage = true
 
 	if err := root.ExecuteContext(ctx); err != nil {
 		if !output.IsSilent(err) {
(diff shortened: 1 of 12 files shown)


anisaoshafi (approved):
LGTM!
Thanks for cleaning up that noise, and adding many tests 👏🏼 
image
image

PR #90, opened 2026-03-09
Make pulling output use primary colors

gtsiolis:
See DES-152 for more context.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-09 at 21 21 21 | Screenshot 2026-03-09 at 21 19 25 |

### DIFF
--- a/internal/ui/components/pull_progress.go
+++ b/internal/ui/components/pull_progress.go
@@ -115,7 +115,7 @@ func (p PullProgress) View() string {
 	// Fixed-width: "Downloading" is the longest phase (11 chars), counts padded to "XX/XX"
 	label := fmt.Sprintf("%-11s %2d/%-2d layers", phase, done, total)
 	return fmt.Sprintf("  %s  %s",
-		label,
+		styles.Secondary.Render(label),
 		p.bar.View(),
 	)
 }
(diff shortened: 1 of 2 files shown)


silv-io (approved):
Yep, makes way more sense like this!

This will change all other occurrences of the spinner, though (login, logout, etc.)

internal/ui/components/pull_progress.go
```diff
@@ -115,7 +115,7 @@ func (p PullProgress) View() string {
 	// Fixed-width: "Downloading" is the longest phase (11 chars), counts padded to "XX/XX"
 	label := fmt.Sprintf("%-11s %2d/%-2d layers", phase, done, total)
 	return fmt.Sprintf("  %s  %s",
-		label,
+		styles.Secondary.Render(label),
```
silv-io, line 118: Maybe we want to use the `Secondary` style for the percentage as well? We can do that in the bubbles.progress styles when creating it. WDYT?

gtsiolis: Heads-up: I will get back to this later today or this week. Holding back because I kinda wanted to keep the Logging out output as secondary information.

gtsiolis: Going to close for now to keep logging out, and waiting for authentication, etc messages in secondary color.

Thanks for taking a look, @silv-io! 🙏

PR #91, opened 2026-03-09
Reduce header left padding

gtsiolis:
See DES-152 for more context.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-09 at 21 27 25 | Screenshot 2026-03-09 at 21 26 22 |


silv-io (approved):
I personally like how the larger padding looks more, but reducing it allows us to put more content in the header so let's go for it :)

gtsiolis:

PR #92, opened 2026-03-10
Set up LocalStack AWS profile

carole-lavillonniere:
Auto-configure AWS credentials for new users to immediately execute AWS CLI commands against LocalStack.
Closes DRG-565

- Adds `internal/awsconfig` package to manage the LocalStack AWS profile in `~/.aws/config` and `~/.aws/credentials`
- Adds `internal/endpoint` package with DNS-aware host resolution (falls back to `127.0.0.1` when DNS rebind protection blocks `localhost.localstack.cloud`). Can be overwritten with env var `LOCALSTACK_HOST` for consistency with container
- Post-start hook mechanism in `container/start.go` to initiate the profile setup automatically on `lstk start`: prompts in interactive mode, emits a note in non-interactive mode
- Profile content validation: detects stale/incorrect profiles and prompts to update only the files that need it, not both unconditionally
- Prompt scope is specific: tells the user exactly which files (`~/.aws/config` and/or `~/.aws/credentials`) will be modified
image


anisaoshafi (approved):
LGTM, my manual tests worked as expected 🎉

internal/endpoint/endpoint.go
```diff
@@ -0,0 +1,27 @@
...
+// If override is non-empty it is returned as-is. Otherwise a DNS check is performed;
+// if Hostname does not resolve to 127.0.0.1 (e.g. DNS rebind protection is active),
+// it falls back to 127.0.0.1 directly.
+func ResolveHost(port, override string) (host string, dnsOK bool) {
+	if override != "" {
+		return override, true
+	}
+	// Use a "test." subdomain: *.localhost.localstack.cloud has wildcard DNS that resolves
+	// to 127.0.0.1, so any subdomain works as a probe without hitting the actual service.
+	addrs, err := net.LookupHost("test." + Hostname)
```
anisaoshafi, line 17: small FYI: I had endpoint_url `http://127.0.0.1:4566` in my profile, but it asked me to override the profile (with `http://localhost.localstack.cloud:4566`), if I skip it it will keep asking me each time I start lstk again.
I was wondering if we want to prompt the user to override their profile if they have `http://127.0.0.1:4566` or `http://localhost` which also work fine (if I remember correctly localstack toolkit doesn't prompt in this case).
I'm fine with how this looks now, no need to overcomplicate the code by adding another condition.

cmd/root.go
```diff
@@ -82,9 +82,9 @@ func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *teleme
 
 	platformClient := api.NewPlatformClient(cfg.APIEndpoint)
 	if isInteractiveMode(cfg) {
-		return ui.Run(ctx, rt, version.Version(), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL)
+		return ui.Run(ctx, rt, version.Version(), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL, cfg.LocalStackHost)
 	}
-	return container.Start(ctx, rt, output.NewPlainSink(os.Stdout), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL, false)
+	return container.Start(ctx, rt, output.NewPlainSink(os.Stdout), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL, false, cfg.LocalStackHost)
 }
```
silv-io, line 94: thought: Both of these have quite a long list of parameters now. Is there a way to do this more nicely? 🤔

internal/awsconfig/awsconfig.go
```diff
@@ -0,0 +1,236 @@
...
+// In non-interactive mode, emits a note instead of prompting.
+func Setup(ctx context.Context, sink output.Sink, interactive bool, port string, localStackHost string) error {
+	configPath, credsPath, err := awsPaths()
+	if err != nil {
+		output.EmitWarning(sink, fmt.Sprintf("could not determine AWS config paths: %v", err))
+		return nil
+	}
+
+	var dnsOK bool
+	localStackHost, dnsOK = endpoint.ResolveHost(port, localStackHost)
```
silv-io: nit: we're resolving this also in the ui.Run. Maybe we can unify that somehow?

silv-io (approved):
LGTM!

internal/awsconfig/awsconfig.go
```diff
@@ -0,0 +1,236 @@
...
+			}
+		}
+		if status.credsNeeded {
+			if err := writeCredsProfile(credsPath); err != nil {
+				output.EmitWarning(sink, fmt.Sprintf("could not update ~/.aws/credentials: %v", err))
+				return nil
+			}
+		}
+		output.EmitSuccess(sink, fmt.Sprintf("LocalStack AWS profile written to %s", files))
+		output.EmitNote(sink, fmt.Sprintf("Try: aws s3 mb s3://test --profile %s", profileName))
```
gtsiolis, line 249: **praise**: YASSS! ❤️ This will resolve DRG-363.

internal/endpoint/endpoint.go
carole-lavillonniere: Thanks for testing, and good point! I updated the logic now so it does not overwrite in those cases.

cmd/root.go
carole-lavillonniere: True! Introduced:
```
type StartOptions struct {
	PlatformClient   api.PlatformAPI
	AuthToken        string
	ForceFileKeyring bool
	WebAppURL        string
	LocalStackHost   string
}
```
These are all the user options, they are passed in each case: with or without TUI.

internal/awsconfig/awsconfig.go
carole-lavillonniere: True, done :heavy_check_mark:

PR #93, opened 2026-03-10
Pressing Enter confirms prompts with an uppercase default

carole-lavillonniere:
- Case-insensitive key matching for input prompts (e.g. pressing Y now works, not just y)
- Pressing Enter selects the option with an uppercase label as the default — no need to explicitly add {Key: "enter"} to every prompt definition; just capitalize the label (e.g. {Key: "y", Label: "Y"})

Example: Pressing Enter or y here will confirm as well
image

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -94,47 +94,12 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return a, tea.Quit
 		}
 		if a.pendingInput != nil {
-			// "any" option: any keypress resolves the prompt
-			for _, opt := range a.pendingInput.Options {
-				if opt.Key == "any" {
-					a.lines = appendLine(a.lines, styledLine{text: formatResolvedInput(*a.pendingInput, "any")})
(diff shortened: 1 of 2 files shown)


internal/ui/app.go
```diff
@@ -113,10 +114,20 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 						return a, responseCmd
 					}
 				}
+				// fall back to the option with an uppercase label (conventional default)
+				for _, opt := range a.pendingInput.Options {
```
anisaoshafi: Do we need to iterate twice on the `pendingInput.Options` (line 107), or could we move the new if block up there? 
Seems like most of the things done in these two blocks are similar.

anisaoshafi (approved):
LGTM 🙌🏼  just added one question

internal/ui/app.go
carole-lavillonniere: Good catch, there were 4 identical loops in there :see_no_evil: 
Extracted to a helper func with unit tests: 6d36242638e5016626af09d5f7adb577fd6071a1

PR #94, opened 2026-03-10
fix: prevent silent errors from overwriting TUI error display

carole-lavillonniere:
Fix regression and bring back change from https://github.com/localstack/lstk/pull/72

| Before   |      After      | 
|----------|:-------------:|
| before | after |


gtsiolis (approved):
UX works as expected, thanks for fixing this, @carole-lavillonniere! 🏀

silv-io (approved):
Great catch, thank you for the fix!

PR #95, opened 2026-03-10
Add update command

silv-io:
Adds `lstk update` to self-update the CLI, respecting the original installation method:

- **Homebrew**: detects Caskroom path, runs `brew upgrade localstack/tap/lstk`
- **npm**: detects `node_modules` path, runs `npm install -g` (global) or `npm install` from project dir (local)
- **Binary**: downloads matching archive from GitHub releases and replaces in place
- `--check` flag to only check for updates without applying -> will enable update notifications later :) 

Also generalizes `ContainerLogLineEvent` → `LogLineEvent` with a `Source` field, so brew/npm subprocess output streams through the TUI with a `brew | ...` / `npm | ...` prefix.

### Tests

- Unit tests for install method detection (`classifyPath`) and npm project dir resolution
- Integration tests for all three update paths (binary in-place, Homebrew, npm local install)

[FLC-413](https://linear.app/localstack/issue/FLC-413/add-update-command)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,11 +98,16 @@ jobs:
             go.sum
             test/integration/go.sum
 
+      - name: Install lstk via Homebrew
+        if: matrix.os == 'macos-latest'
+        run: brew install localstack/tap/lstk
+
       - name: Run integration tests
         run: make test-integration
         env:
           CREATE_JUNIT_REPORT: "true"
           LOCALSTACK_AUTH_TOKEN: ${{ secrets.LOCALSTACK_AUTH_TOKEN }}
+          LSTK_TEST_HOMEBREW: ${{ matrix.os == 'macos-latest' && '1' || '0' }}
 
       - name: Upload test results
         uses: actions/upload-artifact@v7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,6 +108,7 @@ jobs:
           CREATE_JUNIT_REPORT: "true"
           LOCALSTACK_AUTH_TOKEN: ${{ secrets.LOCALSTACK_AUTH_TOKEN }}
           LSTK_TEST_HOMEBREW: ${{ matrix.os == 'macos-latest' && '1' || '0' }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Upload test results
         uses: actions/upload-artifact@v7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,6 +98,8 @@ jobs:
             go.sum
             test/integration/go.sum
 
+      # Install lstk via Homebrew so the Homebrew update integration test
+      # (TestUpdateHomebrew) can locate a real Caskroom binary path.
       - name: Install lstk via Homebrew
         if: matrix.os == 'macos-latest'
         run: brew install localstack/tap/lstk
@@ -108,7 +110,7 @@ jobs:
           CREATE_JUNIT_REPORT: "true"
           LOCALSTACK_AUTH_TOKEN: ${{ secrets.LOCALSTACK_AUTH_TOKEN }}
           LSTK_TEST_HOMEBREW: ${{ matrix.os == 'macos-latest' && '1' || '0' }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          LSTK_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Upload test results
         uses: actions/upload-artifact@v7
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -1,0 +1,31 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/localstack/lstk/internal/env"
+	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/ui"
+	"github.com/localstack/lstk/internal/update"
+	"github.com/spf13/cobra"
+)
+
+func newUpdateCmd(cfg *env.Env) *cobra.Command {
+	var checkOnly bool
+
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update lstk to the latest version",
+		Long:  "Check for and apply updates to the lstk CLI. Respects the original installation method (Homebrew, npm, or direct binary).",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if isInteractiveMode(cfg) {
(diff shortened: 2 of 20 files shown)


internal/output/events.go
```diff
@@ -103,8 +103,9 @@ type UserInputRequestEvent struct {
 	ResponseCh chan<- InputResponse
 }
 
-type ContainerLogLineEvent struct {
-	Line string
+type LogLineEvent struct {
+	Source string // e.g., "container", "brew", "npm"
+	Line   string
```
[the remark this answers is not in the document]
silv-io: intentional

internal/update/github.go
```diff
@@ -0,0 +1,70 @@
...
+	"os"
+	"path/filepath"
+	goruntime "runtime"
+)
+
+func updateBinary(ctx context.Context, tag string) error {
+	ver := normalizeVersion(tag)
+	assetName := buildAssetName(ver, goruntime.GOOS, goruntime.GOARCH)
+
+	downloadURL := fmt.Sprintf("https://github.com/%s/releases/download/%s/%s", githubRepo, tag, assetName)
```
silv-io, line 64: I think we should replace this with a location we control in the future

internal/container/logs.go
```diff
@@ -33,7 +33,7 @@ func Logs(ctx context.Context, rt runtime.Runtime, sink output.Sink, follow bool
 
 	scanner := bufio.NewScanner(pr)
 	for scanner.Scan() {
-		output.EmitContainerLogLine(sink, scanner.Text())
+		output.EmitLogLine(sink, "container", scanner.Text())
```
carole-lavillonniere: nit: I know this was not changed in this PR, but do we really want to show "container" or maybe "emulator"? (Also, atm these logs are not displayed using the TUI so we never show the Source for now)

internal/update/github.go
carole-lavillonniere: Do you mean that we should host the binaries ourselves? Why?

internal/update/update.go
```diff
@@ -0,0 +1,121 @@
...
+	"strings"
+	"sync"
+
+	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/version"
+)
+
+// Check reports whether a newer version is available.
+// Returns the latest version string and true if an update is available.
+func Check(ctx context.Context, sink output.Sink) (string, bool, error) {
```
carole-lavillonniere: question: can we use this for any command to warn when a newer version is available and ask user if they want to install? (May belong to another issue/PR)
Are we worried that Github might end up throttling our users' requests if we check too much?

cmd/update.go
```diff
@@ -0,0 +1,32 @@
...
+				return ui.RunUpdate(cmd.Context(), checkOnly)
+			}
+			return update.Update(cmd.Context(), output.NewPlainSink(os.Stdout), checkOnly)
+		},
+	}
+
+	cmd.Flags().BoolVar(&checkOnly, "check", false, "Only check for updates without applying them")
+
+	return cmd
+}
```
carole-lavillonniere, line 32: nit: `CLAUDE.md` needs an update to document the new command

test/integration/update_test.go
```diff
@@ -0,0 +1,205 @@
...
+	updateOut, err := updateCmd.CombinedOutput()
+	updateStr := string(updateOut)
+	require.NoError(t, err, "lstk update failed: %s", updateStr)
+	assert.Contains(t, updateStr, "Homebrew", "should detect Homebrew install")
+	assert.Contains(t, updateStr, "brew upgrade", "should mention brew upgrade")
+}
+
+func npmPlatformPackage() string {
+	return "lstk_" + runtime.GOOS + "_" + runtime.GOARCH
+}
```
carole-lavillonniere, line 205: praise: Beautiful tests 😍

internal/update/github.go
```diff
@@ -0,0 +1,113 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Accept", "application/vnd.github+json")
+	// TODO: this is in here temporarily while we need github for binary downloads. that's why it's not using the Env
+	if token := os.Getenv("GITHUB_TOKEN"); token != "" {
+		req.Header.Set("Authorization", "Bearer "+token)
+	}
+	return http.DefaultClient.Do(req)
```
carole-lavillonniere, line 38: question: isn't it likely to stay like this for a while though? I would move to env.go

.github/workflows/ci.yml
```diff
@@ -98,11 +98,17 @@ jobs:
             go.sum
             test/integration/go.sum
 
+      - name: Install lstk via Homebrew
+        if: matrix.os == 'macos-latest'
+        run: brew install localstack/tap/lstk
```
carole-lavillonniere, line 105: nit: can we add a comment to explain why the new step is needed?

carole-lavillonniere (approved):
👏 great work! Just a couple of questions/nits

internal/container/logs.go
silv-io: I was thinking of maybe using this source field also to differentiate how container logs are printed vs. npm and homebrew. So in the TUI it then wouldn't do `container | LOGS` but something else entirely maybe?

internal/update/github.go
silv-io: GitHub rate limits and in general GitHub reliability.

We could instead have an S3 bucket that we manage ourselves

internal/update/update.go
silv-io: Yes this was my intention with it! 

And yes, I'm worried about that which is why our own hosting location would make sense

test/integration/update_test.go
silv-io: yes, I tested it all manually and then I figured it must be possible to test that automatically as well. Only problem is that it relies on downloads so they might be flaky -> will have to see.

internal/update/github.go
silv-io: I hope it isn't. But you're right - nothing is more permanent than a temporary solution :D 

will change

internal/container/logs.go
carole-lavillonniere: In this case we should probably make them constants?

internal/update/github.go
carole-lavillonniere: Makes sense! Can you take care of adding an issue for changing the hosting?

anisaoshafi (approved):
Nice piece of work, good job Silvio 👏🏼  

The code looks good to me. I look forward to testing this once we release a new version

internal/update/github.go
silv-io: Done, ENG-501 is the issue

PR #96, opened 2026-03-10
Warn unsupported multiple containers of the same type

anisaoshafi:
Show a warning to users that have multiple containers of the same type defined in their config to avoid confusion when they start lstk:
image

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -40,7 +40,7 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, platformCl
 	}
 
 	if hasDuplicateContainerTypes(cfg.Containers) {
-		output.EmitWarning(sink, "multiple emulators of the same type are defined in your config; this setup is not supported yet")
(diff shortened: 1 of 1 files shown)


internal/container/start.go
```diff
@@ -39,6 +39,10 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, platformCl
 		return fmt.Errorf("failed to get config: %w", err)
 	}
 
+	if hasDuplicateContainerTypes(cfg.Containers) {
+		output.EmitWarning(sink, "multiple emulators of the same type are defined in your config; this setup is not supported yet")
```
silv-io: nit: I guess we didn't establish a wording style guide yet, but I think other messages start capitalized. non blocking but something to think about :D

silv-io (approved):
LGTM! Nice and clean change

internal/container/start.go
anisaoshafi: Good point, changed it. I'll remember that for next time 🙌🏼

PR #97, opened 2026-03-10
Improve port-in-use error

gtsiolis:
| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-10 at 21 14 37 | Screenshot 2026-03-10 at 21 09 37 |


silv-io (approved):
Great fix! LGTM

Only thing is the tests. I think it's because we now have different output in plain vs TUI mode?

gtsiolis:

PR #98, opened 2026-03-10
Improve input prompt for AWS profile

gtsiolis:
Some improvements on the input prompt for AWS profile for https://github.com/localstack/lstk/pull/92.

Open to rebasing and merging afterwards.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-10 at 21 42 20 | Screenshot 2026-03-10 at 21 38 14 |

### DIFF
--- a/internal/ui/components/input_prompt.go
+++ b/internal/ui/components/input_prompt.go
@@ -47,12 +47,7 @@ func (p InputPrompt) View() string {
 	// "?" prefix in secondary color
 	sb.WriteString(styles.Secondary.Render("? "))
 
-	// Style trailing "?" with secondary color
-	if before, found := strings.CutSuffix(firstLine, "?"); found {
-		sb.WriteString(styles.Message.Render(before + "?"))
-	} else {
-		sb.WriteString(styles.Message.Render(firstLine))
-	}
+	sb.WriteString(styles.Message.Render(firstLine))
 
 	// Style option labels with secondary color
 	labels := make([]string, 0, len(p.options))
--- a/internal/ui/components/input_prompt.go
+++ b/internal/ui/components/input_prompt.go
@@ -1,6 +1,9 @@
 package components
 
 import (
+	"fmt"
+	"strings"
+
 	"github.com/localstack/lstk/internal/output"
 	"github.com/localstack/lstk/internal/ui/styles"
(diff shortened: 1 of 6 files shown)


internal/ui/components/input_prompt.go
```diff
@@ -36,5 +39,38 @@ func (p InputPrompt) View() string {
...
-	return styles.SecondaryMessage.Render(output.FormatPrompt(p.prompt, p.options))
+	lines := strings.Split(p.prompt, "\n")
+	firstLine := lines[0]
+
+	var sb strings.Builder
+
+	// "?" prefix in secondary color
+	sb.WriteString(styles.Secondary.Render("? "))
+
+	// Style trailing "?" with secondary color
```
carole-lavillonniere: issue: we never seem to be styling the "trailing ? with secondary color" like the comment says, so not sure what is the intent of this code

internal/ui/components/input_prompt.go
```diff
@@ -36,5 +39,38 @@ func (p InputPrompt) View() string {
...
+	} else if len(labels) > 1 {
+		sb.WriteString(styles.Secondary.Render(fmt.Sprintf(" [%s]", strings.Join(labels, "/"))))
+	}
+
+	if len(lines) > 1 {
+		sb.WriteString("\n")
+		sb.WriteString(styles.SecondaryMessage.Render(strings.Join(lines[1:], "\n")))
+	}
+
+	return sb.String()
```
carole-lavillonniere, line 60: suggestion: it looks like the logic to format the labels is duplicated in `plain_format.go:62-89`. Can we introduce a new helper `FormatPromptLabels` that's used in both places?

internal/ui/components/input_prompt.go
gtsiolis: Oops, seems like a leftover as I was trying to have only the the prefix "?" use secondary color.

Nice catch! ⚾ Changed in 1790406.

internal/ui/components/input_prompt.go
gtsiolis: Made an attempt for this in fa61ef1.

carole-lavillonniere (approved):

PR #99, opened 2026-03-11
Fix keyring tests

carole-lavillonniere:
Integrations tests relying on keyring on linux+gnome keyring started failing after this PR was merged https://github.com/localstack/lstk/pull/87
We need to set the default wallet in tests as well.

### DIFF
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -97,8 +97,9 @@ func TestMain(m *testing.M) {
 	}
 
 	keyringConfig := keyring.Config{
-		ServiceName: keyringService,
-		FileDir:     filepath.Join(configDir(), keyringFilename),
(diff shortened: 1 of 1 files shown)


anisaoshafi (approved):
LGTM :shipit: 

**question**: Did this fail in the github actions (I can't find an instance of the failure), or you see this when running tests in your system?

silv-io (approved):
LGTM!

carole-lavillonniere:
> LGTM :shipit:
> 
> **question**: Did this fail in the github actions (I can't find an instance of the failure), or you see this when running tests in your system?

Only on my system. We decided to skip those tests on CI on linux, because the runner does not have a keyring by default. We could look into installing one in the future to allow running those tests.

PR #102, opened 2026-03-11
Improve auth flow and append code in auth request URL

gtsiolis:
Fix DES-160 and DES-157.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-11 at 19 02 39 | Screenshot 2026-03-11 at 20 52 29 |


silv-io (commented):
TBH I like it more with the code displayed. It makes it look more actionable? 

What if the link fails somehow and the code isn't copied? What if there is a MITM somehow? I think it should still be easily recognizable what the code is.

gtsiolis:
Good point on the code print, @silv-io! Added the code separately on its own line, but moved it further below using secondary text since with the auto-fill this is no longer an important instruction for users to follow.  ✔️ 

Updated also screenshots in the PR description. ✔️ 

---

Adding an updated plan ahead for more context on the direction for future reference:

1. **Browser auto-open.** We now auto-open the browser. Today the code is not auto-filled yet, so users still land on a form they need to complete manually without clear instructions. The auto-fill from the query param should make this smoother.
2. **Device flow as the current baseline.** For now we default to the device flow. This keeps the behavior consistent and avoids relying on the previous redirect-based approach, which had awkward UX and weaker security semantics, see [previous planning](https://github.com/localstack/localstack-web/pull/2101#issuecomment-3849449650).
3. **Authorization step (planned).** A next improvement would be introducing a proper authorization page where the user explicitly approves access before the token exchange, [DES-115](https://linear.app/localstack/issue/DES-115/authorize-access-before-redirect).
4. **Authentication flow choice.** Once the authorization step described in [3] exists, we can decide whether to default to the browser authorization flow, with device flow as fallback when no browser is detected, or explicitly let users choose the auth method.
5. **Polling is needed.** Regardless of which path we choose for [4], the CLI should poll the backend while waiting for authentication instead of asking the user to press for user input when ready, [PRO-215](https://linear.app/localstack/issue/PRO-215/add-long-polling-for-better-auth-experience-flow).

carole-lavillonniere (approved):
👏

gtsiolis:
Thanks for taking a look, @carole-lavillonniere @silv-io! 🏀

PR #104, opened 2026-03-11
Add new commands in README

gtsiolis:


anisaoshafi (approved):
Thanks George for keeping things up-to-date! 🆕 👏🏼

README.md
```diff
@@ -78,12 +80,27 @@ To see which config file is currently in use:
...
+```
+
 ## Environment Variables
 
 | Variable | Description |
 |---|---|
-| `LOCALSTACK_AUTH_TOKEN` | Auth token; for CI only |
+| `LOCALSTACK_AUTH_TOKEN` | Auth token used for non-interactive runs or to skip browser login |
 | `LOCALSTACK_DISABLE_EVENTS=1` | Disables telemetry event reporting |
+| `LSTK_GITHUB_TOKEN` | GitHub token used by `lstk update` to avoid API rate limits |
```
carole-lavillonniere: question: this is not user facing but only for CI (and only as long as we host the binaries on github, which will change with ENG-501). Not sure we should document it, it might confuse users?

gtsiolis: Good point, removed in 83a81ce.

gtsiolis:
Thanks @anisaoshafi @carole-lavillonniere for taking a look! 🏀

PR #112, opened 2026-03-12
fix: correct port binding for non-default ports

anisaoshafi:
When configuring a non-default port (e.g. 4567), the Docker binding was 4567:4567 instead of 4567:4566, so the health check could not reach localstack and lstk would hang.

Fix port binding for non-default ports so the emulator actually becomes ready.

### DIFF
--- a/internal/runtime/docker.go
+++ b/internal/runtime/docker.go
@@ -107,10 +107,8 @@ func (d *DockerRuntime) PullImage(ctx context.Context, imageName string, progres
 	return nil
 }
 
-const emulatorContainerPort = nat.Port("4566/tcp")
-
 func (d *DockerRuntime) Start(ctx context.Context, config ContainerConfig) (string, error) {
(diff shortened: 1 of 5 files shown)


internal/runtime/docker.go
```diff
@@ -107,10 +107,12 @@ func (d *DockerRuntime) PullImage(ctx context.Context, imageName string, progres
 	return nil
 }
 
+const emulatorContainerPort = nat.Port("4566/tcp")
```
carole-lavillonniere: thought: ultimately, when we support other emulators, we will to configure it per emulator and might want to move this value to `runtime.ContainerConfig`, but we can do it later!

carole-lavillonniere (approved):
👏

silv-io (approved):

anisaoshafi:
@coderabbitai recheck

PR #113, opened 2026-03-13
Use checkmark for success message

gtsiolis:
This will changes success messages rendering from text to a shared checkmark marker. ✔️ 

This makes success output more compact and visually consistent across renderers. 💄 

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-13 at 15 09 24 | Screenshot 2026-03-13 at 15 03 16 |

### DIFF
--- a/internal/output/style.go
+++ b/internal/output/style.go
@@ -1,7 +1,0 @@
-package output
-
-const SuccessColorHex = "#B7C95C"
-
-func SuccessMarkerText() string {
-	return "✔︎"
-}
--- a/internal/output/style.go
+++ b/internal/output/style.go
@@ -1,0 +1,13 @@
+package output
+
+import "fmt"
+
+const SuccessColorHex = "#B7C95C"
+
+func SuccessMarker() string {
+	return fmt.Sprintf("\x1b[38;2;183;201;92m%s\x1b[0m", SuccessMarkerText())
+}
+
+func SuccessMarkerText() string {
+	return "✔︎"
+}
--- a/internal/output/style.go
+++ b/internal/output/style.go
@@ -1,12 +1,6 @@
 package output
 
-import "fmt"
-
 const SuccessColorHex = "#B7C95C"
-
-func SuccessMarker() string {
-	return fmt.Sprintf("\x1b[38;2;183;201;92m%s\x1b[0m", SuccessMarkerText())
-}
 
 func SuccessMarkerText() string {
 	return "✔︎"
--- a/internal/ui/styles/styles.go
+++ b/internal/ui/styles/styles.go
@@ -1,6 +1,9 @@
 package styles
 
-import "github.com/charmbracelet/lipgloss"
+import (
+	"github.com/charmbracelet/lipgloss"
+	"github.com/localstack/lstk/internal/output"
+)
 
 const (
 	NimboDarkColor  = "#3F51C7"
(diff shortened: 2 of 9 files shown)


internal/output/style.go
```diff
@@ -0,0 +1,13 @@
+package output
+
+import "fmt"
+
+const SuccessColorHex = "#B7C95C"
+
+func SuccessMarker() string {
+	return fmt.Sprintf("\x1b[38;2;183;201;92m%s\x1b[0m", SuccessMarkerText())
+}
```
silv-io: The problem with having color in the non-TUI output as well is that it won't respect the `NO_COLOR` environment variable that bubbletea handles. 

In general, I don't see the reason for color output in non-TUI mode. It will just make the output unreadable if the target is unable to render those colors.
This is very likely in non-TTY outputs!

gtsiolis: Good point. I didn't intent to add colors in non-TUI output. Made some changes in 5e1991c.

gtsiolis:
Rebased to keep fresh. 🥗

gtsiolis:
Rebased to keep fresh. 🥗

internal/output/style.go
gtsiolis: Could you take another look when you're free, @silv-io?

internal/output/style.go
```diff
@@ -0,0 +1,7 @@
+package output
+
+const SuccessColorHex = "#B7C95C"
```
silv-io: ~q: Do we need this color still? it's defined in the ui package, no?~

Noticed it's just used in the UI package to define the style. IMO it should be defined there. Colors are only something for the UI and should all be defined in the same place.

internal/output/style.go
silv-io, line 1: I think this can become the `symbols.go` file that we talked about some time ago. 

`SuccessMarkerText()` can then be `SuccessMarker()` maybe?

internal/ui/styles/styles.go
```diff
@@ -39,7 +42,7 @@ var (
 
 	// Message severity styles
 	Success = lipgloss.NewStyle().
-		Foreground(lipgloss.Color("42"))
+		Foreground(lipgloss.Color(output.SuccessColorHex))
```
silv-io: I would define the success color itself here in the styles like all the other colors

silv-io (commented):
LGTM, just some clean up we'll need I think.

All the color stuff should go to ui styles. Then the output "styles" can become the symbols file instead.

gtsiolis:
Thanks, @silv-io! Pushed 8a425d9 to address all three comments above. Please, take another look when you're free.

gtsiolis:
Rebased to keep fresh. 🥗

silv-io:
> Thanks, @silv-io! Pushed [8a425d9](https://github.com/localstack/lstk/commit/8a425d97d451a8ba4432fdbcf392c30ad77578eb) to address all three comments above. Please, take another look when you're free.

@gtsiolis It went a bit backwards now and uses ui code in the plain output. I'll take this over and fix it and merge then if it's ok? :)

gtsiolis:
Absolutely, @silv-io! Leaving this to you. 🏀

silv-io:
Adapted it now, will merge if all tests pass ✔️

silv-io (approved):
LGTM now, merging

gtsiolis:
Woohoo!

PR #114, opened 2026-03-13
List endpoint and web app on start

gtsiolis:
| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-13 at 17 45 19 | Screenshot 2026-03-13 at 17 41 10 |

### DIFF
--- a/internal/ui/components/message.go
+++ b/internal/ui/components/message.go
@@ -16,7 +16,7 @@ func RenderWrappedMessage(e output.MessageEvent, width int) string {
 	prefixText, prefix := messagePrefix(e)
 	if prefixText == "" {
 		style := styles.Message
-		if isDecorativeMessage(e.Text) {
+		if e.Severity == output.SeveritySecondary {
 			style = styles.SecondaryMessage
 		}
 		return style.Render(strings.Join(wrap.SoftWrap(e.Text, width), "\n"))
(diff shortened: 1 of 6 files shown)


internal/update/install_method.go
```diff
@@ -1,9 +1,12 @@
 package update
```
carole-lavillonniere, line 1: issue: was this file accidentally committed?

internal/ui/components/message.go
```diff
@@ -15,6 +17,13 @@ func RenderMessage(e output.MessageEvent) string {
...
+		if isDecorativeMessage(e.Text) {
+			return styles.SecondaryMessage.Render(e.Text)
+		}
 		return styles.Message.Render(e.Text)
 	}
 }
+
+func isDecorativeMessage(text string) bool {
+	return strings.HasPrefix(text, "• ") || strings.HasPrefix(text, "> Tip:")
+}
```
carole-lavillonniere: issue: we should avoid parsing message content to decide how to style it. This creates a dependency between text content and style that is fragile. Can we use a new message type or a new severity instead?

internal/update/install_method.go
gtsiolis: Oops! Good catch! ⚾ Removed.

internal/ui/components/message.go
gtsiolis: ABSOLUTELY! I should have done this already, thanks for the nudge! 😁

carole-lavillonniere (approved):

gtsiolis:
Thanks for taking a look, @carole-lavillonniere! 🏀

PR #115, opened 2026-03-16
Add exit code assertions to integration tests

carole-lavillonniere:
Exit codes are part of the contract and must never change with new versions.
Add explicit assertions so that any accidental change to exit behavior is caught immediately.

Towards DRG-581


anisaoshafi (approved):
LGTM!

PR #118, opened 2026-03-16
Fix agent instructions

gtsiolis:
Cc @carole-lavillonniere @silv-io because https://github.com/localstack/lstk/pull/116


carole-lavillonniere (approved):
Oops, thanks for fixing my mess @gtsiolis

PR #120, opened 2026-03-17
Validate container port at config load time

gtsiolis:
Ports like "abc" or "99999" previously passed through config loading silently and only failed at runtime. 

This will check port is numeric and within 1–65535, etc.

Fix DES-171


anisaoshafi (approved):
Good catch 👏🏼

gtsiolis:
Thanks for taking a look, @anisaoshafi! 🏀

PR #121, opened 2026-03-17
Add timeout to container stop operations

gtsiolis:
Fix DES-172

### DIFF
--- a/internal/container/stop.go
+++ b/internal/container/stop.go
@@ -17,8 +17,8 @@ func Stop(ctx context.Context, rt runtime.Runtime, sink output.Sink, containers 
 		name := c.Name()
 
 		checkCtx, checkCancel := context.WithTimeout(ctx, 5*time.Second)
-		defer checkCancel()
 		running, err := rt.IsRunning(checkCtx, name)
+		checkCancel()
 		if err != nil {
 			return fmt.Errorf("checking %s running: %w", name, err)
 		}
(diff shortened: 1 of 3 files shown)


internal/container/stop.go
```diff
@@ -3,27 +3,34 @@ package container
...
 
-func Stop(ctx context.Context, rt runtime.Runtime, sink output.Sink, containers []config.ContainerConfig) error {
+func Stop(ctx context.Context, rt runtime.Runtime, sink output.Sink, containers []config.ContainerConfig, timeout time.Duration) error {
 	for _, c := range containers {
 		name := c.Name()
-		running, err := rt.IsRunning(ctx, name)
+
+		checkCtx, checkCancel := context.WithTimeout(ctx, 5*time.Second)
+		running, err := rt.IsRunning(checkCtx, name)
+		checkCancel()
```
carole-lavillonniere: issue:
We should always defer calls to a cancel context func. 
So it should become `defer checkCancel() ` and be moved before the `IsRunning` call.

internal/container/stop.go
```diff
@@ -3,27 +3,34 @@ package container
...
 		}
 		if !running {
 			return fmt.Errorf("LocalStack is not running")
 		}
 		output.EmitSpinnerStart(sink, "Stopping LocalStack...")
-		if err := rt.Stop(ctx, name); err != nil {
+		stopCtx, stopCancel := context.WithTimeout(ctx, timeout)
+		if err := rt.Stop(stopCtx, name); err != nil {
 			output.EmitSpinnerStop(sink)
+			stopCancel()
```
carole-lavillonniere: issue: same here. We need only one `defer stopCancel()` and we can move it to right before calling `Stop`

internal/container/stop.go
```diff
@@ -3,27 +3,34 @@ package container
...
+		checkCancel()
 		if err != nil {
 			return fmt.Errorf("checking %s running: %w", name, err)
 		}
 		if !running {
 			return fmt.Errorf("LocalStack is not running")
 		}
 		output.EmitSpinnerStart(sink, "Stopping LocalStack...")
-		if err := rt.Stop(ctx, name); err != nil {
+		stopCtx, stopCancel := context.WithTimeout(ctx, timeout)
```
carole-lavillonniere: question: I don't see the benefit of passing the value of `timeout` all the way down from `cmd/stop.go` line 33 instead of just hardcoding here?

gtsiolis:
Thanks, @carole-lavillonniere! Tried to address all three comments in a610317. 

Could you take another look when you're free?

carole-lavillonniere (approved):

internal/container/stop.go
carole-lavillonniere: I see now why it's problematic to defer in a loop, your solution looks good :+1:

PR #122, opened 2026-03-17
Add volume mount for emulator persistence

silv-io:
Mount a host directory into `/var/lib/localstack` for caching and opt-in persistence (`PERSISTENCE=1`).

## Changes

- Add `Volume` field to `ContainerConfig` and `VolumeDir()` method for resolving the host path
- Default path: `os.UserCacheDir()/lstk/volume/<container-name>` (e.g. `~/Library/Caches/lstk/volume/localstack-aws`)
- Configurable via `volume` field in `[[containers]]` config
- Include `volume` in default config written on first run
- Create directory automatically (`os.MkdirAll`) before container start
- Add bind mount `volumeDir:/var/lib/localstack`

## Tests

- Integration subtest in `TestStartCommandSetsUpContainerCorrectly` verifying the volume bind mount targets `/var/lib/localstack`

## Related

- Closes [FLC-473](https://linear.app/localstack/issue/FLC-473)
- Closes [FLC-474](https://linear.app/localstack/issue/FLC-474)

### DIFF
--- a/internal/config/containers.go
+++ b/internal/config/containers.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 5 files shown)


anisaoshafi (approved):
Looks good 💯

internal/config/containers.go
```diff
@@ -33,13 +35,28 @@ var emulatorHealthPaths = map[EmulatorType]string{
...
+	Volume string       `mapstructure:"volume"`
 	// Env is a list of named environment references defined in the top-level [env.*] config sections.
 	Env []string `mapstructure:"env"`
 }
 
+// VolumeDir returns the host directory to mount into the container for persistence/caching.
+// If Volume is set in the config, it is returned as-is. Otherwise, a default is computed
+// from os.UserCacheDir()/lstk/volume/<container-name>.
+func (c *ContainerConfig) VolumeDir() (string, error) {
+	if c.Volume != "" {
```
carole-lavillonniere, line 50: question: isn't this never empty since we set a default value in `internal/config/config.go` line 27? Maybe let's remove the viper default so we generate the default path here?

carole-lavillonniere (approved):
Looks good, just one question

internal/config/containers.go
silv-io: it can be empty if someone does a custom config and forgets to set it. I wanted it to be set in the default config though so that it's more discoverable. I can also remove it if needed though.

silv-io: I'll remove it for now and then we can revisit later. Maybe discoverability is better solved by just having a big all options example config file.

PR #127, opened 2026-03-17
Track telemetry events

anisaoshafi:
We'll start tracking these events:
- `lstk_command`: generic command execution event with one stable schema (no subtype)
- `lstk_lifecycle`: emulator lifecycle event family, emitted per emulator/container instance with lifecycle subtypes (`start_success`, `stop`, `start_error`)

### Test
To test in local use: `LSTK_ANALYTICS_ENDPOINT=http://localhost:8000/v1/events ./bin/lstk <cmd>`

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -19,22 +19,20 @@ func newLoginCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.
 		Short:   "Manage login",
 		Long:    "Manage login and store credentials in system keyring",
 		PreRunE: initConfig,
-		RunE: commandWithTelemetry("login", tel, func() string {
-			tokenStorage, err := auth.NewTokenStorage(cfg.ForceFileKeyring, logger)
-			if err != nil {
-				return cfg.AuthToken
-			}
-			token, err := tokenStorage.GetAuthToken()
-			if err != nil {
-				return cfg.AuthToken
-			}
-			return token
-		}, func(cmd *cobra.Command, args []string) error {
+		RunE: commandWithTelemetry("login", tel, func(cmd *cobra.Command, args []string) error {
 			if !isInteractiveMode(cfg) {
 				return fmt.Errorf("login requires an interactive terminal")
 			}
 			platformClient := api.NewPlatformClient(cfg.APIEndpoint, logger)
-			return ui.RunLogin(cmd.Context(), version.Version(), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL, logger)
+			if err := ui.RunLogin(cmd.Context(), version.Version(), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL, logger); err != nil {
+				return err
+			}
+			if tokenStorage, err := auth.NewTokenStorage(cfg.ForceFileKeyring, logger); err == nil {
+				if token, err := tokenStorage.GetAuthToken(); err == nil && token != "" {
+					tel.SetAuthToken(token)
+				}
+			}
+			return nil
 		}),
 	}
 }
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/localstack/lstk/internal/api"
+	"github.com/localstack/lstk/internal/auth"
 	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/log"
 	"github.com/localstack/lstk/internal/telemetry"
@@ -18,7 +19,17 @@ func newLoginCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.
 		Short:   "Manage login",
 		Long:    "Manage login and store credentials in system keyring",
 		PreRunE: initConfig,
-		RunE: withCommandTelemetry("login", tel, cfg.AuthToken, func(cmd *cobra.Command, args []string) error {
+		RunE: withCommandTelemetry("login", tel, func() string {
+			tokenStorage, err := auth.NewTokenStorage(cfg.ForceFileKeyring, logger)
+			if err != nil {
+				return cfg.AuthToken
+			}
+			token, err := tokenStorage.GetAuthToken()
+			if err != nil {
+				return cfg.AuthToken
+			}
+			return token
+		}, func(cmd *cobra.Command, args []string) error {
 			if !isInteractiveMode(cfg) {
 				return fmt.Errorf("login requires an interactive terminal")
 			}
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -19,7 +19,7 @@ func newLoginCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.
 		Short:   "Manage login",
 		Long:    "Manage login and store credentials in system keyring",
 		PreRunE: initConfig,
-		RunE: withCommandTelemetry("login", tel, func() string {
+		RunE: commandWithTelemetry("login", tel, func() string {
 			tokenStorage, err := auth.NewTokenStorage(cfg.ForceFileKeyring, logger)
 			if err != nil {
 				return cfg.AuthToken
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -6,23 +6,24 @@ import (
 	"github.com/localstack/lstk/internal/api"
 	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/log"
+	"github.com/localstack/lstk/internal/telemetry"
 	"github.com/localstack/lstk/internal/ui"
 	"github.com/localstack/lstk/internal/version"
 	"github.com/spf13/cobra"
 )
 
-func newLoginCmd(cfg *env.Env, logger log.Logger) *cobra.Command {
+func newLoginCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.Command {
 	return &cobra.Command{
 		Use:     "login",
 		Short:   "Manage login",
 		Long:    "Manage login and store credentials in system keyring",
 		PreRunE: initConfig,
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: withCommandTelemetry("login", tel, cfg.AuthToken, func(cmd *cobra.Command, args []string) error {
 			if !isInteractiveMode(cfg) {
 				return fmt.Errorf("login requires an interactive terminal")
 			}
 			platformClient := api.NewPlatformClient(cfg.APIEndpoint, logger)
 			return ui.RunLogin(cmd.Context(), version.Version(), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL, logger)
-		},
+		}),
 	}
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -146,10 +146,8 @@ func runStart(ctx context.Context, cmdFlags *pflag.FlagSet, rt runtime.Runtime, 
 	return runErr
 }
 
-// commandWithTelemetry wraps a RunE function so that an lstk_command event is
-// emitted after every invocation. Use this for commands that do not emit
-// lstk_lifecycle events (i.e. everything except start/stop, which manage their
-// own telemetry).
+// wraps a RunE function so that an lstk_command event is emitted after every invocation
+// used for commands that do not emit lstk_lifecycle events (i.e. status, logs, config path, etc)
 func commandWithTelemetry(name string, tel *telemetry.Client, fn func(*cobra.Command, []string) error) func(*cobra.Command, []string) error {
 	return func(cmd *cobra.Command, args []string) error {
 		startTime := time.Now()
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -58,7 +58,7 @@ func NewRootCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.C
 		newLogoutCmd(cfg, tel, logger),
 		newStatusCmd(cfg, tel),
 		newLogsCmd(cfg, tel),
-		newConfigCmd(),
+		newConfigCmd(cfg, tel),
 		newUpdateCmd(cfg, tel),
 	)
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -145,6 +145,7 @@ func runStart(cmd *cobra.Command, rt runtime.Runtime, cfg *env.Env, tel *telemet
 		}
 	}
 	tel.Emit(cmd.Context(), "lstk_command", telemetry.ToMap(telemetry.CommandEvent{
+		EventID:     commandEventID,
 		Environment: tel.GetEnvironment(cfg.AuthToken),
 		Parameters:  telemetry.CommandParameters{Command: "start", Flags: flags},
 		Result: telemetry.CommandResult{
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/localstack/lstk/internal/api"
 	"github.com/localstack/lstk/internal/config"
@@ -17,6 +18,7 @@ import (
 	"github.com/localstack/lstk/internal/ui"
 	"github.com/localstack/lstk/internal/version"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 )
 
 func NewRootCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.Command {
@@ -30,7 +32,7 @@ func NewRootCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.C
 			if err != nil {
 				return err
 			}
-			return runStart(cmd.Context(), rt, cfg, tel, logger)
+			return runStart(cmd, rt, cfg, tel, logger)
 		},
 	}
 
@@ -50,13 +52,13 @@ func NewRootCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.C
 
 	root.AddCommand(
 		newStartCmd(cfg, tel, logger),
-		newStopCmd(cfg),
-		newLoginCmd(cfg, logger),
-		newLogoutCmd(cfg, logger),
-		newStatusCmd(cfg),
-		newLogsCmd(cfg),
+		newStopCmd(cfg, tel),
+		newLoginCmd(cfg, tel, logger),
+		newLogoutCmd(cfg, tel, logger),
+		newStatusCmd(cfg, tel),
+		newLogsCmd(cfg, tel),
 		newConfigCmd(),
-		newUpdateCmd(cfg),
+		newUpdateCmd(cfg, tel),
 	)
 
 	return root
@@ -87,9 +89,7 @@ func Execute(ctx context.Context) error {
 	return nil
 }
 
-func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *telemetry.Client, logger log.Logger) error {
-	// TODO: replace map with a typed payload struct once event schema is finalised
-	tel.Emit(ctx, "cli_cmd", map[string]any{"cmd": "lstk start", "params": []string{}})
+func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *telemetry.Client, commandEventID string, logger log.Logger) error {
 
 	appConfig, err := config.Get()
 	if err != nil {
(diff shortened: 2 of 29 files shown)


cmd/login.go
```diff
@@ -6,23 +6,24 @@ import (
...
 		PreRunE: initConfig,
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: withCommandTelemetry("login", tel, cfg.AuthToken, func(cmd *cobra.Command, args []string) error {
 			if !isInteractiveMode(cfg) {
 				return fmt.Errorf("login requires an interactive terminal")
 			}
 			platformClient := api.NewPlatformClient(cfg.APIEndpoint, logger)
 			return ui.RunLogin(cmd.Context(), version.Version(), platformClient, cfg.AuthToken, cfg.ForceFileKeyring, cfg.WebAppURL, logger)
-		},
+		}),
```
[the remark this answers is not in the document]
anisaoshafi: recheck

cmd/stop.go
```diff
@@ -29,11 +33,45 @@ func newStopCmd(cfg *env.Env) *cobra.Command {
 				return fmt.Errorf("failed to get config: %w", err)
 			}
 
+			commandEventID := telemetry.NewEventID()
+			stopOpts := container.StopOptions{
+				Telemetry:      tel,
+				TriggerEventID: commandEventID,
+				AuthToken:      cfg.AuthToken,
```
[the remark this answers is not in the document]
anisaoshafi: recheck

cmd/root.go
```diff
@@ -113,6 +124,69 @@ func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *teleme
...
+	}))
+
+	return runErr
+}
+
+// withCommandTelemetry wraps a RunE function so that an lstk_command event is
+// emitted after every invocation. Use this for commands that do not emit
+// lstk_lifecycle events (i.e. everything except start/stop, which manage their
+// own telemetry).
+func withCommandTelemetry(name string, tel *telemetry.Client, resolveAuthToken func() string, fn func(*cobra.Command, []string) error) func(*cobra.Command, []string) error {
```
carole-lavillonniere: nit: the name is not very idiomatic-go. `commandWithTelemetry` would be more appropriate (name of what it produces)

cmd/root.go
```diff
@@ -30,7 +33,7 @@ func NewRootCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.C
 			if err != nil {
 				return err
 			}
-			return runStart(cmd.Context(), rt, cfg, tel, logger)
+			return runStart(cmd, rt, cfg, tel, logger)
```
carole-lavillonniere: suggestion: it would be more conventional to keep the context as first argument and add the  `cmd.Flags()` as arg (as far as I can tell, it's the reason why we pass `cmd` here)

internal/telemetry/client.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func userAgent() string {
-	return fmt.Sprintf("localstack lstk/%s (%s; %s)", version.Version(), runtime.GOOS, runtime.GOARCH)
+	return fmt.Sprintf("localstack lstk/%s", version.Version())
```
carole-lavillonniere: question: why did we remove the last 2 values?

internal/container/start.go
```diff
@@ -37,6 +38,43 @@ type StartOptions struct {
...
+	authToken string
+}
+
+func (t telCtx) emitStartError(ctx context.Context, c runtime.ContainerConfig, errorCode, errorMsg string) {
+	if t.tel == nil {
+		return
+	}
+	t.tel.Emit(ctx, "lstk_lifecycle", telemetry.ToMap(telemetry.LifecycleEvent{
+		EventType:   telemetry.LifecycleStartError,
+		Environment: t.tel.GetEnvironment(t.authToken),
```
carole-lavillonniere: suggestion: can we set the `authToken` once in the `telemetry.Client` instance after it's resolved to avoid having to pass it everywhere? That would also remove the need for `telCtx`?

internal/container/start.go
```diff
@@ -209,6 +253,7 @@ func pullImages(ctx context.Context, rt runtime.Runtime, sink output.Sink, conta
 				Title:   fmt.Sprintf("Failed to pull %s", c.Image),
 				Summary: err.Error(),
 			})
+			tel.emitStartError(ctx, c, telemetry.ErrCodeImagePullFailed, err.Error())
```
carole-lavillonniere:
```suggestion
			tel.emitEmulatorStartError(ctx, c, telemetry.ErrCodeImagePullFailed, err.Error())
```
suggestion: Maybe a name that indicates it's an event about the emulator start (and avoid confusion with the start command)?

cmd/root.go
```diff
@@ -113,6 +124,69 @@ func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *teleme
...
+	}
+	tel.Emit(cmd.Context(), "lstk_command", telemetry.ToMap(telemetry.CommandEvent{
+		Environment: tel.GetEnvironment(cfg.AuthToken),
+		Parameters:  telemetry.CommandParameters{Command: "start", Flags: flags},
+		Result: telemetry.CommandResult{
+			DurationMS: time.Since(startTime).Milliseconds(),
+			ExitCode:   exitCode,
+			ErrorMsg:   errorMsg,
+		},
+	}))
```
carole-lavillonniere: suggestion: what about having helper methods such as `EmitCommand` and `EmitEmulatorLifecyleEvent` on `*telemetry.Client`? We would not need to know about `telemetry.ToMap`, `tel.GetEnvironment` etc here.

cmd/root.go
```diff
@@ -50,13 +53,13 @@ func NewRootCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.C
...
-		newLoginCmd(cfg, logger),
-		newLogoutCmd(cfg, logger),
-		newStatusCmd(cfg),
-		newLogsCmd(cfg),
+		newStopCmd(cfg, tel),
+		newLoginCmd(cfg, tel, logger),
+		newLogoutCmd(cfg, tel, logger),
+		newStatusCmd(cfg, tel),
+		newLogsCmd(cfg, tel),
 		newConfigCmd(),
```
carole-lavillonniere: question: why no telemetry for the config command?

internal/container/start.go
```diff
@@ -258,6 +309,7 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 		}
 		if err := ports.CheckAvailable(c.Port); err != nil {
 			emitPortInUseError(sink, c.Port)
+			tel.emitStartError(ctx, c, telemetry.ErrCodePortConflict, fmt.Sprintf("port %s already in use", c.Port))
```
carole-lavillonniere: question: why not?
```suggestion
			tel.emitStartError(ctx, c, telemetry.ErrCodePortConflict, err.Error())
```

test/integration/telemetry_test.go
```diff
@@ -80,17 +80,72 @@ func TestStartCommandSendsTelemetryEvent(t *testing.T) {
...
+	require.True(t, ok, "expected lstk_command event")
+	cp := command["payload"].(map[string]any)
+	params, ok := cp["parameters"].(map[string]any)
+	require.True(t, ok)
+	assert.Equal(t, "stop", params["command"])
+	result, ok := cp["result"].(map[string]any)
+	require.True(t, ok)
+	assert.InDelta(t, 0, result["exit_code"], 0)
+}
+
```
carole-lavillonniere, line 148: suggestion: can we add more integration test cases, also for failure cases?
My worry is that it is easy to accidentally drop an `Emit`.
I'd say we don't have to test the payload but at least that an event was emitted.
And maybe we can just add an extra assertion for the Emit in the existing test cases, instead of centralizing all tests for all commands in this file `test/integration/telemetry_test.go`

cmd/login.go
```diff
@@ -4,25 +4,37 @@ import (
...
+			tokenStorage, err := auth.NewTokenStorage(cfg.ForceFileKeyring, logger)
+			if err != nil {
+				return cfg.AuthToken
+			}
+			token, err := tokenStorage.GetAuthToken()
+			if err != nil {
+				return cfg.AuthToken
+			}
+			return token
+		}, func(cmd *cobra.Command, args []string) error {
```
carole-lavillonniere: question: why do we have this special logic to get the token for the `login` command only?

internal/telemetry/events.go
```diff
@@ -0,0 +1,101 @@
...
+	LifecycleStop         = "stop"
+	LifecycleStartError   = "start_error"
+)
+
+// Error codes for start_error lifecycle events.
+const (
+	ErrCodePortConflict    = "port_conflict"
+	ErrCodeImagePullFailed = "image_pull_failed"
+	ErrCodeLicenseInvalid  = "license_invalid"
+	ErrCodeStartFailed     = "start_failed"
```
carole-lavillonniere, line 79: question: do we only have error codes for start command?

carole-lavillonniere (commented):
Great job Anisa. I added a bunch of questions and suggestions but overall it looks good!

cmd/root.go
anisaoshafi: Thanks, done in https://github.com/localstack/lstk/pull/127/changes/37629c1d221938d393d22d697dd6f58bb9473e21

internal/container/start.go
```diff
@@ -217,35 +262,41 @@ func pullImages(ctx context.Context, rt runtime.Runtime, sink output.Sink, conta
...
 		output.EmitStatus(sink, "waiting", c.Name, "")
 		healthURL := fmt.Sprintf("http://localhost:%s%s", c.Port, c.HealthPath)
 		if err := awaitStartup(ctx, rt, sink, containerID, "LocalStack", healthURL); err != nil {
+			tel.emitStartError(ctx, c, telemetry.ErrCodeStartFailed, err.Error())
 			return err
 		}
 
 		output.EmitStatus(sink, "ready", c.Name, fmt.Sprintf("containerId: %s", containerID[:12]))
+
+		lsInfo, _ := fetchLocalStackInfo(ctx, c.Port)
```
carole-lavillonniere, line 289: potential issue: `fetchLocalStackInfo` is called synchronously adding up to 2 seconds of latency if the container is slow/unreachable. Maybe this call should happen asynchronously in the telemetry client's worker. At the same time it's a http request to localhost and it should return pretty much instantly 🤔 So the concern is more theoretical than practical. WDYT @thrau?

internal/telemetry/client.go
anisaoshafi: According to the proposal `os` and `arch` are already logged as part of the payload.environment section which is redundant (as done in [here](https://github.com/localstack/lstk/blob/main/internal/telemetry/client.go#L81-L82)). But brought it back just in case: https://github.com/localstack/lstk/pull/127/changes/ca5628d596191b901b088d136e6ac1796e08cca9.

cmd/root.go
anisaoshafi: My bad, thanks for the heads up. Added here: https://github.com/localstack/lstk/pull/127/changes/463ce792ac0a312000709c4a8b0ea14dc2a72b89.

internal/container/start.go
anisaoshafi: Done in https://github.com/localstack/lstk/pull/127/changes/02c82a5777c74913f9565637242c1c776ddf1643

internal/container/start.go
anisaoshafi: My bad 😅 , done in https://github.com/localstack/lstk/pull/127/changes/02c82a5777c74913f9565637242c1c776ddf1643. Thanks!

cmd/root.go
anisaoshafi: Done [here](https://github.com/localstack/lstk/pull/127/commits/ef07e5c34e4939f52c6459da53451a10f9e12ae6), thanks!

internal/container/start.go
anisaoshafi: Good point, that was too much. Will do 🏃🏼

cmd/login.go
anisaoshafi: In the user has run `lstk logout` for example, they can run`login` independently.

This logic is to catch the token that was just created for the event, but maybe not needed to over-engineer just for tracking purpose, I can remove it as well 🤷🏼

internal/container/start.go
anisaoshafi: Gave it a try here: https://github.com/localstack/lstk/pull/127/changes/2330ac1c8681d0ce747dfcdd62c769790193cd4b

internal/telemetry/events.go
anisaoshafi: Yes. 
This `error_code` is captured for start errors and only for `lstk_lifecycle` events  from what I assumed. 
While `error_message` is logged for `lstk_cmd` as well. 
FYI @thrau

internal/telemetry/events.go
```diff
@@ -0,0 +1,101 @@
...
+	"runtime"
+
+	"github.com/localstack/lstk/internal/version"
+)
+
+// Environment is the common environment block included in all telemetry events.
+type Environment struct {
+	LstkVersion string `json:"lstk_version"`
+	AuthTokenID string `json:"auth_token_id,omitempty"`
+	MachineID   string `json:"machine_id,omitempty"`
```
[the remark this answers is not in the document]
anisaoshafi: expected

cmd/root.go
anisaoshafi: Addressed here: https://github.com/localstack/lstk/pull/127/changes/04d5449a158a6da5524a48968b8faddcee2cb75f

test/integration/telemetry_test.go
anisaoshafi: Good point, attempted here: https://github.com/localstack/lstk/pull/127/changes/5da780fc96861e8a549298a1b0579c170a085b4d

cmd/root.go
```diff
@@ -113,6 +124,69 @@ func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *teleme
...
+		errorMsg = runErr.Error()
+	}
+	tel.Emit(cmd.Context(), "lstk_command", telemetry.ToMap(telemetry.CommandEvent{
+		Environment: tel.GetEnvironment(cfg.AuthToken),
+		Parameters:  telemetry.CommandParameters{Command: "start", Flags: flags},
+		Result: telemetry.CommandResult{
+			DurationMS: time.Since(startTime).Milliseconds(),
+			ExitCode:   exitCode,
+			ErrorMsg:   errorMsg,
+		},
```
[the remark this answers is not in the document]
anisaoshafi: Valid points. Would like to get reviewers' POV on this topic

internal/container/start.go
thrau: i think this is fine. i don't see a common case where it would  actually block for the full 2 seconds. for most errors that are plausible here, for example connection refused, it will return immediately. fine for now 👍

thrau (approved):
Thanks for putting this together @anisaoshafi (and @carole-lavillonniere for the initial client!). This is a overall great first iteration of a concrete telemetry layer for CLIv2 💯 

I have no blocking comments and am happy to approve the PR as is. Here's what stands out positively:

* Telemetry client and events are neatly encapsulated
* Telemetry concerns throughout the codebase are mostly encapsulated and isolated
* Implementation maps to the code style of the codebase (which is heavily functional)
* Payloads modeled very closely to the design doc with only a few changes that I think are good
* It's clearly well tested. It worked with this PR I created on the analytics backend side https://github.com/localstack/analytics-backend/pull/213
* I tested this end-to-end with our local analytics-backend setup, with common commands (start/stop/config path/config path -f), and in error scenarios. It all worked as expected :ok_hand:

Thanks @carole-lavillonniere for the detailed review! I have some additional high-level non-blocking comments and improvement suggestions around the implementation, but they feel like good backlog items rather than changes for this PR. :-)

## Improvement suggestions

Again, given that telemetry is not in the critical path, I consider all these things optional for now.

### Functional vs object-oriented style

This is more of a general statement about the style of the codebase, but noticed it by reviewing the telemetry code.

I think these emit helpers are a sign that we're forcing a stateless functional style onto behavior that actually has stable dependencies. In particular, having to thread `*telemetry.Client` through multiple layers makes the signatures noisy and the calling code more difficult to read. The result is that `emitEmulatorStartSuccess(...)` is less readable than the underlying `tel.EmitEmulatorLifecycleEvent(...)` call it wraps.

For example, I find this:

```go
      tel.EmitEmulatorLifecycleEvent(ctx, telemetry.LifecycleEvent{
            EventType:      telemetry.LifecycleStartSuccess,
            Emulator:       c.EmulatorType,
            Image:          c.Image,
            ContainerID:    containerID,
            DurationMS:     durationMS,
            Pulled:         pulled,
            LocalStackInfo: info,
      })
```

Much easier to read than
```go
emitEmulatorStartSuccess(ctx, tel, c, containerID[:12], time.Since(startTime).Milliseconds(), pulled[c.Name], lsInfo)
```

I find the telemetry Client itself to be a good example of a well encapsulated system that reads well.
I would love to see the codebase evolve more into that direction, and the telemetry system with it!

Nothing actionable for this PR at the moment.

### Targeted instrumentation of cli commands using the `commandWithTelemetry` decorator

I think this pattern is not unreasonable but has a few drawbacks that could be addressed with a more idiomatic approach using cobra hooks.

Here are some of the weakness I see:
* Telemetry is a concern that we want to isolate as much as possible
* The pattern relies on passing dependencies like the `telemetry.Client` down through every command constructor. This mixes concerns creates noise and makes command wiring less readable.
* Command names are manual strings which can drift from Cobra command paths

Instead, we could centralize this into `root.go` and remove the `tel` dependency out of the constructors completely.

Here's a concrete implementation suggestion on how to give the framework full control over instrumentation, instead of leaving it to downstream consumers.

* You can use Cobra's `PersistentPreRun` func to store data globally that you need (for example the `startTime`)
* In `NewRootCmd`, you call `instrumentCommands` on the root commands
* in `root.go`, we create a new function `instrumentCommands` that iterates over all children and wraps `cmd.RunE` with the telemetry emitting functionality
* You can construct the command path name recursively via cmd.Parent() and cmd.Name()
* To have certain CLI commands opt out of telemetry, you could use `Annotations` of a command:
  ```go
    &cobra.Command{
        Use:   "status",
        Annotations: map[string]string{
              "no-telemetry": "true",
        },
        ...
    }
  ```

It would add a bit more complexity, but would free all command constructors of telemetry dependencies.

Again, I don't see this as a blocker, this is more about where want to evolve the design.

cmd/root.go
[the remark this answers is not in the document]
carole-lavillonniere: @thrau what's your opinion on this one?

carole-lavillonniere (approved):
Thanks Anisa, this looks good! 
One last small thing could be to update the command skill to never forget to add telemetry to a new command.

anisaoshafi:
> Thanks Anisa, this looks good! One last small thing could be to update the command skill to never forget to add telemetry to a new command.

Thanks @carole-lavillonniere 🙌🏼 
I have some changes in `.claude/skills/add-command/SKILL.md`, can you check if that is enough?

carole-lavillonniere:
> > Thanks Anisa, this looks good! One last small thing could be to update the command skill to never forget to add telemetry to a new command.
> 
> Thanks @carole-lavillonniere 🙌🏼 I have some changes in `.claude/skills/add-command/SKILL.md`, can you check if that is enough?

Sorry for missing it! Looks perfect

anisaoshafi:
Thanks @thrau for the review, testing this and for the suggestions 🙌🏼 
We will discuss your suggestions with the team and decide if we want to move in that direction.

anisaoshafi:
Will merge for now, let's not forget to follow up on the security concern from coderabbit [here](https://github.com/localstack/lstk/pull/127#discussion_r2954581442).

carole-lavillonniere:
> Thanks for putting this together @anisaoshafi (and @carole-lavillonniere for the initial client!). This is a overall great first iteration of a concrete telemetry layer for CLIv2 💯
> 
> I have no blocking comments and am happy to approve the PR as is. Here's what stands out positively:
> 
>     * Telemetry client and events are neatly encapsulated
> 
>     * Telemetry concerns throughout the codebase are mostly encapsulated and isolated
> 
>     * Implementation maps to the code style of the codebase (which is heavily functional)
> 
>     * Payloads modeled very closely to the design doc with only a few changes that I think are good
> 
>     * It's clearly well tested. It worked with this PR I created on the analytics backend side [localstack/analytics-backend#213](https://github.com/localstack/analytics-backend/pull/213)
> 
>     * I tested this end-to-end with our local analytics-backend setup, with common commands (start/stop/config path/config path -f), and in error scenarios. It all worked as expected 👌
> 
> 
> Thanks @carole-lavillonniere for the detailed review! I have some additional high-level non-blocking comments and improvement suggestions around the implementation, but they feel like good backlog items rather than changes for this PR. :-)
> ## Improvement suggestions
> 
> Again, given that telemetry is not in the critical path, I consider all these things optional for now.
> ### Functional vs object-oriented style
> 
> This is more of a general statement about the style of the codebase, but noticed it by reviewing the telemetry code.
> 
> I think these emit helpers are a sign that we're forcing a stateless functional style onto behavior that actually has stable dependencies. In particular, having to thread `*telemetry.Client` through multiple layers makes the signatures noisy and the calling code more difficult to read. The result is that `emitEmulatorStartSuccess(...)` is less readable than the underlying `tel.EmitEmulatorLifecycleEvent(...)` call it wraps.
> 
> For example, I find this:
> 
> ```go
>       tel.EmitEmulatorLifecycleEvent(ctx, telemetry.LifecycleEvent{
>             EventType:      telemetry.LifecycleStartSuccess,
>             Emulator:       c.EmulatorType,
>             Image:          c.Image,
>             ContainerID:    containerID,
>             DurationMS:     durationMS,
>             Pulled:         pulled,
>             LocalStackInfo: info,
>       })
> ```
> 
> Much easier to read than
> 
> ```go
> emitEmulatorStartSuccess(ctx, tel, c, containerID[:12], time.Since(startTime).Milliseconds(), pulled[c.Name], lsInfo)
> ```
> 
> I find the telemetry Client itself to be a good example of a well encapsulated system that reads well. I would love to see the codebase evolve more into that direction, and the telemetry system with it!
> 
> Nothing actionable for this PR at the moment.
> ### Targeted instrumentation of cli commands using the `commandWithTelemetry` decorator
> 
> I think this pattern is not unreasonable but has a few drawbacks that could be addressed with a more idiomatic approach using cobra hooks.
> 
> Here are some of the weakness I see:
> 
>     * Telemetry is a concern that we want to isolate as much as possible
> 
>     * The pattern relies on passing dependencies like the `telemetry.Client` down through every command constructor. This mixes concerns creates noise and makes command wiring less readable.
> 
>     * Command names are manual strings which can drift from Cobra command paths
> 
> 
> Instead, we could centralize this into `root.go` and remove the `tel` dependency out of the constructors completely.
> 
> Here's a concrete implementation suggestion on how to give the framework full control over instrumentation, instead of leaving it to downstream consumers.
> 
>     * You can use Cobra's `PersistentPreRun` func to store data globally that you need (for example the `startTime`)
> 
>     * In `NewRootCmd`, you call `instrumentCommands` on the root commands
> 
>     * in `root.go`, we create a new function `instrumentCommands` that iterates over all children and wraps `cmd.RunE` with the telemetry emitting functionality
> 
>     * You can construct the command path name recursively via cmd.Parent() and cmd.Name()
> 
>     * To have certain CLI commands opt out of telemetry, you could use `Annotations` of a command:
>       ```go
>         &cobra.Command{
>             Use:   "status",
>             Annotations: map[string]string{
>                   "no-telemetry": "true",
>             },
>             ...
>         }
>       ```
> 
> 
> It would add a bit more complexity, but would free all command constructors of telemetry dependencies.
> 
> Again, I don't see this as a blocker, this is more about where want to evolve the design.

Thanks @thrau. Your comment is seen, appreciated, and will be followed up on. Only not now, since there are more pressing matters.

PR #128, opened 2026-03-18
Wrap error title to terminal width

gtsiolis:
This will:
1. Soft-wrap the error display title so long messages (e.g. license validation failures) wrap at word boundaries instead of being truncated at the terminal edge.
1. Use rune count instead of byte length for the `✗ ` prefix to get correct indentation on continuation lines
1. Add test for title wrapping behavior

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-18 at 10 41 47 | Screenshot 2026-03-18 at 10 41 30 |

### DIFF
--- a/internal/ui/components/error_display.go
+++ b/internal/ui/components/error_display.go
@@ -2,6 +2,7 @@ package components
 
 import (
 	"strings"
+	"unicode/utf8"
 
 	"github.com/localstack/lstk/internal/output"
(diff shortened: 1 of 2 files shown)


carole-lavillonniere (approved):

gtsiolis:
Thanks for taking a look, @carole-lavillonniere! 🍿

PR #132, opened 2026-03-18
Fix status command failing with non-default port

carole-lavillonniere:
## Motivation

`lstk status` was always using the port from config to reach LocalStack's HTTP API. If the user changed the config port while a container was still running on the old port, status would try the wrong port and return a connection refused error.

## Changes

- `status` now reads the actual host port from Docker's port bindings via `GetBoundPort`, falling back to the config port if unavailable
- Added `GetBoundPort` to the `Runtime` interface, implemented in `DockerRuntime` and the mock
- Merged `startTestContainer` and `startTestContainerWithPortBinding` into a single variadic function

## Tests

Added `TestStatusCommandWorksWithNonDefaultPort`: starts a container with a non-default port binding, writes a config with the wrong port (4566), and verifies that `lstk status` still connects successfully using Docker's actual bound port.

Closes DRG-642

### DIFF
--- a/internal/container/status.go
+++ b/internal/container/status.go
@@ -39,7 +39,16 @@ func Status(ctx context.Context, rt runtime.Runtime, containers []config.Contain
 			return output.NewSilentError(fmt.Errorf("%s is not running", name))
 		}
 
-		host, _ := endpoint.ResolveHost(c.Port, localStackHost)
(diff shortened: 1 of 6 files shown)


silv-io (approved):
LGTM!

PR #136, opened 2026-03-18
Add update notification prompt on command start

silv-io:
## Motivation

Users don't know when a new CLI version is available unless they manually run `lstk update --check`. This adds a check on every start so users are prompted to update.

## Changes

- Add `CheckQuietly`, `NotifyUpdate`, and `UpdateCommandHint` in `internal/update/notify.go` to silently check for updates and prompt interactive users
- Extract `applyUpdate` from `Update` so the notify flow can run the update without a redundant version check
- Add `update_prompt` config default (`true`) and `config.Set` for persisting config changes
- Wire update notification into `ui.Run` (TUI) and the non-interactive path in `cmd/root.go`
- Interactive mode shows a prompt: Update / SKIP / Never ask again
- Non-interactive mode emits a note with the update command hint
- Change `latestReleaseURL` from const to var to allow test overrides

## Tests

- Unit tests for `CheckQuietly` (dev build, network error, update available, up-to-date)
- Unit tests for `UpdateCommandHint` (homebrew, npm, binary)
- Unit tests for `NotifyUpdate` (no update, prompt disabled, skip, never, cancelled)
- Manual TUI verification via tmux for all three prompt options (u/s/n) and non-interactive mode

## Related

- [FLC-472](https://linear.app/localstack/issue/FLC-472/prompt-notifications-to-update-cli)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -122,7 +122,7 @@ func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *t
 
 	notifyOpts := update.NotifyOptions{
 		GitHubToken:    cfg.GitHubToken,
-		UpdatePrompt:   config.IsUpdatePromptEnabled(),
+		UpdatePrompt:   appConfig.UpdatePrompt,
 		PersistDisable: config.DisableUpdatePrompt,
 	}
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -21,7 +21,6 @@ import (
 	"github.com/localstack/lstk/internal/version"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"github.com/spf13/viper"
 )
 
 func NewRootCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.Command {
@@ -121,12 +120,10 @@ func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *t
 		Telemetry:        tel,
 	}
 
-	notifyOpts := ui.UpdateNotifyOptions{
-		GitHubToken:  cfg.GitHubToken,
-		UpdatePrompt: viper.GetBool("update_prompt"),
-		PersistDisable: func() error {
-			return config.Set("update_prompt", false)
-		},
+	notifyOpts := update.NotifyOptions{
+		GitHubToken:    cfg.GitHubToken,
+		UpdatePrompt:   config.IsUpdatePromptEnabled(),
+		PersistDisable: config.DisableUpdatePrompt,
 	}
 
 	if isInteractiveMode(cfg) {
@@ -134,7 +131,7 @@ func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *t
 	}
 
 	sink := output.NewPlainSink(os.Stdout)
-	update.NotifyUpdate(ctx, sink, notifyOpts.GitHubToken, false, nil)
+	update.NotifyUpdate(ctx, sink, update.NotifyOptions{GitHubToken: cfg.GitHubToken})
 	return container.Start(ctx, rt, sink, opts, false)
 }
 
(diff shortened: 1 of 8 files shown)


cmd/root.go
```diff
@@ -107,10 +109,21 @@ func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *teleme
 		Logger:           logger,
 	}
 
+	notifyOpts := ui.UpdateNotifyOptions{
+		GitHubToken:  cfg.GitHubToken,
+		UpdatePrompt: viper.GetBool("update_prompt"),
```
carole-lavillonniere: issue: why adding a viper dependency here instead of using the config abstraction layer? (maybe worth adding to CLAUDE.md/skill)

cmd/root.go
```diff
@@ -107,10 +109,21 @@ func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *teleme
...
+		PersistDisable: func() error {
+			return config.Set("update_prompt", false)
```
carole-lavillonniere: issue: what about a `config.DisableUpdatePrompt()` to not rely on the string value here? Or at least introduce a type for the string.

cmd/root.go
```diff
@@ -107,10 +109,21 @@ func runStart(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *teleme
...
+
 	if isInteractiveMode(cfg) {
-		return ui.Run(ctx, rt, version.Version(), opts)
+		return ui.Run(ctx, rt, version.Version(), opts, notifyOpts)
 	}
-	return container.Start(ctx, rt, output.NewPlainSink(os.Stdout), opts, false)
+
+	sink := output.NewPlainSink(os.Stdout)
+	update.NotifyUpdate(ctx, sink, notifyOpts.GitHubToken, false, nil)
+	return container.Start(ctx, rt, sink, opts, false)
```
carole-lavillonniere, line 135: thought: we're adding latency to all commands with that check, so I am wondering if we should do the check asynchronously or just less often in the future (say once per 24h). Just a note for the future, no action required.

internal/ui/run.go
```diff
@@ -24,7 +25,13 @@ func (s programSender) Send(msg any) {
 	s.p.Send(msg)
 }
 
-func Run(parentCtx context.Context, rt runtime.Runtime, version string, opts container.StartOptions) error {
+type UpdateNotifyOptions struct {
+	GitHubToken    string
+	UpdatePrompt   bool
+	PersistDisable func() error
+}
```
carole-lavillonniere: nit: what about moving the struct to the internal/update package, feels more like the right domain?

internal/update/notify_test.go
```diff
@@ -0,0 +1,177 @@
...
+	sink := output.SinkFunc(func(event any) {
+		events = append(events, event)
+		if req, ok := event.(output.UserInputRequestEvent); ok {
+			req.ResponseCh <- output.InputResponse{Cancelled: true}
+		}
+	})
+
+	exit := notifyUpdateWithVersion(context.Background(), sink, "", true, nil, "1.0.0", testFetcher(server.URL))
+	assert.False(t, exit)
+}
```
carole-lavillonniere, line 166: question: what about adding integration tests? It should be possible if we make the github URL overwritable via env var, right?

carole-lavillonniere (commented):
A couple of issues and nits, but overall looks good!

cmd/root.go
silv-io: good catch, you're right :)

cmd/root.go
silv-io: Created an issue already for this after noticing. What other tools do is do the check in the background, write it down and on next start show it.

For now I've set the timeout to 500ms. If it fails because of the timeout we won't show it.

internal/config/config.go
```diff
@@ -96,6 +96,14 @@ func Set(key string, value any) error {
 	return viper.WriteConfig()
 }
 
+func IsUpdatePromptEnabled() bool {
+	return viper.GetBool("update_prompt")
```
carole-lavillonniere: question: shouldn't this go as a new field in the Config struct instead?
```
UpdatePrompt bool `mapstructure:"update_prompt"`
```

silv-io: yes 😬

carole-lavillonniere (approved):

PR #139, opened 2026-03-18
Avoid duplicate keychain prompts during auth token lookup

gtsiolis:
Remove the eager `keyring.Get(...)` probe from `NewTokenStorage()` to avoid duplicate macOS Keychain prompts on startup.

**Before:** `NewTokenStorage()` probed the keyring once to check availability, then the normal auth path read it again — triggering up to 4 Keychain dialogs for a single token lookup.

**After:** `systemTokenStorage` holds a `keyringer` interface (injectable, testable) and falls back to file-based storage lazily, per operation. No more probe.

Cc @carole-lavillonniere because https://github.com/localstack/lstk/pull/119.

| Dialog 1/4 | Dialog 2/4 | Dialog 3/4 | Dialog 4/4 |
|--------|--------|--------|--------|
| Screenshot 2026-03-19 at 00 09 37 | Screenshot 2026-03-19 at 00 09 40 | Screenshot 2026-03-19 at 00 09 42 | Screenshot 2026-03-19 at 00 09 45 |

### DIFF
--- a/internal/auth/mock_token_storage.go
+++ b/internal/auth/mock_token_storage.go
@@ -81,3 +81,70 @@ func (mr *MockAuthTokenStorageMockRecorder) SetAuthToken(token any) *gomock.Call
 	mr.mock.ctrl.T.Helper()
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "SetAuthToken", reflect.TypeOf((*MockAuthTokenStorage)(nil).SetAuthToken), token)
 }
+
+// Mockkeyringer is a mock of keyringer interface.
+type Mockkeyringer struct {
+	ctrl     *gomock.Controller
+	recorder *MockkeyringerMockRecorder
+	isgomock struct{}
+}
+
+// MockkeyringerMockRecorder is the mock recorder for Mockkeyringer.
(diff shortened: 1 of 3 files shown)


gtsiolis:
This comes with a lower priority and possibly something to revist after the release since it's internal improvement. 🙏

carole-lavillonniere:
Great catch @gtsiolis. It makes a lot of sense to remove the probe on start. I pushed a commit to refactor the code a bit and I also updated the PR description. It would be great if you could test again on macOS and I would like to get a review from @silv-io or @anisaoshafi on this.

gtsiolis:
Weirdly, I see no keyring dialogs anymore even on `main`. 🙌

anisaoshafi:
> Weirdly, I see no keyring dialogs anymore even on `main`. 🙌

Same here, I tested on both main and this branch but couldn't reproduce the duplicate prompts ➕

anisaoshafi (approved):
Didn't reproduce the issue in description, but the refactor makes sense and great to have more tests. LGTM 👏🏼

silv-io (approved):
Great catch! looks good to me

PR #143, opened 2026-03-19
Use real LocalStack resources in status command integration test

carole-lavillonniere:
## Motivation

`TestStatusCommandShowsResourcesWhenRunning` was testing resource display against a mock HTTP server, which meant it couldn't catch issues with how `lstk status` interprets actual LocalStack responses.

## Changes

- Start a real LocalStack instance instead of an alpine stub + mock HTTP server
- Create an S3 bucket and SQS queue via AWS SDK v2, then verify both appear in `lstk status` output
- Add AWS SDK v2 packages to the integration test module


anisaoshafi (approved):
Nice 👏🏼

PR #146, opened 2026-03-19
Refine AWS profile setup flow

gtsiolis:
This PR improves the AWS profile setup flow to reduce repeated prompts during `lstk start` without adding extra persisted config state.

Users running lstk start repeatedly were prompted every time even when a partial profile existed. This change makes the prompt conditional while still guiding users to set up a complete profile via lstk setup aws.

1. `lstk start` now prompts only when the localstack profile is missing from both `~/.aws/config` and `~/.aws/credentials`
1. If the profile exists in only one file, or exists in both but is incomplete, lstk start shows a non-blocking note instead of prompting
1. Added `lstk setup aws` command for interactive AWS profile setup
1. Updated prompt text to "Set up a LocalStack profile for AWS CLI and SDKs in ~/.aws?"
1. Shows success message on confirm and skip note on decline

| No config | Config created | Config skipped | Partial config | Config command |
|--------|--------|--------|--------|--------|
| Screenshot 2026-03-23 at 18 10 41 | Screenshot 2026-03-23 at 18 10 50 | Screenshot 2026-03-23 at 18 11 15 | Screenshot 2026-03-23 at 18 23 09 | Screenshot 2026-03-23 at 18 25 43 |

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -6,6 +6,7 @@ import (
 	"github.com/localstack/lstk/internal/config"
 	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/telemetry"
+	"github.com/localstack/lstk/internal/ui"
 	"github.com/spf13/cobra"
 )
 
@@ -14,8 +15,30 @@ func newConfigCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
 		Use:   "config",
 		Short: "Manage configuration",
 	}
+	cmd.AddCommand(newConfigProfileCmd(cfg, tel))
 	cmd.AddCommand(newConfigPathCmd(cfg, tel))
 	return cmd
+}
+
+func newConfigProfileCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
+	return &cobra.Command{
+		Use:     "profile",
+		Short:   "Deprecated: use 'lstk setup aws' instead",
+		PreRunE: initConfig,
+		RunE: commandWithTelemetry("config profile", tel, func(cmd *cobra.Command, args []string) error {
+			appConfig, err := config.Get()
+			if err != nil {
+				return fmt.Errorf("failed to get config: %w", err)
+			}
+
+			if !isInteractiveMode(cfg) {
+				return fmt.Errorf("config profile requires an interactive terminal")
+			}
+
+			// Delegate to the same handler as "lstk setup aws"
+			return ui.RunConfigProfile(cmd.Context(), appConfig.Containers, cfg.LocalStackHost)
+		}),
+	}
 }
 
 func newConfigPathCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
--- a/internal/awsconfig/awsconfig.go
+++ b/internal/awsconfig/awsconfig.go
@@ -76,8 +76,12 @@ func (s profileStatus) anyNeeded() bool {
 	return s.configNeeded || s.credsNeeded
 }
 
-// checkProfileStatus determines which AWS profile files need to be written or updated.
-func checkProfileStatus(configPath, credsPath, resolvedHost string) (profileStatus, error) {
+// CheckProfileStatus determines which AWS profile files need to be written or updated.
+func CheckProfileStatus(resolvedHost string) (profileStatus, error) {
+	configPath, credsPath, err := awsPaths()
+	if err != nil {
+		return profileStatus{}, err
+	}
 	configNeeded, err := configNeedsWrite(configPath, resolvedHost)
 	if err != nil {
 		return profileStatus{}, err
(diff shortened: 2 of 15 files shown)


test/integration/awsconfig_test.go
```diff
@@ -65,35 +65,24 @@ func TestStartPromptsToCreateAWSProfileWhenMissing(t *testing.T) {
...
 	configContent, err := os.ReadFile(filepath.Join(tmpHome, ".aws", "config"))
 	require.NoError(t, err, "~/.aws/config should have been created")
 	assert.Contains(t, string(configContent), "[profile localstack]")
-	assert.Contains(t, string(configContent), "endpoint_url")
 
 	credsContent, err := os.ReadFile(filepath.Join(tmpHome, ".aws", "credentials"))
 	require.NoError(t, err, "~/.aws/credentials should have been created")
-	normalizedCreds := strings.Join(strings.Fields(string(credsContent)), " ")
-	assert.Contains(t, normalizedCreds, "[localstack]")
-	assert.Contains(t, normalizedCreds, "aws_access_key_id = test")
```
anisaoshafi, line 92: ❓ did you intend to remove these asserts?

test/integration/awsconfig_test.go
```diff
@@ -164,5 +153,141 @@ func TestStartNonInteractiveEmitsNoteWhenAWSProfileMissing(t *testing.T) {
...
+	out := &syncBuffer{}
+	outputCh := make(chan struct{})
+	go func() {
+		_, _ = io.Copy(out, ptmx)
+		close(outputCh)
+	}()
+
+	require.Eventually(t, func() bool {
+		return bytes.Contains(out.Bytes(), []byte("LocalStack AWS profile is incomplete. Run 'lstk config profile'."))
+	}, 2*time.Minute, 200*time.Millisecond, "AWS profile note should appear")
```
anisaoshafi, line 200: question: I see the 2 min timeout was introduced in other tests. Do you know if it actually needs to wait for that long?

anisaoshafi (approved):
LGTM, tested and worked well for me 🚀

cmd/config.go
```diff
@@ -14,10 +17,47 @@ func newConfigCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
...
+
+			resolvedHost, dnsOK := endpoint.ResolveHost(awsContainer.Port, cfg.LocalStackHost)
+			if !dnsOK {
+				output.EmitNote(output.NewPlainSink(cmd.OutOrStdout()), `Could not resolve "localhost.localstack.cloud" - your system may have DNS rebind protection enabled. Using 127.0.0.1 as the endpoint.`)
+			}
+
+			if !isInteractiveMode(cfg) {
+				return fmt.Errorf("config profile requires an interactive terminal")
+			}
+
```
silv-io: issue: this shouldn't happen in the cmd code but we can polish this later

test/integration/awsconfig_test.go
silv-io: `require.Eventually` is a retry loop. So it will first only wait for the 200ms. So if it's faster than that it will be fine I think.

If you mean that 2 minute total timeout seems a bit long, I tend to agree. that should be more in the seconds since we're just writing to some file.

internal/awsconfig/awsconfig.go
```diff
@@ -203,21 +232,52 @@ func Setup(ctx context.Context, sink output.Sink, interactive bool, resolvedHost
...
+	configPath, credsPath, err := awsPaths()
+	if err != nil {
+		output.EmitWarning(sink, fmt.Sprintf("could not determine AWS config paths: %v", err))
 		return nil
 	}
 
 	responseCh := make(chan output.InputResponse, 1)
 	output.EmitUserInputRequest(sink, output.UserInputRequestEvent{
-		Prompt:     "Configure AWS profile in ~/.aws/?",
+		Prompt:     "Set up a LocalStack profile for AWS CLI and SDKs?",
```
silv-io: I think we got a specific concern that our prompt doesn't say where it will be saved to. This is why it says `.aws`. I don't think we should remove that again.

silv-io (approved):
LGTM feature wise but it will require some cleanup after. 

If this doesn't need to go into the release I'd fix that issue before merging, though.

test/integration/awsconfig_test.go
gtsiolis: Nice catch! ⚾ Probably removed these as I was approaching the current UX flow. Restored in 95e8554.

test/integration/awsconfig_test.go
gtsiolis: Agree, changed in 7384cc9.

cmd/config.go
gtsiolis: Made an attempt to change this in d90de77.

internal/awsconfig/awsconfig.go
gtsiolis: Restored in da25541. The concern[[1](https://linear.app/localstack/issue/DES-178/lstk-recreates-stale-config-entries-after-removal#comment-10c4b4dd)] can still exist as it does not clarify config or credentials file, but I agree the explicit mention here is better.

carole-lavillonniere:
@gtsiolis How do we envision `lstk config profile` to work when we support multiple emulators?

I am concerned a "profile" is a very aws-specific thing and I feel that "lstk config profile" does not fit our existing naming scheme because:
- config currently refers to lstk's own configuration
- it sounds like it's a readonly operation, like `lstk config path` is

What about `lstk setup aws`? Since "setup" is a verb it's unambiguously an action.

internal/ui/run_awsconfig.go
```diff
@@ -0,0 +1,87 @@
...
+		return runErr
+	}
+
+	return nil
+}
+
+// RunAWSProfileSetup runs the AWS profile setup flow with TUI output.
+// resolvedHost must be a host:port string (e.g. "localhost.localstack.cloud:4566").
+// Deprecated: Use RunConfigProfile instead, which accepts container config and resolves the host.
+func RunAWSProfileSetup(parentCtx context.Context, resolvedHost string) error {
```
carole-lavillonniere: nit: it it's deprecated can we remove this code?

internal/config/containers.go
```diff
@@ -140,3 +140,14 @@ func (c *ContainerConfig) ProductName() (string, error) {
...
+// GetAWSContainer returns the AWS container config from the list of containers.
+// Returns nil if no AWS container is configured.
+func GetAWSContainer(containers []ContainerConfig) *ContainerConfig {
+	for i := range containers {
+		if containers[i].Type == EmulatorAWS {
+			return &containers[i]
+		}
+	}
+	return nil
+}
```
carole-lavillonniere: nit: this method does not really belong in this file in my opinion, it could be moved to the call place and it does not need to be exported.

internal/container/start.go
```diff
@@ -184,7 +184,7 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, opts Start
 	// Maps emulator types to their post-start setup functions.
 	// Add an entry here to run setup for a new emulator type (e.g. Azure, Snowflake).
 	setups := map[config.EmulatorType]postStartSetupFunc{
-		config.EmulatorAWS: awsconfig.Setup,
+		config.EmulatorAWS: awsconfig.NotifyIfMissing,
```
carole-lavillonniere: nit: this new name is misleading. It implies a read-only operation, but it can actually trigger the full setup flow. Suggestions:  `EnsureProfile`?

gtsiolis:
Great point, @carole-lavillonniere! 💯 

I was kinda hoping for some pushback on this one! 🙏 

Liked the `setup aws` command, since there's no similar profile for Azure, etc.

Rebased and added that, updated also PR description. 🐰

internal/config/containers.go
carole-lavillonniere: Looks like this is still not addressed?

cmd/config.go
```diff
@@ -14,10 +15,32 @@ func newConfigCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
...
+			// Delegate to the same handler as "lstk setup aws"
+			return ui.RunConfigProfile(cmd.Context(), appConfig.Containers, cfg.LocalStackHost)
+		}),
+	}
+}
+
```
carole-lavillonniere, line 43: suggestion: We can add the new command to readme + doc?

carole-lavillonniere (approved):
Just 2 minor comments left, great improvement! (Also one test is still failing)

gtsiolis:
Thanks for taking another look, @carole-lavillonniere! Tried to address review round comments in a5ebba4.

internal/config/containers.go
carole-lavillonniere: ping @gtsiolis

gtsiolis:
Rebased ot keep fresh! 🥗

gtsiolis:
Rebased to keep fresh! 🥗

gtsiolis:
Rebased to keep fresh! 🥗

Could you take another look, @carole-lavillonniere?

internal/config/containers.go
gtsiolis: Thanks for the nudge, @carole-lavillonniere! Made an attempt to resolve this in 45fbc85.

carole-lavillonniere (approved):

PR #147, opened 2026-03-20
Use catalog API to resolve latest version before pulling

anisaoshafi:
When the image tag is "latest", the CLI previously had to pull the image first to resolve the version for license validation.

This changes the order: the catalog API  (/v1/license/catalog/{emulator_type}/version) is called first to get the latest released version, so license validation happens before any image data is downloaded. 

If the catalog API is unavailable, the original behavior is preserved as a fallback (pull first, inspect image for version, then validate).

### DIFF
--- a/internal/api/catalog_version_test.go
+++ b/internal/api/catalog_version_test.go
@@ -1,0 +1,121 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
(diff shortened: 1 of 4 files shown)


carole-lavillonniere (approved):
Great work 👏

silv-io (approved):
LGTM!

PR #149, opened 2026-03-20
Use .lstk/config.toml for project-local config

silv-io:
## Motivation

Local config used `./lstk.toml` while all other config locations used `config.toml`, which was confusing — users naturally create `config.toml` in their project directory and it doesn't get picked up. This changes the local config to `.lstk/config.toml` inside a conventional dotdir, making the filename consistent across all config locations.

## Changes

- Move config path constants from `containers.go` to `paths.go` where they belong
- Change local config path from `./lstk.toml` to `./.lstk/config.toml`
- Unify on a single `configFileName` constant since all locations now use `config.toml`
- Replace manual path search in `Init()` with Viper's built-in `AddConfigPath`/`SetConfigName` search
- Replace `configSearchPaths()` (full file paths) with `configSearchDirs()` (directories only)
- Update integration test and documentation

## Tests

- All existing integration tests pass, including `TestLocalConfigTakesPrecedence` updated to use the new `.lstk/config.toml` path

## Related

- [FLC-536](https://linear.app/localstack/issue/FLC-536/lstk-doesnt-favour-local-config)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,8 +43,8 @@ lstk always writes diagnostic logs to `$CONFIG_DIR/lstk.log` (appends across run
 
 # Configuration
 
-Uses Viper with TOML format. lstk uses the first config file found in this order:
-1. `./lstk.toml` (project-local)
+Uses Viper with TOML format. lstk uses the first `config.toml` found in this order:
+1. `./.lstk/config.toml` (project-local)
 2. `$HOME/.config/lstk/config.toml`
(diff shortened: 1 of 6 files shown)


gtsiolis (commented):
Go boring solutions! It's an elegant solution, compared to other ideas discussed.

anisaoshafi (approved):
LGTM, local test worked fine 🎉  thanks Silvio 🙏🏼

PR #151, opened 2026-03-21
Bind container ports to 127.0.0.1

gtsiolis:
Change container port binding to 127.0.0.1.

PR #155, opened 2026-03-23
Remove 127.0.0.2 workaround from status integration test

carole-lavillonniere:
Now that `DockerRuntime` binds ports to `127.0.0.1` instead of `0.0.0.0` (see https://github.com/localstack/lstk/pull/151), the `127.0.0.2` trick used to avoid conflicts between the mock HTTP server and the test container is no longer necessary for the integration test `TestStatusCommandWorksWithNonDefaultPort`


carole-lavillonniere:
Closing because this workaround is actually still needed.

PR #156, opened 2026-03-23
Bind port 443 for HTTPS 

carole-lavillonniere:
## Motivation

Port 443 was exposed by the LocalStack container but never mapped to the host, making HTTPS endpoints unreachable (e.g. \`https://localhost.localstack.cloud/_localstack/health\` returned connection refused).

## Tests

New sub-tests in \`TestStartCommandSetsUpContainerCorrectly\` verify that both \`http://localhost.localstack.cloud:4566/_localstack/health\` and \`https://localhost.localstack.cloud/_localstack/health\` return 200 after \`lstk start\`. The HTTPS test was failing before this fix.

Closes DRG-661


anisaoshafi (approved):
👏🏼

PR #159, opened 2026-03-23
Simplify update notification flow

gtsiolis:
This PR simplifies the update notification flow and removes the "Never ask again" option.

Users now see a cleaner three-option prompt: Update now [U], Remind me next time [R], Skip this version [S].

The "Skip" option persists the skipped version to config, users won't be prompted again until a newer version is available.

#### Update Notification
- Removed the `"Never ask again [N]"` option from the prompt
- Increased version check timeout from 500ms to 2s
- Skip option persists the skipped version to config via `PersistSkipVersion` callback
- Removed legacy config migration code for old `update_prompt` and `update_skipped_version` keys at root level

#### Config Schema
- Removed `update_prompt = true` from default config template (true is already the default via viper)
- Config `Set()` now writes only the changed key to the TOML file instead of rewriting all keys (avoids leaking Viper defaults)

#### UI Improvements
- Added `Vertical` field to `UserInputRequestEvent` to explicitly indicate vertical layout
- Replaced string parsing (`strings.Contains(options[0].Label, "[")`) with explicit boolean field
- Extracted `handleVerticalPromptKey()` method from Update() to reduce complexity
- Removed update-specific text rendering from generic `formatResolvedInput()` function
- UI code no longer contains hard-coded update prompts like "Updating lstk..." or "Won't ask again"
- Fixed resolved vertical prompt showing all options instead of only the selected one

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-23 at 10 38 51 | Screenshot 2026-03-23 at 14 42 39 |

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -14,8 +14,7 @@ import (
 var defaultConfigTemplate string
 
 type CLIConfig struct {
-	UpdatePrompt         bool   `mapstructure:"update_prompt"`
-	UpdateSkippedVersion string `mapstructure:"update_skipped_version"`
+	UpdatePrompt bool `mapstructure:"update_prompt"`
 }
 
 type Config struct {
@@ -33,7 +32,6 @@ func setDefaults() {
 		},
 	})
 	viper.SetDefault("cli.update_prompt", true)
-	viper.SetDefault("cli.update_skipped_version", "")
 }
 
 func loadConfig(path string) error {
@@ -122,14 +120,6 @@ func DisableUpdatePrompt() error {
 	return Set("cli.update_prompt", false)
 }
 
-func SetUpdateSkippedVersion(version string) error {
-	return Set("cli.update_skipped_version", version)
-}
-
-func GetUpdateSkippedVersion() string {
-	return viper.GetString("cli.update_skipped_version")
-}
-
 func Get() (*Config, error) {
 	var cfg Config
 	if err := viper.Unmarshal(&cfg); err != nil {
@@ -137,9 +127,6 @@ func Get() (*Config, error) {
 	}
 	if !viper.InConfig("cli.update_prompt") && viper.InConfig("update_prompt") {
 		cfg.CLI.UpdatePrompt = viper.GetBool("update_prompt")
-	}
-	if !viper.InConfig("cli.update_skipped_version") && viper.InConfig("update_skipped_version") {
-		cfg.CLI.UpdateSkippedVersion = viper.GetString("update_skipped_version")
 	}
 	for i := range cfg.Containers {
 		if err := cfg.Containers[i].Validate(); err != nil {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,11 +1,13 @@
 package config
 
 import (
+	"bufio"
 	_ "embed"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/spf13/viper"
 )
@@ -14,7 +16,8 @@ import (
 var defaultConfigTemplate string
 
 type CLIConfig struct {
-	UpdatePrompt bool `mapstructure:"update_prompt"`
+	UpdatePrompt         bool   `mapstructure:"update_prompt"`
+	UpdateSkippedVersion string `mapstructure:"update_skipped_version"`
 }
 
 type Config struct {
@@ -113,11 +116,109 @@ func resolvedConfigPath() string {
 
 func Set(key string, value any) error {
 	viper.Set(key, value)
-	return viper.WriteConfig()
+	return setInFile(viper.ConfigFileUsed(), key, value)
+}
+
+// setInFile updates a single key in the TOML config file without
+// rewriting unrelated keys (avoids Viper dumping all defaults).
+func setInFile(path, key string, value any) error {
+	// Split "cli.update_skipped_version" into section "cli" and field "update_skipped_version".
+	parts := strings.SplitN(key, ".", 2)
+	if len(parts) != 2 {
+		// Top-level keys: fall back to full rewrite.
+		return viper.WriteConfig()
+	}
+	section, field := parts[0], parts[1]
+
+	formatted := formatTOMLValue(value)
+	targetLine := field + " = " + formatted
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to read config file: %w", err)
+	}
+
+	var result []string
+	scanner := bufio.NewScanner(strings.NewReader(string(data)))
+	inSection := false
+	replaced := false
+	sectionHeader := "[" + section + "]"
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		trimmed := strings.TrimSpace(line)
+
+		// Detect section headers.
+		if strings.HasPrefix(trimmed, "[") {
+			if trimmed == sectionHeader {
+				inSection = true
+			} else if inSection {
+				// Leaving our section without having replaced — insert before the new section.
+				if !replaced {
+					result = append(result, targetLine)
+					replaced = true
+				}
+				inSection = false
+			}
+		}
+
+		// Replace existing key in the target section.
+		if inSection && strings.HasPrefix(trimmed, field+" ") || inSection && strings.HasPrefix(trimmed, field+"=") {
+			result = append(result, targetLine)
+			replaced = true
+			continue
+		}
+
+		result = append(result, line)
+	}
+
+	// Section exists but key was not found — append to end of file (still in section).
+	if !replaced && inSection {
+		result = append(result, targetLine)
+		replaced = true
+	}
+
+	// Section doesn't exist at all — append section and key.
+	if !replaced {
+		if len(result) > 0 && strings.TrimSpace(result[len(result)-1]) != "" {
+			result = append(result, "")
+		}
+		result = append(result, sectionHeader)
+		result = append(result, targetLine)
+	}
+
+	output := strings.Join(result, "\n")
+	if !strings.HasSuffix(output, "\n") {
+		output += "\n"
+	}
+
+	return os.WriteFile(path, []byte(output), 0644)
+}
+
+func formatTOMLValue(v any) string {
+	switch val := v.(type) {
+	case string:
+		return fmt.Sprintf("%q", val)
+	case bool:
+		if val {
+			return "true"
+		}
+		return "false"
+	default:
+		return fmt.Sprintf("%v", val)
+	}
 }
 
 func DisableUpdatePrompt() error {
 	return Set("cli.update_prompt", false)
+}
+
+func SetUpdateSkippedVersion(version string) error {
+	return Set("cli.update_skipped_version", version)
+}
+
+func GetUpdateSkippedVersion() string {
+	return viper.GetString("cli.update_skipped_version")
 }
 
 func Get() (*Config, error) {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -131,7 +131,7 @@ func setInFile(path, key string, value any) error {
 		return fmt.Errorf("failed to read config file: %w", err)
 	}
 
-	var doc map[string]any
+	doc := map[string]any{}
 	if err := toml.Unmarshal(data, &doc); err != nil {
 		return fmt.Errorf("failed to parse config file: %w", err)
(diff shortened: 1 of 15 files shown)


gtsiolis:
Could you take a look since you added https://github.com/localstack/lstk/pull/136, @silv-io?

Comments or commits, welcome!

internal/config/config.go
```diff
@@ -110,14 +116,28 @@ func Set(key string, value any) error {
...
 func DisableUpdatePrompt() error {
-	return Set("update_prompt", false)
+	return Set("cli.update_prompt", false)
+}
+
+func SetUpdateSkippedVersion(version string) error {
+	return Set("cli.update_skipped_version", version)
+}
+
+func GetUpdateSkippedVersion() string {
```
anisaoshafi: this is unused 👀

internal/update/notify.go
```diff
@@ -12,19 +12,21 @@ import (
...
-	UpdatePrompt   bool
-	PersistDisable func() error
+	GitHubToken        string
+	UpdatePrompt       bool
+	SkippedVersion     string
+	PersistSkipVersion func(version string) error
 }
 
-const checkTimeout = 500 * time.Millisecond
+const checkTimeout = 2 * time.Second
```
anisaoshafi, line 21: nit: PR description says the timeout was reduced

internal/ui/app.go
```diff
@@ -296,6 +316,19 @@ func (a *App) flushBufferedLines() {
...
+	// Special handling for lstk update prompt
+	if len(req.Options) > 0 && strings.Contains(req.Options[0].Label, "Update now") {
+		checkmark := styles.Success.Render(output.SuccessMarker())
+		switch selectedKey {
+		case "u":
+			return checkmark + " Updating lstk..."
+		case "s":
+			return checkmark + " Skipped this version"
+		case "n":
+			return checkmark + " Won't ask again"
```
anisaoshafi: seems like this code is unreachable 🤔

anisaoshafi (commented):
The UX looks really nice, thanks George 🤩 

I left a few comments

internal/config/default_config.toml
```diff
@@ -1,6 +1,10 @@
 # lstk configuration file
 # Run 'lstk config path' to see where this file lives.
 
+# CLI settings
+[cli]
+update_prompt = true
+
```
carole-lavillonniere: nit: not sure we want this here since `true` is the default value anyways? Showing to users might invite them to disable it? Not sure what's right. If not documented here, we don't really have another place where we document all settings...

internal/config/config.go
```diff
@@ -110,14 +116,28 @@ func Set(key string, value any) error {
...
 	if err := viper.Unmarshal(&cfg); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal config: %w", err)
 	}
+	if !viper.InConfig("cli.update_prompt") && viper.InConfig("update_prompt") {
+		cfg.CLI.UpdatePrompt = viper.GetBool("update_prompt")
+	}
+	if !viper.InConfig("cli.update_skipped_version") && viper.InConfig("update_skipped_version") {
+		cfg.CLI.UpdateSkippedVersion = viper.GetString("update_skipped_version")
+	}
 	for i := range cfg.Containers {
```
carole-lavillonniere, line 166: issue: this looks like migration code, is this right? 
- `update_prompt`: I think we could ignore the fact that some users previously had this setting one level up in the config. They would just be asked again and in exchange we get rid of some complexity.
- `update_skipped_version`: This is introduced in this PR so there is nothing to migrate.
In short: can we remove these lines?

internal/ui/app.go
```diff
@@ -95,6 +95,26 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+					if idx >= 0 && idx < len(a.pendingInput.Options) {
+						opt := a.pendingInput.Options[idx]
+						a.lines = appendLine(a.lines, styledLine{text: formatResolvedInput(*a.pendingInput, opt.Key)})
+						responseCmd := sendInputResponseCmd(a.pendingInput.ResponseCh, output.InputResponse{SelectedKey: opt.Key})
+						a.pendingInput = nil
+						a.inputPrompt = a.inputPrompt.Hide()
+						return a, responseCmd
+					}
+				}
+			}
```
carole-lavillonniere, line 125: nit: the `Update` method is already doing a lot and the PR adds another ~20 lines. Can we extract this in a new method, something like:
```
func (a App) handleVerticalPromptKey(msg tea.KeyMsg) (App, tea.Cmd, bool) {
      ...
      return a, nil, false
}
```

internal/ui/app.go
```diff
@@ -316,6 +349,10 @@ func formatResolvedInput(req output.UserInputRequestEvent, selectedKey string) s
 	return fmt.Sprintf("%s %s", firstLine, selected)
 }
 
+func componentsUsesVerticalPrompt(prompt components.InputPrompt, options []output.InputOption) bool {
+	return prompt.Visible() && len(options) > 1 && strings.Contains(options[0].Label, "[")
```
carole-lavillonniere: issue:

> `strings.Contains(options[0].Label, "[")`

We should not rely on string parsing. Can we add something in the model that indicates that this user input should be displayed vertically? Maybe as simple as:
```
 type UserInputRequestEvent struct {
      Prompt     string
      Options    []InputOption
      ResponseCh chan<- InputResponse
      Vertical   bool
  }
```
Not sure if it would help but we might want to add this to CLAUDE.md since it's come back a few times.

internal/ui/app.go
```diff
@@ -296,6 +316,19 @@ func (a *App) flushBufferedLines() {
...
+		checkmark := styles.Success.Render(output.SuccessMarker())
+		switch selectedKey {
+		case "u":
+			return checkmark + " Updating lstk..."
+		case "s":
+			return checkmark + " Skipped this version"
+		case "n":
+			return checkmark + " Won't ask again"
+		}
+	}
```
carole-lavillonniere, line 342: issue: this file is generic UI rendering code, it should not know anything about the update logic.
We need to find another solution for this. Maybe a confirmation text can be added on the input options.

gtsiolis:
Tried to address all review comments in one commit, haven't tested yet. I will re-request reviews after testing, unless someone wants to pick this up in the meantime. ⚠️

internal/update/notify.go
```diff
@@ -89,20 +95,21 @@ func promptAndUpdate(ctx context.Context, sink output.Sink, githubToken string,
...
-			if err := persistDisable(); err != nil {
-				output.EmitWarning(sink, fmt.Sprintf("Failed to save preference: %v", err))
-			}
-		}
+	case "r":
 		return false
-	default:
+	case "s":
+		if opts.PersistSkipVersion != nil {
+			_ = opts.PersistSkipVersion(latest)
```
carole-lavillonniere: issue: we should not ignore the returned error here. Can we emit a warning upon error just like we used to?

internal/config/config.go
```diff
@@ -113,13 +116,111 @@ func resolvedConfigPath() string {
...
 func DisableUpdatePrompt() error {
 	return Set("cli.update_prompt", false)
 }
 
+func SetUpdateSkippedVersion(version string) error {
+	return Set("cli.update_skipped_version", version)
+}
+
+func GetUpdateSkippedVersion() string {
+	return viper.GetString("cli.update_skipped_version")
```
carole-lavillonniere: issue: this is not used. We don't really need getters for config, since we access directly the values from the `C`LIConfig struct.

internal/config/config.go
```diff
@@ -109,18 +116,119 @@ func resolvedConfigPath() string {
...
+		result = append(result, targetLine)
+	}
+
+	output := strings.Join(result, "\n")
+	if !strings.HasSuffix(output, "\n") {
+		output += "\n"
+	}
+
+	return os.WriteFile(path, []byte(output), 0644)
+}
```
carole-lavillonniere: suggestion: can we use the `go-toml/v2` dependency to achieve updating a single value, instead of writing that func from scratch? I think it would remove a lot of complexity/fragility here.

internal/config/config.go
```diff
@@ -1,22 +1,29 @@
...
+	"strings"
 
 	"github.com/spf13/viper"
 )
 
 //go:embed default_config.toml
 var defaultConfigTemplate string
 
+type CLIConfig struct {
+	UpdatePrompt         bool   `mapstructure:"update_prompt"`
```
carole-lavillonniere: suggestion: as far as I understand, we only left this here for backward-compatibility. The new version will not let the users choose to never be prompted again for update from the UI. But we have this code here to make sure that users that have `update_prompt` set to true in their config, will indeed never be prompted again.
My suggestion is: can we remove this code completely and ignore if they have that setting? That would clean up the code quite a bit. That would also allow getting rid of the migration logic lines 229-231.

gtsiolis:
Thanks for taking a look, @carole-lavillonniere! Pushed a commit trying to address all comments. 

Please, take another look when you're free! 👀 

I noticed on skip [S], the template config is replaced. ⚠️ 

#### BEFORE
```sh
# lstk configuration file
# Run 'lstk config path' to see where this file lives.

# Each [[containers]] block defines an emulator instance.
# You can define multiple to run them side by side.
[[containers]]
type = "aws"     # Emulator type. Currently supported: "aws"
tag  = "latest"  # Docker image tag, e.g. "latest", "2026.03"

# Environment profiles let you group environment variables and reference
# them by name in one or more containers via the 'env' field above.
#
# Example variables based on commonly used current config options:
#
#   DEBUG=1                 - Enable verbose logging
#   PERSISTENCE=1           - Persist LocalStack state across restarts
#   ENFORCE_IAM=1           - Enable IAM policy enforcement
#   SERVICES=s3,sqs         - Limit services to load
#   EAGER_SERVICE_LOADING=1 - Preload services at startup
#
# See full list of configuration options:
# > https://docs.localstack.cloud/references/configuration/
#
# Example profiles:
#
# [env.debug]
# DEBUG = "1"
# PERSISTENCE = "1"
# ENFORCE_IAM = "1"
#
# [env.ci]
# SERVICES = "s3,sqs"
# EAGER_SERVICE_LOADING = "1"
```

#### AFTER

```sh
[cli]
update_skipped_version = 'v0.5.7'

[[containers]]
port = '4566'
tag = 'latest'
type = 'aws'
```

carole-lavillonniere:
> Thanks for taking a look, @carole-lavillonniere! Pushed a commit trying to address all comments.
> 
> Please, take another look when you're free! 👀
> 
> I noticed on skip [S], the template config is replaced. ⚠️
> #### BEFORE
> 
> ```shell
> # lstk configuration file
> # Run 'lstk config path' to see where this file lives.
> 
> # Each [[containers]] block defines an emulator instance.
> # You can define multiple to run them side by side.
> [[containers]]
> type = "aws"     # Emulator type. Currently supported: "aws"
> tag  = "latest"  # Docker image tag, e.g. "latest", "2026.03"
> port = "4566"    # Host port the emulator will be accessible on
> # volume = ""    # Host directory for persistent state (default: OS cache dir)
> # env = []       # Named environment profiles to apply (see [env.*] sections below)
> 
> # Environment profiles let you group environment variables and reference
> # them by name in one or more containers via the 'env' field above.
> #
> # Example variables based on commonly used current config options:
> #
> #   DEBUG=1                 - Enable verbose logging
> #   PERSISTENCE=1           - Persist LocalStack state across restarts
> #   ENFORCE_IAM=1           - Enable IAM policy enforcement
> #   SERVICES=s3,sqs         - Limit services to load
> #   EAGER_SERVICE_LOADING=1 - Preload services at startup
> #
> # See full list of configuration options:
> # > https://docs.localstack.cloud/references/configuration/
> #
> # Example profiles:
> #
> # [env.debug]
> # DEBUG = "1"
> # PERSISTENCE = "1"
> # ENFORCE_IAM = "1"
> #
> # [env.ci]
> # SERVICES = "s3,sqs"
> # EAGER_SERVICE_LOADING = "1"
> ```
> 
> #### AFTER
> 
> ```shell
> [cli]
> update_skipped_version = 'v0.5.7'
> 
> [[containers]]
> port = '4566'
> tag = 'latest'
> type = 'aws'
> ```

Pushed a few commits to fix the config being overwritten and to add a few tests.

carole-lavillonniere (approved):

PR #162, opened 2026-03-23
Windows: show correct Docker Desktop start command

anisaoshafi:
On Windows, lstk now detects if the user is running PowerShell or cmd.exe and suggests the correct command to start Docker Desktop. It used to show the PowerShell command regardless of shell.
We use the `PSModulePath` environment variable, which PowerShell always sets (cmd.exe does not). 

Also took advantage to suppress the verbose Windows Docker error message since it isn't actionable.


silv-io (approved):
LGTM!

PR #163, opened 2026-03-24
Migrate AWS emulator to localstack/localstack image

carole-lavillonniere:
## Motivation

LocalStack consolidated `localstack/localstack-pro` into `localstack/localstack`. The CLI needs to pull from the new image name.

## Changes

- Update `emulatorImages` map to use `localstack` instead of `localstack-pro` for the AWS emulator — affects both the Docker image pulled and the product name sent to the license API
- Add integration test asserting the container is started from `localstack/localstack:latest`
- Replace hardcoded real image/product names in unit tests with generic placeholders (`test-image:latest`, `test-product:1.2.3`) since those tests don't assert on the actual name
- Document in CLAUDE.md that integration tests are black-box and must not import internal packages

## Tests

- Existing unit tests pass with `test-image:latest` placeholders
- New `image` sub-test in `TestStartCommandSetsUpContainerCorrectly` verifies the container is started from `localstack/localstack:latest`
- `TestLicenseValidationSuccess` updated to expect `localstack` as the product name in the license request

Closes DRG-591

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -106,11 +106,6 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, opts Start
 		if err != nil {
 			return err
 		}
-		productName, err := c.ProductName()
-		if err != nil {
-			return err
-		}
-
 		containerPort, err := c.ContainerPort()
 		if err != nil {
 			return err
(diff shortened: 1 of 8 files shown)


internal/container/start.go
anisaoshafi, line 359: Will `containerConfig.ProductName` be empty here? Can we hard-code as well?

silv-io:
I'm wondering, this change should then also block users from using versions lower than 2026.3, because that won't be the pro image anymore, no?

carole-lavillonniere:
Closing for now because of the concern mentioned by @silv-io https://github.com/localstack/lstk/pull/163#issuecomment-4116522591

PR #165, opened 2026-03-24
Lower go directive to minimum required version

silv-io:
## Motivation

The `go` directive in `go.mod` was being bumped to the latest Go release weekly, but it specifies the **minimum** version required to build the module — not the target. Setting it to bleeding-edge made it harder for downstream packagers (AUR, Nix /cc @purcell) to build lstk without the latest Go toolchain.

## Changes

- Set `go` directive to `1.25.0` (the actual minimum, constrained by `golang.org/x/sys` and `golang.org/x/term`)
- Add explicit `toolchain go1.26.1` directive so CI and developers still use the latest Go via auto-download
- Update weekly-go-upgrade workflow to bump `toolchain` instead of `go` — packagers using `GOTOOLCHAIN=local` are unaffected by toolchain bumps
- CI keeps using `go-version-file: go.mod` (which reads the `toolchain` directive)

## Tests

- Existing CI will validate — no code changes, only build/workflow config
- The new `tidy` CI job will verify `go mod tidy` produces no diff

## Related

- [FLC-545](https://linear.app/localstack/issue/FLC-545/look-into-gomod-version)


purcell:
Nice! Yes, I can confirm this will be helpful downstream.

carole-lavillonniere (approved):

PR #168, opened 2026-03-24
Show container start guidance on more commands

gtsiolis:
Show actionable Docker errors consistently across all commands

When Docker isn't available, status, stop, and logs now all show a clear, styled error message with actionable next steps when running in interactive mode.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-25 at 01 45 05 | Screenshot 2026-03-25 at 01 42 28 |


cmd/root.go
```diff
@@ -185,6 +185,17 @@ func isInteractiveMode(cfg *env.Env) bool {
...
+// through the sink if not in interactive mode. Returns a SilentError to
+// suppress duplicate error printing.
+func checkRuntimeHealth(ctx context.Context, rt runtime.Runtime, cfg *env.Env) error {
+	if err := rt.IsHealthy(ctx); err != nil {
+		rt.EmitUnhealthyError(output.NewPlainSink(os.Stdout), err)
+		return output.NewSilentError(err)
+	}
+	return nil
+}
+
```
carole-lavillonniere: issue: the `cfg *env.Env` param can be removed since it is unused.

cmd/status.go
```diff
@@ -25,9 +25,19 @@ func newStatusCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
...
+				output.EmitError(output.NewPlainSink(os.Stdout), output.ErrorEvent{
+					Title:   "Docker is not available",
+					Summary: err.Error(),
+					Actions: []output.ErrorAction{
+						{Label: "See help:", Value: "lstk -h"},
+						{Label: "Install Docker:", Value: "https://docs.docker.com/get-docker/"},
+					},
+				})
+				return output.NewSilentError(err)
+			}
```
carole-lavillonniere: question: why this special "docker not available error" handling only for the status command?

internal/output/plain_format.go
```diff
@@ -4,6 +4,16 @@ import (
...
+
+	"github.com/charmbracelet/lipgloss"
+)
+
+var (
+	errorTitleStyle     = lipgloss.NewStyle().Bold(true).Foreground(lipgloss.Color("#C33820"))
+	errorSecondaryStyle = lipgloss.NewStyle().Foreground(lipgloss.Color("241"))
+	errorActionStyle    = lipgloss.NewStyle().Foreground(lipgloss.Color("69"))
+	errorValueStyle     = lipgloss.NewStyle().Bold(true)
+	errorMutedStyle     = lipgloss.NewStyle().Foreground(lipgloss.Color("245"))
```
carole-lavillonniere: issue: do we really mean to redeclare all these, since they already exist in `internal/ui/styles/styles.go`?

cmd/stop.go
```diff
@@ -28,6 +28,9 @@ func newStopCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			if err := checkRuntimeHealth(cmd.Context(), rt, cfg); err != nil {
+				return err
+			}
```
carole-lavillonniere, line 30: issue: it looks like the check is missing for the start command

carole-lavillonniere, line 30: issue: we should add a new integration test case for each command to test the behavior when docker is not healthy.

internal/output/plain_format.go
```diff
@@ -130,21 +140,27 @@ func formatMessageEvent(e MessageEvent) string {
...
+		sb.WriteString("\n")
+		for i, action := range e.Actions {
+			sb.WriteString("\n")
+			if i > 0 {
+				sb.WriteString(errorMutedStyle.Render(ErrorActionPrefix + action.Label + " " + action.Value))
+			} else {
+				sb.WriteString(errorActionStyle.Render(ErrorActionPrefix+action.Label+" "))
+				sb.WriteString(errorValueStyle.Render(action.Value))
+			}
+		}
```
carole-lavillonniere: question: I would like to hear what @silv-io says here. We were previously not doing any styling in non-interactive mode.

cmd/root.go
gtsiolis: Removed param. ✔️

cmd/status.go
gtsiolis: Extracted `dockerNotAvailableError` helper.

cmd/stop.go
gtsiolis: Fixed.

gtsiolis: Added tests covering start, stop, status, logs.

internal/output/plain_format.go
silv-io: The plain output should remain styling free IMO. If we want a non-interactive styled output I think that should be a separate sink

carole-lavillonniere: Why do you think so?
Right now it's binary: interactive TTY gets Bubble Tea, everything else gets plain. A third styled-but-non-interactive sink would need a trigger. What would it be? CI is a good example: no TTY, so non-interactive, but GitHub Actions and most CI systems do render ANSI colors fine.

silv-io: the trigger word there is "most" for me. I don't want to force color onto systems where it might not be rendered well. If we add color styling to the plain output, we should at the least check that it will handle `NO_COLOR` correctly and add a test for that.<br><br>Also, then I think the styling should live outside of the UI package and in its own "style" package to avoid potential circular dependencies in the future.

carole-lavillonniere: I think for now we could just remove colors/styling from plain sink and revisit later if we need a third kind of sink.
Fine for you @gtsiolis?
I'm wondering if this was intentional in the first place?

gtsiolis: The intention was to add similar UX for both start and status commands when Docker is not detected.

| start | status |
|--------|--------|
| Screenshot 2026-04-21 at 19 08 33 | Screenshot 2026-04-21 at 19 08 36 | 

Added 45f92f07758300a421fffe9c7652475bf4018d51 to try how the extra sink would look like, what do you think?

Happy to drop the colors for now. Easy revert.

carole-lavillonniere: There is a misunderstanding here. It makes sense to add similar UX for the status command, but for that we should use the TUI (but not add colors in non-interactive mode). I will take over the PR.

carole-lavillonniere: Pushed 8a96ecb2a391b964103eadcd6cf9a8229450276d and tested the output for all commands: 

image

carole-lavillonniere: It looks similar but without colors when running `--non-interactive`:
image

carole-lavillonniere (approved):

gtsiolis:
Looks good, thanks for picking this up, @carole-lavillonniere! 🏀

PR #169, opened 2026-03-25
Use stable image by default

anisaoshafi:
From now on, we set "stable" tag on config.toml by default. 
Validation for "stable" tag license is similar to "latest" tag.

Current default tag `latest` was causing UX issues, because it was frequently pulling the newest dev version from latest, so using stable will improve that situation.

_Limitation_: We're not changing "latest" with "stable" for customers that have lstk already, even if they update. The users can change it manually.

Towards DRG-663

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ lstk --config /path/to/config.toml start
 ```toml
 [[containers]]
 type = "aws"     # Emulator type. Currently supported: "aws"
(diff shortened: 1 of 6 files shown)


silv-io (approved):
LGTM!

carole-lavillonniere:
Can we please convert to draft and not merge this until we made a call about the semantic of the `latest` tag?

anisaoshafi:
Closing for now, will reopen if necessary

PR #170, opened 2026-03-25
Add automated weekly patch release

carole-lavillonniere:
## Motivation

Releases currently require a manual trigger. This automates patch releases every Thursday so we ship consistently without manual intervention.

## Changes

- Add `automated-release.yml` workflow: runs every Thursday at 10 AM UTC, checks for commits since the last tag, runs CI fresh, and creates a patch tag if everything is green
- Make `ci.yml` callable as a reusable workflow (`workflow_call`) so the automated release can run CI before tagging
- Make `create-release-tag.yml` callable as a reusable workflow (`workflow_call`) so the automated release reuses the existing tag creation logic without duplication

## Tests

Not tested yet — `workflow_dispatch` requires the workflow to exist on the default branch, so it can only be triggered after this is merged:
```
gh workflow run automated-release.yml --repo localstack/lstk
```

Closes DRG-667

### DIFF
--- a/.github/workflows/automated-release.yml
+++ b/.github/workflows/automated-release.yml
@@ -1,0 +1,69 @@
+name: Automated Weekly Release
+
+on:
+  schedule:
+    - cron: "0 10 * * 4"  # Every Thursday at 10 AM UTC
+  workflow_dispatch:
+
+permissions:
+  contents: write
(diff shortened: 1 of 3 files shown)


silv-io (approved):
LGTM!

PR #171, opened 2026-03-26
Fix automated releases

carole-lavillonniere:
## Motivation

The first run of the automated weekly release (#170) failed because GitHub Actions does not reliably queue a job that uses `uses:` (reusable workflow call) when it depends via `needs` on another job that also uses `uses:`. The `create-tag` job never appeared in the run.

## Changes

- Extract tag creation logic into a composite action (`.github/actions/create-release-tag/`) shared by both workflows
- Convert `create-tag` in `automated-release.yml` back to a regular job that calls the composite action — fixing the GitHub limitation without duplicating logic
- Simplify `create-release-tag.yml` to use the same composite action; remove the `workflow_call` trigger that is no longer needed

## Tests

Cannot be tested from a feature branch: the `create-tag` job always checks out `main`, so the composite action file must exist on `main` to be found. This PR must be merged first, then triggered via:

```
gh workflow run automated-release.yml --repo localstack/lstk
```

Towards DRG-667

### DIFF
--- a/.github/actions/create-release-tag/action.yml
+++ b/.github/actions/create-release-tag/action.yml
@@ -1,0 +1,64 @@
+name: Create Release Tag
+description: Compute the next version tag, create it, and push it.
+
+inputs:
+  bump:
+    description: Version bump type (patch or minor)
+    required: true
+
+outputs:
+  tag:
+    description: The created tag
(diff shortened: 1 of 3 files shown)


anisaoshafi (approved):
thanks for the quick fix :shipit:

PR #173, opened 2026-03-27
Improve config error for legacy YAML config files

silv-io:
## Motivation

When an old `config.yaml` (from a previous lstk version) exists in the config search path, viper picks it up and tries to parse it as TOML (due to `SetConfigType("toml")`), resulting in a vague error like `failed to read config file: While parsing config: toml: expected character =` with no hint about which file was read or why.

## Changes

- After `ReadInConfig()` fails with a non-NotFound error, check `viper.ConfigFileUsed()` (populated by viper before parsing, even on failure) to detect if a `.yaml`/`.yml` file was picked up
- If so, return a targeted error naming the file and explaining that lstk now uses TOML format
- Add integration test `TestLegacyYAMLConfigGivesHelpfulError` that places a legacy `config.yaml` in `~/.config/lstk/` and asserts the error mentions both the filename and TOML

## Tests

New integration test added; verified it fails before the fix and passes after.

## Related

- [FLC-544](https://linear.app/localstack/issue/FLC-544/improve-config-error-for-legacy-yaml-config-files)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -63,6 +63,9 @@ func Init() error {
 	if err := viper.ReadInConfig(); err != nil {
 		var notFoundErr viper.ConfigFileNotFoundError
 		if !errors.As(err, &notFoundErr) {
+			if used := viper.ConfigFileUsed(); filepath.Ext(used) == ".yaml" || filepath.Ext(used) == ".yml" {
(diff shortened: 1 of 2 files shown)


anisaoshafi (approved):
Nice catch! 👏🏼

PR #174, opened 2026-03-27
Fix docker socket mount for Windows

anisaoshafi:
Attempt to fix the docker socket mount for Windows. 
Currently it evaluates as "", and as a result users can't work with Lambda and other services that spin containers.


anisaoshafi:
@silv-io would appreciate a review here when you have some time 🙏🏼

silv-io (approved):

PR #176, opened 2026-03-27
Add AWS CLI proxy command

gtsiolis:
Updates the AWS CLI proxy command to show a loading spinner while waiting for a response, then automatically hide it as soon as output starts streaming.
  - Wrapped stdout/stderr writers to detect first output byte
  - Spinner stops and clears itself immediately when AWS CLI begins responding
  - Output streams in real-time (no buffering)
  - Non-interactive mode unchanged (no spinner)

See DES-190 and DRG-364 for more context.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-27 at 15 41 05 | Screenshot 2026-03-27 at 15 41 11 |

### DIFF
--- a/cmd/aws.go
+++ b/cmd/aws.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"io"
 	"os"
 
@@ -10,6 +11,7 @@ import (
 	"github.com/localstack/lstk/internal/endpoint"
 	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/runtime"
 	"github.com/localstack/lstk/internal/telemetry"
 	"github.com/localstack/lstk/internal/terminal"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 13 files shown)


anisaoshafi (commented):
Thanks George for investing in this nice feature ⭐ 

1. **Issue**: The spinner behaviour is weird, it gets stuck e.g:
- a) when you try `lstk aws s3 ls` when you have no bucket
- b) create a bucket and then delete bucket. I suggest to revisit it to make it more robust 🙏🏼 

Image

Image




2. **Suggestion**: would be nice to add a more descriptive message when LocalStack is not running:

Image

internal/terminal/spinner.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+var dotFrames = []string{"⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷"}
+
+// ANSI color codes matching lstk's spinner style (color 69 blue) and secondary (color 241 gray)
+const (
+	spinnerColor   = "\033[38;5;69m"
+	secondaryColor = "\033[38;5;241m"
+	resetColor     = "\033[0m"
+)
```
anisaoshafi: **issue**: didn't we have these properties defined somewhere for Nimbo? I think they're duplicated.

Also, In [components package](https://github.com/localstack/lstk/blob/main/internal/ui/components/spinner.go), we have spinner defined. Why are we redefining it again in this PR, can you try reuse them?

Btw 🎗️, we have this `/review-pr ` claude skill: `/review-pr #176` that the team added, and it can be run in local development to catch this architectural pattern issues 🙏🏼

carole-lavillonniere:
@gtsiolis the PR title and description are confusing to me. Aren't we adding a brand new `lstk aws` command here?

cmd/aws.go
```diff
@@ -0,0 +1,50 @@
...
+			host, _ := endpoint.ResolveHost(port, cfg.LocalStackHost)
+			return awscli.Exec(cmd.Context(), "http://"+host, args)
+		}),
+	}
+}
+
+func resolveAWSPort() string {
+	const defaultPort = "4566"
+	if err := config.Init(); err != nil {
+		return defaultPort
```
carole-lavillonniere: question: by swallowing the error here and returning a default port, we treat "no config file" as an acceptable state. Is it though? Shouldn't we always create the default config first? 
Also worth having an integration test case for this specific case.

carole-lavillonniere: Also if we really want to return the default port here, I would move the const definition to the right domain, something like:
```
  // internal/config/config.go
  const DefaultAWSPort = "4566"
```

internal/terminal/spinner.go
carole-lavillonniere: Hopefully we can reuse the color but I don't think we can reuse the spinner. 
The one we have requires Bubble tea and here we don't have Bubble tea because we are just passing through stdout/stderr.
I think a new spinner cannot be avoided, but maybe we can move it to a more generic place such as `internal/terminal` so it can be reused for other commands in the future.

internal/awscli/exec.go
```diff
@@ -0,0 +1,90 @@
...
+		return fmt.Errorf("aws CLI not found in PATH — install it from https://aws.amazon.com/cli/")
+	}
+
+	cmdArgs := make([]string, 0, len(args)+2)
+	cmdArgs = append(cmdArgs, "--endpoint-url", endpointURL)
+	cmdArgs = append(cmdArgs, args...)
+
+	cmd := exec.CommandContext(ctx, awsBin, cmdArgs...)
+	cmd.Stdin = os.Stdin
+	cmd.Env = BuildEnv(os.Environ())
```
carole-lavillonniere: question: we already setup the aws profile as part of `lstk start`. Did we add this so it works when user has not run `lstk start` yet? Can we instead reuse that logic instead of reinventing it?

internal/output/errors.go
```diff
@@ -26,3 +26,32 @@ func IsSilent(err error) bool {
...
+
+// ExitCode returns the exit code if err (or any error in its chain) is an ExitCodeError,
+// or 1 as a default.
+func ExitCode(err error) int {
+	var exitErr *ExitCodeError
+	if errors.As(err, &exitErr) {
+		return exitErr.Code
+	}
+	return 1
+}
```
carole-lavillonniere: issue: this new `ExitCodeError` struct does not seem needed. `exec.ExitError` (from go standard lib, returned by `cmd.Run()` when the process exits with a non-zero status) already carries the exit code via `.ExitCode()`. 

I think the whole `ExitCodeError` type and  `output.ExitCode()` helper can be dropped.


You'd have to do this in `main.go`:

```
  -     os.Exit(output.ExitCode(err))
  +     var exitErr *exec.ExitError
  +     if errors.As(err, &exitErr) {
  +             os.Exit(exitErr.ExitCode())
  +     }
  +     os.Exit(1)
```

And this in `internal/awscli/exec.go`:

```
  -     if errors.As(err, &exitErr) {
  -             return output.NewSilentError(output.NewExitCodeError(exitErr.ExitCode(), err))
  -     }
  +     if errors.As(err, &exitErr) {
  +             return output.NewSilentError(err)
  +     }

```

gtsiolis:
> The spinner behaviour is weird,

Great catch, @anisaoshafi! Removed the loading state when no buckets are there and when deleting buckets to recreate the UX of the AWS CLI itself.

> would be nice to add a more descriptive message when LocalStack is not running:

Agree, we can have a better control of this, but kept it as is for now to recreate the AWS CLI UX. Thoughts?

---

@carole-lavillonniere I pushed bf06946 to attempt to resolve the comments. 👀

carole-lavillonniere:
Pushed a few commits  to address:
### localstack aws profile not configured
no-aws-profile

### localstack aws emulator not running
ls-not-running

### docker daemon not running
docker-not-running


Note: The error messages use a plain sink (no TUI) for now, because `lstk aws` is a passthrough that streams the real aws CLI output and there's no TUI session to attach to. We could setup Bubble Tea just for the error messages (before the aws CLI execution) but it sounds overkill for now. We can revisit later.


@gtsiolis

carole-lavillonniere (approved):

gtsiolis (commented):
Let's ship this! :shipit:

PR #178, opened 2026-03-27
Fix errcheck lint violations in start_test.go

gtsiolis:
Check error return values from resp.Body.Close() in HTTP health endpoint tests.

Cc @carole-lavillonniere because https://github.com/localstack/lstk/pull/156


anisaoshafi (approved):
TIL why I was getting these linting issues in local

gtsiolis:
Thanks for taking a look, @anisaoshafi! 🙌

PR #180, opened 2026-03-28
Show plan name and config path in header

gtsiolis:
This will make the header contain more useful and valuable information:
1. Display the user's plan name in the TUI header by fetching it from the API, with a loading animation.
1. Replace the emulator type which is less actionable, and the endpoint is already shown in the output below.
1. Cache the resolved plan label locally so subsequent runs show it instantly without the loading state.
1. Replace the endpoint in the header with the resolved config file path (project-relative, ~/-prefixed, or absolute).
1. Return LicenseResponse from GetLicense so callers can extract the license_type field for known plan tiers.
1. Show "LocalStack (No license)" when the plan cannot be determined (no token, API error, or no license assigned).

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-03-28 at 03 17 57 | Screenshot 2026-03-28 at 03 18 27 |


cmd/root.go
```diff
@@ -181,6 +200,56 @@ func commandWithTelemetry(name string, tel *telemetry.Client, fn func(*cobra.Com
 	}
 }
 
+// resolveEmulatorLabel tries to fetch the plan name from the license API
+// to build a label like "LocalStack Ultimate". Falls back to
+// "LocalStack (No license)" when the plan cannot be determined.
+func resolveEmulatorLabel(ctx context.Context, client api.PlatformAPI, containers []config.ContainerConfig, token string, logger log.Logger) string {
```
anisaoshafi: could we move this business logic function to an internal package to comply with the architecture rules?

internal/ui/run.go
```diff
@@ -25,24 +24,30 @@ func (s programSender) Send(msg any) {
 	s.p.Send(msg)
 }
 
-func Run(parentCtx context.Context, rt runtime.Runtime, version string, opts container.StartOptions, notifyOpts update.NotifyOptions) error {
+func Run(parentCtx context.Context, rt runtime.Runtime, version string, opts container.StartOptions, notifyOpts update.NotifyOptions, configPath, emulatorLabel string, labelCh <-chan string, animateHeader bool) error {
```
anisaoshafi: suggestion: the signature has become huge with the new parameters. feel free to move some options in a struct to make it easier to call

internal/api/client.go
```diff
@@ -65,6 +65,32 @@ type MachineInfo struct {
 	PlatformRelease string `json:"platform_release,omitempty"`
 }
 
+type LicenseResponse struct {
+	LicenseType string `json:"license_type"`
+}
+
+var planDisplayNames = map[string]string{
```
anisaoshafi, line 72: style: we have a code-style rule to not have package-level global variables (CONTRIBUTING.md). could you make it local or move inside the function that uses it?

cmd/root.go
```diff
@@ -127,11 +128,29 @@ func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *t
...
+			config.CachePlanLabel(label)
+			labelCh <- label
+		}()
+
+		return ui.Run(ctx, rt, version.Version(), opts, notifyOpts, configPath, cached, labelCh, animateHeader)
 	}
 
 	sink := output.NewPlainSink(os.Stdout)
+	if configPath != "" {
+		output.EmitInfo(sink, configPath)
```
anisaoshafi: ux: when run in non-interactive mode it shows the config file path only, which doesn't look intuitive what it refers to. maybe we can remove it altogether for non-interactive mode 👀

anisaoshafi (commented):
I like this design better, makes is more multi-emulator friendly 👏🏼 

This looks good. I left a few comments and am happy to recheck this again 😃

anisaoshafi (approved):

anisaoshafi:
Pushed a few commits to fix the tests and some improvements. LGTM!

PR #181, opened 2026-03-28
Show endpoint and tips when emulator is already running

gtsiolis:
This will:
1. Show the endpoint, web app URL, and a random tip when the emulator is already running, matching the start output.
1. Render the "already running" message as a note (> Note:) instead of a plain info line

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-04-21 at 18 57 44 | Screenshot 2026-04-21 at 18 57 49 |

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -226,7 +226,7 @@ func runPostStartSetups(ctx context.Context, sink output.Sink, containers []conf
 		if setup, ok := setups[t]; ok {
 			resolvedHost, dnsOK := endpoint.ResolveHost(firstByType[t].Port, localStackHost)
 			if !dnsOK {
-				output.EmitNote(sink, `Could not resolve "localhost.localstack.cloud" — your system may have DNS rebind protection enabled. Using 127.0.0.1 as the endpoint.`)
+				output.EmitNote(sink, endpoint.DNSRebindNote)
 			}
 			if err := setup(ctx, sink, interactive, resolvedHost); err != nil {
 				return err
@@ -363,7 +363,7 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 			output.EmitNote(sink, "LocalStack is already running")
 			resolvedHost, dnsOK := endpoint.ResolveHost(c.Port, localStackHost)
 			if !dnsOK {
-				output.EmitNote(sink, `Could not resolve "localhost.localstack.cloud" — your system may have DNS rebind protection enabled. Using 127.0.0.1 as the endpoint.`)
+				output.EmitNote(sink, endpoint.DNSRebindNote)
 			}
 			emitPostStartPointers(sink, resolvedHost, webAppURL)
 			continue
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -361,7 +361,10 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 		}
 		if running {
 			output.EmitNote(sink, "LocalStack is already running")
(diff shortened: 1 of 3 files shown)


internal/ui/app.go
```diff
@@ -141,15 +143,22 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 		a.spinner, _ = a.spinner.Stop()
 		return a, nil
+	case hideHeaderMsg:
+		a.hideHeader = true
+		return a, nil
 	case output.MessageEvent:
 		msgCopy := msg
 		line := styledLine{text: components.RenderMessage(msg), message: &msgCopy}
+		var cmd tea.Cmd
+		if msg.Severity == output.SeverityNote {
```
anisaoshafi: UX wise looks good 🧹 

Question: This is going to hide the TUI header for all SeverityNote messages, not only in the case emulator is running. Is this expected @gtsiolis?

anisaoshafi (approved):

internal/container/start.go
```diff
@@ -352,15 +352,17 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink,
...
 	var filtered []runtime.ContainerConfig
 	for _, c := range containers {
 		running, err := rt.IsRunning(ctx, c.Name)
 		if err != nil && !errdefs.IsNotFound(err) {
 			return nil, fmt.Errorf("failed to check container status: %w", err)
 		}
 		if running {
-			output.EmitInfo(sink, "LocalStack is already running")
+			output.EmitNote(sink, "LocalStack is already running")
+			resolvedHost, _ := endpoint.ResolveHost(c.Port, localStackHost)
```
carole-lavillonniere: We should check the result `dnsOK` of `ResolveHost` like we do line 227 in current file

internal/ui/app.go
gtsiolis: Good point, @anisaoshafi. Now that the banner is more functional with plan and config path with https://github.com/localstack/lstk/pull/180, we can leave it as is for now. In the future we may add a boolean to hie the banner per case. 

Cc @carole-lavillonniere

internal/container/start.go
gtsiolis: Nice one, added in 41add77.

gtsiolis:
Thanks both, @anisaoshafi and @carole-lavillonniere!

internal/container/start.go
```diff
@@ -352,15 +352,20 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink,
...
 		running, err := rt.IsRunning(ctx, c.Name)
 		if err != nil && !errdefs.IsNotFound(err) {
 			return nil, fmt.Errorf("failed to check container status: %w", err)
 		}
 		if running {
-			output.EmitInfo(sink, "LocalStack is already running")
+			output.EmitNote(sink, "LocalStack is already running")
+			resolvedHost, dnsOK := endpoint.ResolveHost(c.Port, localStackHost)
+			if !dnsOK {
+				output.EmitNote(sink, `Could not resolve "localhost.localstack.cloud" — your system may have DNS rebind protection enabled. Using 127.0.0.1 as the endpoint.`)
```
anisaoshafi: nit: this same message is being used in two other places already, could you please put it in constant and reuse it? 🙏🏼

anisaoshafi (approved):
LGTM! ✨

internal/container/start.go
gtsiolis: Changed in e744cdf.

PR #183, opened 2026-04-02
Add volume path and clear subcommands

gtsiolis:
This will add `lstk volume` command group with two subcommands:
1. `lstk volume path`: prints the resolved volume directory path for each configured emulator
2. `lstk volume clear`: removes all data from emulator volume directories to reset cached state
    - `volume clear` asks for interactive confirmation by default; requires `--force` in non-interactive mode
    - `volume clear` supports `--type` to target a specific emulator (e.g., `--container aws`)

| `volume path` | `volume clear` |
|--------|--------|
| Screenshot 2026-04-02 at 14 03 30 | Screenshot 2026-04-02 at 14 03 35 | 

Fix DES-199, DES-198


cmd/volume.go
```diff
@@ -0,0 +1,105 @@
...
+				sink := output.NewPlainSink(os.Stdout)
+				return volume.Clear(cmd.Context(), sink, containers, true)
+			}
+
+			return ui.RunVolumeClear(cmd.Context(), containers)
+		}),
+	}
+
+	cmd.Flags().BoolVar(&force, "force", false, "Skip confirmation prompt")
+	cmd.Flags().StringVar(&containerName, "container", "", "Target a specific emulator container name")
```
anisaoshafi: @gtsiolis Is `container` the right word for the flag or should we use `emulator` for consistency?

gtsiolis: Nice catch, thanks! Changed in 02e75e8. Could you take another look when you're free, @anisaoshafi?

carole-lavillonniere: I am also unsure about the value of `--emulator localstack-aws`. "localstack-aws" is not something referenced anywhere else, how would a user guess it? Suggestion: `--emulator aws`or `--type aws` which matches what we have in the config: 
image

carole-lavillonniere: Maybe in the future for interactive mode: offer a list to choose the container from

cmd/volume.go
```diff
@@ -0,0 +1,105 @@
...
+				sink := output.NewPlainSink(os.Stdout)
+				return volume.Clear(cmd.Context(), sink, containers, true)
+			}
+
+			return ui.RunVolumeClear(cmd.Context(), containers)
+		}),
+	}
+
+	cmd.Flags().BoolVar(&force, "force", false, "Skip confirmation prompt")
+	cmd.Flags().StringVar(&containerName, "emulator", "", "Target a specific emulator container name")
```
carole-lavillonniere: We should also update the description here

test/integration/volume_test.go
```diff
@@ -0,0 +1,180 @@
...
+		requireExitCode(t, 0, err)
+
+		assertCommandTelemetry(t, events, "volume clear", 0)
+	})
+}
+
+// escapeTomlPath escapes backslashes for Windows paths in TOML quoted strings.
+func escapeTomlPath(path string) string {
+	return strings.ReplaceAll(path, `\`, `\\`)
+}
```
carole-lavillonniere: Can we add some integration tests for the interactive path (with PTY)?

internal/volume/clear.go
```diff
@@ -0,0 +1,120 @@
...
+				return err
+			}
+			size += info.Size()
+		}
+		return nil
+	})
+	return size, err
+}
+
+func formatSize(bytes int64) string {
```
carole-lavillonniere: Instead of our custom implementation, could we use github.com/docker/go-units which is already in go.mod as an indirect dependency? units.BytesSize does exactly this and is tested/maintained.

internal/ui/run_volume_clear.go
```diff
@@ -0,0 +1,47 @@
...
+		return output.NewSilentError(app.Err())
+	}
+
+	runErr := <-runErrCh
+	if runErr != nil && !errors.Is(runErr, context.Canceled) {
+		return runErr
+	}
+
+	return nil
+}
```
carole-lavillonniere, line 15: This is duplicating code in `runWithTUI` (`run.go:106`). The only unique part is the `volume.Clear(...)` call. We can replace this with:

```go
func RunVolumeClear(parentCtx context.Context, containers []config.ContainerConfig) error {
    return runWithTUI(parentCtx, withoutHeader(), func(ctx context.Context, sink output.Sink) error {
        return volume.Clear(ctx, sink, containers, false)
    })
}
```

(`run_stop.go` has the same duplication issue)

cmd/volume.go
gtsiolis: Great point, @carole-lavillonniere! I previously used the `container` term to be more specific, but also would like to avoid surfacing container topology to users when possible. 

The `--type` suggestion sounds a better option. Replaced in 5a08c95.

internal/ui/run_volume_clear.go
gtsiolis: Nice catch! ⚾

gtsiolis:
Could you take another look, @carole-lavillonniere? 

Added also https://github.com/localstack/lstk/pull/183/changes/31be41e64de9f5d92fafd2ff18a29e637d2ff68a as some tests were failing in main locally.

carole-lavillonniere:
Tested locally and pushed aad62c857ac185a843a8ad0be3bd11135d262200 for this situation:
image
@gtsiolis or @anisaoshafi can you have a look and give me a :+1: ?
As far as I understood that would only happen on linux because docker does not run containers in a VM like it does on Windows and macOS.

carole-lavillonniere (approved):

gtsiolis:
Thanks, @carole-lavillonniere! I can reproduce the bug and fix in Linux. ✔️

PR #185, opened 2026-04-08
Cache and mount license file after validation

silv-io:
## Motivation

The CLI fetches a license file per emulator/version on every start. This caches it locally to avoid redundant fetches and mounts it into the container so the emulator doesn't need to re-fetch it either.

## Changes

- `GetLicense` now returns `([]byte, error)` so the license content can be used by the caller
- Added `config.LicenseFilePath()` — returns `$CACHE_DIR/lstk/license.json` (OS-appropriate)
- After successful license validation, writes the license file to the cache path
- Mounts the cached license file read-only into each container at `/etc/localstack/conf.d/license.json`

## Tests

Existing integration tests cover the license validation flow. Manual verification of cache write and container mount.

## Related

- [FLC-546](https://linear.app/localstack/issue/FLC-546/cache-and-mount-license-files)

### DIFF
--- a/cmd/logout.go
+++ b/cmd/logout.go
@@ -43,7 +43,8 @@ func newLogoutCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra
 			if err != nil {
 				return fmt.Errorf("failed to initialize token storage: %w", err)
 			}
-			a := auth.New(sink, platformClient, tokenStorage, cfg.AuthToken, "", false)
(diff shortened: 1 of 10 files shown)


anisaoshafi (commented):
Nice optimization 👏🏼 

Could you please clean up the cache file in `cleanupLicense`?
Also, would be great to add an integration test

anisaoshafi (approved):
LGTM 👏🏼

PR #186, opened 2026-04-09
Enable Snowflake emulator

anisaoshafi:
Enable support for Snowflake emulator.

ℹ️ Licensing checks are skipped for snowflake at the moment, there are efforts to improve this but not in lstk scope for now. Improvements will be incorporated by lstk when ready on the platform side of the things.

There is going to be follow up PRs to smoothen multi-emulator experience (PRO-198, DRG-381)

### DIFF
--- a/internal/config/default_config.toml
+++ b/internal/config/default_config.toml
@@ -12,12 +12,6 @@ port = "4566"    # Host port the emulator will be accessible on
 # volume = ""    # Host directory for persistent state (default: OS cache dir)
 # env = []       # Named environment profiles to apply (see [env.*] sections below)
 
-# To use the Snowflake emulator instead, comment out the AWS block above and uncomment this one.
-# [[containers]]
-# type = "snowflake"
-# tag  = "latest"
-# port = "4566"
-
 # Environment profiles let you group environment variables and reference
 # them by name in one or more containers via the 'env' field above.
 #
--- a/internal/config/default_config.toml
+++ b/internal/config/default_config.toml
@@ -2,7 +2,9 @@
 # Run 'lstk config path' to see where this file lives.
 
 # Each [[containers]] block defines an emulator instance.
-# You can define multiple to run them side by side.
+# Currently, running AWS and Snowflake at the same time is not fully supported —
+# enable only one [[containers]] block at a time.
+
 [[containers]]
 type = "aws"     # Emulator type. Currently supported: "aws", "snowflake"
 tag  = "latest"  # Docker image tag, e.g. "latest", "2026.03"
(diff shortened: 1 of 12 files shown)


internal/container/label.go
```diff
@@ -31,6 +31,9 @@ func ResolveEmulatorLabel(ctx context.Context, client api.PlatformAPI, container
 
 	tag := c.Tag
 	if tag == "" || tag == "latest" {
+		if c.Type == config.EmulatorSnowflake {
+			return "LocalStack Snowflake", false
```
anisaoshafi: The licensing for snowflake is complex and out of scope for now, it makes the TUI header inaccurate because we'd default to `(No License)` [1] as we've no easy way to determine it.

@gtsiolis what do you think about changing it to `LocalStack Snowflake` [2] for now? 

| 1 | 2 |
| Image | Image |

carole-lavillonniere (commented):
It all looks good, I'm just thinking we could add a few integration tests for the cases where the behavior is slightly different than for aws:
- status command
- start skips license validation
WDYT?

anisaoshafi:
> It all looks good, I'm just thinking we could add a few integration tests for the cases where the behavior is slightly different than for aws:
> 
> * status command
> * start skips license validation
>   WDYT?

I had some test for skipping license validation, but decided to remove them. Will add it back together with something for status, it's better to clarify what's currently supported in tests.

internal/config/default_config.toml
```diff
@@ -2,14 +2,22 @@
 # Run 'lstk config path' to see where this file lives.
 
 # Each [[containers]] block defines an emulator instance.
-# You can define multiple to run them side by side.
+# Currently, running AWS and Snowflake at the same time is not fully supported —
+# enable only one [[containers]] block at a time.
+
 [[containers]]
-type = "aws"     # Emulator type. Currently supported: "aws"
+type = "aws"     # Emulator type. Currently supported: "aws", "snowflake"
```
gtsiolis, line 9: **praise**: Thanks for updating the config! 😁

internal/config/default_config.toml
```diff
@@ -2,14 +2,22 @@
...
 port = "4566"    # Host port the emulator will be accessible on
 # volume = ""    # Host directory for persistent state (default: OS cache dir)
 # env = []       # Named environment profiles to apply (see [env.*] sections below)
 
+# To use the Snowflake emulator instead, comment out the AWS block above and uncomment this one.
+# [[containers]]
+# type = "snowflake"
+# tag  = "latest"
+# port = "4566"
+
```
gtsiolis: **suggestion**: I'd leave default config do the explaining, to avoid also encouraging multiple containers. Thoughts?

```suggestion
```

internal/container/label.go
gtsiolis: Totally valid, we need to clarify this on the product side. Can we ...
1. Show _LocalStack_ here until we clarify the licensing? This would avoid leaking back confusing terminology that does not exist, like _LocalStack Snowflake_.
2. Hide the tip row when running SNO, until we figure out the licensing parts if it's trivial?

internal/config/default_config.toml
anisaoshafi: Good point, thanks George. Done ✅

anisaoshafi:
@gtsiolis thanks for your feedback. Feel free to take another look 🙏🏼

carole-lavillonniere (approved):

gtsiolis (approved):

anisaoshafi:
ℹ️ The default tests timeout is 10 mins for go. The tests in ubuntu are taking longer than 10 min now, so increased the limit to 15m for now to unblock my PR.
We'll try to parallelize the tests, but it's out of scope for this PR 🤞🏼

PR #190, opened 2026-04-14
Attach to externally-started LS container

carole-lavillonniere:
## Summary
- `lstk start` now attaches to an already-running LocalStack container instead of failing. It fails if the port is different than the configured one.
- `lstk stop` locates and stops an externally-started container
- `lstk status` returns status of an externally-started container if any

Towards DRG-553

## Started outside of `lstk` (eg CLI v1, or locally built and ran in docker):
running-with-clliv1

## Already running on a different port:
running-other-port

## Status command shows details about the version and container name:
running-local-build

## Possible improvement:
Add interactivity and let user decide if the existing container should be picked up or replaced

### DIFF
--- a/internal/container/running.go
+++ b/internal/container/running.go
@@ -23,8 +23,6 @@ func AnyRunning(ctx context.Context, rt runtime.Runtime, containers []config.Con
 	return false, nil
 }
 
-// First checks the configured name,
-// then falls back to FindRunningByImage for containers started outside lstk
 func resolveRunningContainerName(ctx context.Context, rt runtime.Runtime, c config.ContainerConfig) (string, error) {
 	running, err := rt.IsRunning(ctx, c.Name())
 	if err != nil {
--- a/internal/container/running.go
+++ b/internal/container/running.go
@@ -45,7 +45,7 @@ func resolveRunningContainerName(ctx context.Context, rt runtime.Runtime, c conf
 		return "", err
 	}
 
-	found, err := rt.FindRunningByImage(ctx, imageRepo, containerPort)
+	found, err := rt.FindRunningByImage(ctx, []string{imageRepo, "localstack/localstack"}, containerPort)
 	if err != nil {
 		return "", fmt.Errorf("failed to scan for running containers: %w", err)
 	}
--- a/internal/container/running.go
+++ b/internal/container/running.go
@@ -45,7 +45,7 @@ func resolveRunningContainerName(ctx context.Context, rt runtime.Runtime, c conf
 		return "", err
 	}
 
-	found, err := rt.FindRunningByImage(ctx, imageRepo, containerPort, c.Port)
+	found, err := rt.FindRunningByImage(ctx, imageRepo, containerPort)
 	if err != nil {
 		return "", fmt.Errorf("failed to scan for running containers: %w", err)
 	}
--- a/internal/container/running.go
+++ b/internal/container/running.go
@@ -3,6 +3,7 @@ package container
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/localstack/lstk/internal/config"
 	"github.com/localstack/lstk/internal/runtime"
@@ -10,14 +11,47 @@ import (
 
 func AnyRunning(ctx context.Context, rt runtime.Runtime, containers []config.ContainerConfig) (bool, error) {
 	for _, c := range containers {
-		running, err := rt.IsRunning(ctx, c.Name())
+		name, err := resolveRunningContainerName(ctx, rt, c)
 		if err != nil {
-			return false, fmt.Errorf("checking %s running: %w", c.Name(), err)
+			return false, err
 		}
-		if running {
+		if name != "" {
 			return true, nil
 		}
 	}
 
 	return false, nil
+}
+
+// First checks the configured name,
+// then falls back to FindRunningByImage for containers started outside lstk
+func resolveRunningContainerName(ctx context.Context, rt runtime.Runtime, c config.ContainerConfig) (string, error) {
+	running, err := rt.IsRunning(ctx, c.Name())
+	if err != nil {
+		return "", fmt.Errorf("checking %s running: %w", c.Name(), err)
+	}
+	if running {
+		return c.Name(), nil
+	}
+
+	image, err := c.Image()
+	if err != nil {
+		return "", err
+	}
+	imageRepo, _, _ := strings.Cut(image, ":")
+
+	containerPort, err := c.ContainerPort()
+	if err != nil {
+		return "", err
+	}
+
+	found, err := rt.FindRunningByImage(ctx, imageRepo, containerPort, c.Port)
+	if err != nil {
+		return "", fmt.Errorf("failed to scan for running containers: %w", err)
+	}
+	if found != nil {
+		return found.Name, nil
+	}
+
+	return "", nil
 }
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -372,7 +372,34 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 			emitPostStartPointers(sink, resolvedHost, webAppURL)
 			continue
 		}
+
+		imageRepo, _, _ := strings.Cut(c.Image, ":")
+		found, err := rt.FindRunningByImage(ctx, imageRepo, c.ContainerPort, c.Port)
+		if err != nil {
+			return nil, fmt.Errorf("failed to scan for running containers: %w", err)
+		}
+		if found != nil {
+			runningTag := imageTagFrom(found.Image)
+			configTag := c.Tag
+			if configTag == "" {
+				configTag = "latest"
+			}
+			if runningTag != configTag {
+				output.EmitWarning(sink, fmt.Sprintf(
+					"Found running LocalStack %s, config specifies %s — using the running instance",
+					runningTag, configTag,
+				))
+			} else {
+				output.EmitInfo(sink, "LocalStack is already running")
+			}
+			continue
+		}
+
 		if err := ports.CheckAvailable(c.Port); err != nil {
+			if info, infoErr := fetchLocalStackInfo(ctx, c.Port); infoErr == nil {
+				emitLocalStackAlreadyRunningWarning(sink, c.Port, info.Version, c.Tag)
+				continue
+			}
 			emitPortInUseError(sink, c.Port)
 			emitEmulatorStartError(ctx, tel, c, telemetry.ErrCodePortConflict, err.Error())
 			return nil, output.NewSilentError(err)
@@ -380,6 +407,28 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 		filtered = append(filtered, c)
 	}
 	return filtered, nil
+}
+
+func imageTagFrom(image string) string {
+	_, tag, found := strings.Cut(image, ":")
+	if !found || tag == "" {
+		return "latest"
+	}
+	return tag
+}
+
+func emitLocalStackAlreadyRunningWarning(sink output.Sink, port, runningVersion, configTag string) {
+	if configTag == "" {
+		configTag = "latest"
+	}
+	if runningVersion != configTag {
+		output.EmitWarning(sink, fmt.Sprintf(
+			"LocalStack %s is already running on port %s (config specifies %s) — using the running instance",
+			runningVersion, port, configTag,
+		))
+	} else {
+		output.EmitInfo(sink, fmt.Sprintf("LocalStack %s is already running on port %s", runningVersion, port))
+	}
 }
 
 func emitPortInUseError(sink output.Sink, port string) {
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -390,19 +390,7 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 				emitEmulatorStartError(ctx, tel, c, telemetry.ErrCodePortConflict, fmt.Sprintf("running on port %s, configured port %s", found.BoundPort, c.Port))
 				return nil, output.NewSilentError(fmt.Errorf("LocalStack already running on port %s", found.BoundPort))
 			}
-			runningTag := imageTagFrom(found.Image)
-			configTag := c.Tag
-			if configTag == "" {
-				configTag = "latest"
-			}
-			if runningTag != configTag {
-				output.EmitWarning(sink, fmt.Sprintf(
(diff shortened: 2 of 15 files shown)


gtsiolis:
Looking at this now! 👀

internal/container/running.go
```diff
@@ -3,21 +3,55 @@ package container
...
+		return c.Name(), nil
+	}
+
+	image, err := c.Image()
+	if err != nil {
+		return "", err
+	}
+	imageRepo, _, _ := strings.Cut(image, ":")
+
+	containerPort, err := c.ContainerPort()
```
gtsiolis, line 41: **issue**: Not sure if this is a good line to comment on or if the network I'm is more restrictive, but running on a different port when an instance is running on the default port, users get this error.

```sh
# CLIv1 command
localstack start

# CLIv2 config
port = '4567'
```


```sh
✗ failed to start LocalStack: Error response from daemon: failed to set up
  container networking: driver failed programming external connectivity on endpoint
  localstack-aws
  (f52076ed620d767b8dd174c90008fdee66568bed361521e808214e72712e7529): Bind for
  127.0.0.1:443 failed: port is already allocated
```

Image

internal/container/start.go
```diff
@@ -346,7 +346,34 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+		if found != nil {
+			runningTag := imageTagFrom(found.Image)
+			configTag := c.Tag
+			if configTag == "" {
+				configTag = "latest"
+			}
+			if runningTag != configTag {
+				output.EmitWarning(sink, fmt.Sprintf(
+					"Found running LocalStack %s, config specifies %s — using the running instance",
+					runningTag, configTag,
```
gtsiolis: **issue**: When using 2026.03 with CLIv1 and CLIv2 I see the warning showing up. Maybe the tag is using the commit suffix and confuses the CLI?

```sh
# CLIv1 command
localstack start -s localstack:2026.03

# CLIv2 config
tag = '2026.03'
```

```sh
> Warning: LocalStack 2026.3.0:e28f8cfa4 is already running on port 4566 (config specifies 2026.03) — using the running instance
```

internal/runtime/docker.go
```diff
@@ -329,6 +335,45 @@ func (d *DockerRuntime) GetBoundPort(ctx context.Context, containerName string,
 	return bindings[0].HostPort, nil
 }
 
+func (d *DockerRuntime) FindRunningByImage(ctx context.Context, imageRepo string, containerPort string, hostPort string) (*RunningContainer, error) {
+	list, err := d.client.ContainerList(ctx, container.ListOptions{
+		Filters: filters.NewArgs(filters.Arg("status", "running")),
```
gtsiolis, line 381: **issue**: Not sure if this line is best line to commment on, but would expect status to run normally when the version is different. 

```sh
# CLIv1 command
localstack start -s localstack:2026.03

# CLIv2 config
tag = '2026.03'
```

| START | STATUS |
|-|-|
| Image | Image |

internal/container/start.go
```diff
@@ -346,7 +346,34 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+		}
+		if found != nil {
+			runningTag := imageTagFrom(found.Image)
+			configTag := c.Tag
+			if configTag == "" {
+				configTag = "latest"
+			}
+			if runningTag != configTag {
+				output.EmitWarning(sink, fmt.Sprintf(
+					"Found running LocalStack %s, config specifies %s — using the running instance",
```
gtsiolis: **issue**: Saying using the running instance does not make much sense for start and stop commands. It makes more sense for status or similar commands we add later but even then does not add much value I think. Also the wrapping and intendation used on the command output looks off a bit. 💭

Image

internal/container/start.go
```diff
@@ -356,6 +383,28 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+			runningVersion, port, configTag,
+		))
+	} else {
+		output.EmitInfo(sink, fmt.Sprintf("LocalStack %s is already running on port %s", runningVersion, port))
+	}
+}
+
 func emitPortInUseError(sink output.Sink, port string) {
 	actions := []output.ErrorAction{
 		{Label: "Stop existing emulator:", Value: "lstk stop"},
```
gtsiolis: **suggestion**: Not the best line to comment on, but `stop` and `status` should never refuse to operate on a found container because of a version mismatch.

test/integration/start_test.go
```diff
@@ -116,6 +119,60 @@ func TestStartCommandFailsWhenPortInUse(t *testing.T) {
...
+			return
+		}
+		http.NotFound(w, r)
+	}))
+	srv.Listener = ln
+	srv.Start()
+	defer srv.Close()
+
+	stdout, stderr, err := runLstk(t, testContext(t), "", env.With(env.AuthToken, "fake-token"), "start")
+	require.NoError(t, err, "lstk start should succeed when LocalStack is already running: %s", stderr)
```
gtsiolis, line 167: **suggestion**: When the same version is running, we can mention it with a Note message.
```sh
> Note: LocalStack XYZ is already running on port XYZ
```

In this scope, when tag is different from the running instance, we can show:

```
> Warning: LocalStack XYZ is already running on port XYZ (config specifies XYZ2)
```

gtsiolis (changes requested):
Nice one, @carole-lavillonniere! 💯 

Added a first round of comments in some relevant lines of code, let me know if something does not make sense. 🙏

internal/runtime/docker.go
gtsiolis: On this scope, there are some UX changes we can improve in the existing status command output, but can pick this up in a follow up PR.
- Wrapping early
- Tone down labels
- Replace emulator terminology
- [...]

Screenshot 2026-04-16 at 19 10 04

internal/container/running.go
carole-lavillonniere: Fixed!
running-other-port

internal/container/start.go
carole-lavillonniere: Removed the warning now. We "attach" successfully and without warning even if the version does not match config.

internal/runtime/docker.go
carole-lavillonniere: status will now run successfully and without warning even if version configured does not match running instance.

internal/container/start.go
carole-lavillonniere: Good point! Removed the warning completely.

internal/container/start.go
carole-lavillonniere: Totally. They should just work now.

carole-lavillonniere:
> Nice one, @carole-lavillonniere! 💯
> 
> Added a first round of comments in some relevant lines of code, let me know if something does not make sense. 🙏

Thanks for testing extensively @gtsiolis 
I addressed your comments and would appreciate another round of testing.

test/integration/start_test.go
```diff
@@ -116,6 +119,90 @@ func TestStartCommandFailsWhenPortInUse(t *testing.T) {
...
+
+	analyticsSrv, events := mockAnalyticsServer(t)
+	stdout, stderr, err := runLstk(t, ctx, "", env.With(env.AuthToken, "fake-token").With(env.AnalyticsEndpoint, analyticsSrv.URL), "start")
+	require.NoError(t, err, "lstk start should succeed when external container is running: %s", stderr)
+	requireExitCode(t, 0, err)
+	assert.Contains(t, stdout, "already running")
+	assertCommandTelemetry(t, events, "start", 0)
+}
+
+func TestStartCommandAttachesToSameVersion(t *testing.T) {
```
anisaoshafi: name of the test is a bit misleading, since version becomes irrelevant when starting, right?

anisaoshafi (approved):
Was easy to follow the changes; tested and LGTM 🚀 
Good decision to simplify start/stop commands by dismissing the tag/port and informing the user accordingly!

test/integration/start_test.go
carole-lavillonniere: Good point, I renamed it.

internal/container/start.go
```diff
@@ -363,16 +346,76 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+		imageRepo, _, _ := strings.Cut(c.Image, ":")
+		found, err := rt.FindRunningByImage(ctx, []string{imageRepo, "localstack/localstack"}, c.ContainerPort)
+		if err != nil {
+			return nil, fmt.Errorf("failed to scan for running containers: %w", err)
+		}
+		if found != nil {
+			if found.BoundPort != c.Port {
+				output.EmitError(sink, output.ErrorEvent{
+					Title:   fmt.Sprintf("LocalStack is already running on port %s", found.BoundPort),
+					Summary: fmt.Sprintf("Config expects port %s. Only one instance can run at a time.", c.Port),
```
gtsiolis, line 385: **thought**: We previously didn't have this restriction.

>  Only one instance can run at a time.

I'd expect to be able to spin up more than one instance as long as the port is different. This is how it works now, you can set `GATEWAY_LISTEN="0.0.0.0:4567"` and start a second instance side by side. See also [relevant feedback](https://localstack-cloud.slack.com/archives/C07JLGZF7SN/p1756211855328269).

internal/container/start.go
```diff
@@ -363,16 +346,76 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+		// Check extra ports required by this emulator (443 for HTTPS, 4510-4559 for
+		// the service port range). These are singletons: if any is taken, another
+		// LocalStack instance is likely running and we cannot start a new one.
+		extraSpecs := make([]string, len(c.ExtraPorts))
+		for i, ep := range c.ExtraPorts {
+			extraSpecs[i] = ep.HostPort
+		}
+		if conflictPort, err := ports.CheckAvailable(extraSpecs...); err != nil {
+			output.EmitError(sink, output.ErrorEvent{
+				Title:   fmt.Sprintf("Port %s is already in use", conflictPort),
```
gtsiolis, line 416: **praise**: This works as expected, even if it's another service running! 💯

internal/container/start.go
```diff
@@ -363,16 +346,76 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+				Summary: "LocalStack requires this port. Another instance may be running.",
```
gtsiolis: **thought**: This might not be true when another service is using the port you configured. Could show this line only when port 4566 is configured?

```sh
# When configured to use port 4566
✗ Port 4566 already in use
> LocalStack may already be running.

# When configured to use port any other port other than 4566
✗ Port 4568 already in use
```

internal/container/start.go
```diff
@@ -363,16 +346,76 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+
+		imageRepo, _, _ := strings.Cut(c.Image, ":")
+		found, err := rt.FindRunningByImage(ctx, []string{imageRepo, "localstack/localstack"}, c.ContainerPort)
+		if err != nil {
+			return nil, fmt.Errorf("failed to scan for running containers: %w", err)
+		}
+		if found != nil {
+			if found.BoundPort != c.Port {
+				output.EmitError(sink, output.ErrorEvent{
+					Title:   fmt.Sprintf("LocalStack is already running on port %s", found.BoundPort),
```
gtsiolis, line 384: **issue**: If you are running two instances, on port `4566` and port `4567`, this message now is partially true as it captures only one instance. Since we're doing plain port check for now, I'd highlight the port conflict and only assume LS is running when the config is using `4566`.

internal/container/start.go
```diff
@@ -363,16 +346,76 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+					Summary: fmt.Sprintf("Config expects port %s. Only one instance can run at a time.", c.Port),
+					Actions: []output.ErrorAction{
+						{Label: "Stop existing emulator:", Value: "lstk stop"},
+					},
+				})
+				emitEmulatorStartError(ctx, tel, c, telemetry.ErrCodePortConflict, fmt.Sprintf("running on port %s, configured port %s", found.BoundPort, c.Port))
+				return nil, output.NewSilentError(fmt.Errorf("LocalStack already running on port %s", found.BoundPort))
+			}
+			output.EmitInfo(sink, "LocalStack is already running")
```
gtsiolis, line 393: **issue**: If you are running another service on 4566 or the same port, this message is quite misleading. ⚠️ 

| Another service on port 4566 | Trying to start LS on port 4566 |
|-|-|
| Image | Screenshot 2026-04-23 at 11 02 49 |

gtsiolis (commented):
Left some more comments, @carole-lavillonniere! What do you think?

internal/container/start.go
carole-lavillonniere: Reporting here e the conversation @gtsiolis and I had privately: Because of the fact that the aws emulator also uses other ports (`443` and `4510–4559`), we are not able to run multiple aws emulators simultaneously at the moment (ports would conflict).

internal/container/start.go
carole-lavillonniere: Since we decided to not support multiple containers of same provider, `FindRunningByImage` reliably finds at most one container, and the current message is correct.

internal/container/start.go
carole-lavillonniere: I ended up just removing "Another instance may be running." 17126845de359f4ce4be04e6b56a0a818679d58a
It's actually not likely because when we reach this error we have already run:
1. `FindRunningByImage` -> no LS container running in Docker
2. `CheckAvailable(c.Port)` -> main port is free
So one of the extra ports (443, 4510-4559) is being busy with another process (ie nginx on 443?) and "Another instance may be running" is not right.

internal/container/start.go
carole-lavillonniere: Good point. We only reach this point after:
  1. `FindRunningByImage` → nil (no LS container)
  2. `fetchLocalStackInfo` → fails (not responding as LS)
So we have strong evidence it's NOT LocalStack occupying the port.
Pushed this change 55d01a1ba4d16440d0fdaa3bb9ed5fc55fa3ec64

gtsiolis: Looks good.

gtsiolis (approved):
Looks shippable! :shipit:

internal/container/start.go
gtsiolis: Thanks for _not_ adding "Another instance may be running". 😁

PR #193, opened 2026-04-22
Add restart command

gtsiolis:
Adds `restart` command that stops the running emulator and starts it again in a single command.

The command runs the full stop and start lifecycle: stops the container, re-validates the auth token and license, pulls the image if needed, and waits for the health endpoint before returning. This ensures restart is always safe (no stale state, no skipped validation) rather than just bouncing the container.

Screenshot 2026-04-22 at 16 52 24

### DIFF
--- a/cmd/restart.go
+++ b/cmd/restart.go
@@ -1,0 +1,48 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/localstack/lstk/internal/config"
+	"github.com/localstack/lstk/internal/container"
+	"github.com/localstack/lstk/internal/env"
+	"github.com/localstack/lstk/internal/log"
+	"github.com/localstack/lstk/internal/output"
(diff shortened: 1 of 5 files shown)


anisaoshafi (approved):
LGTM, tested and looks good 🚀


Was thinking if the customers would expect `restart` to re-pull the image. If for some reasons we introduce some API changes (reflecting AWS changes or sth else), or different behaviours that previous to restart worked, can be confusing for the customer, plus the extra wait for docker pull. 
I am in favour of upgrading though, just a random thought.

carole-lavillonniere (approved):
Looks good 👏 
Pushed a small commit to make sure we emit the telemetry events for the new command  76c77dcd1c7638b96407e225cf7560391e5c86ac

gtsiolis:
Thanks for adding commits and merging, @carole-lavillonniere! 🏀

PR #195, opened 2026-04-23
Test PERSISTENCE env vars

anisaoshafi:
Persistence and related snapshot configuration vars may be passed in host env prefixed with LOCALSTACK* or through config.toml env vars.

This PR adds a couple of related tests.

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -536,13 +536,18 @@ func awaitStartup(ctx context.Context, rt runtime.Runtime, sink output.Sink, con
 }
 
 // filterHostEnv returns the subset of host environment entries that should be
-// forwarded to the emulator container. It keeps CI and LOCALSTACK_* variables
+// forwarded to the emulator container. It keeps CI, LOCALSTACK_*, and persistence-related variables
 // but explicitly drops LOCALSTACK_AUTH_TOKEN so the host value cannot overwrite
 // the token resolved by lstk (which may come from the keyring).
 func filterHostEnv(envList []string) []string {
 	var out []string
 	for _, e := range envList {
-		if strings.HasPrefix(e, "CI=") || (strings.HasPrefix(e, "LOCALSTACK_") && !strings.HasPrefix(e, "LOCALSTACK_AUTH_TOKEN=")) {
+		if strings.HasPrefix(e, "CI=") ||
+			(strings.HasPrefix(e, "LOCALSTACK_") && !strings.HasPrefix(e, "LOCALSTACK_AUTH_TOKEN=")) ||
+			strings.HasPrefix(e, "PERSISTENCE=") ||
+			strings.HasPrefix(e, "SNAPSHOT_SAVE_STRATEGY=") ||
+			strings.HasPrefix(e, "SNAPSHOT_LOAD_STRATEGY=") ||
+			strings.HasPrefix(e, "SNAPSHOT_FLUSH_INTERVAL=") {
 			out = append(out, e)
 		}
 	}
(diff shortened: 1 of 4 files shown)


test/integration/start_test.go
```diff
@@ -256,6 +256,67 @@ func TestStartCommandPassesCIAndLocalStackEnvVars(t *testing.T) {
...
+	requireExitCode(t, 0, err)
+
+	inspect, err := dockerClient.ContainerInspect(ctx, containerName)
+	require.NoError(t, err, "failed to inspect container")
+	require.True(t, inspect.State.Running)
+
+	envVars := containerEnvToMap(inspect.Config.Env)
+	assert.Equal(t, "1", envVars["PERSISTENCE"])
+}
+
```
carole-lavillonniere, line 406: praise: good call adding a test to make sure we can enable persistence from config

internal/container/start.go
```diff
@@ -481,13 +481,18 @@ func awaitStartup(ctx context.Context, rt runtime.Runtime, sink output.Sink, con
...
 func filterHostEnv(envList []string) []string {
 	var out []string
 	for _, e := range envList {
-		if strings.HasPrefix(e, "CI=") || (strings.HasPrefix(e, "LOCALSTACK_") && !strings.HasPrefix(e, "LOCALSTACK_AUTH_TOKEN=")) {
+		if strings.HasPrefix(e, "CI=") ||
+			(strings.HasPrefix(e, "LOCALSTACK_") && !strings.HasPrefix(e, "LOCALSTACK_AUTH_TOKEN=")) ||
+			strings.HasPrefix(e, "PERSISTENCE=") ||
+			strings.HasPrefix(e, "SNAPSHOT_SAVE_STRATEGY=") ||
+			strings.HasPrefix(e, "SNAPSHOT_LOAD_STRATEGY=") ||
+			strings.HasPrefix(e, "SNAPSHOT_FLUSH_INTERVAL=") {
```
carole-lavillonniere: issue: I'm not convinced we should explicitly forward `PERSISTENCE` (and the other vars here).
1. We should already forward all vars prefixed with `LOCALSTACK_` so `LOCALSTACK_PERSISTENCE=1` should work without code changes (maybe worth adding an integration test)
2. We can also set `PERSISTENCE=1` in `config.toml` (and you've added a test for it)

Isn't this enough? My concern is that it does not really scale to whitelist env vars and we also don't want to blindly forward all vars in the env because it can contain secrets or other unrelated vars.

anisaoshafi: Good point, I will take a look and adjust this 🏃🏼

anisaoshafi: The PR is now just a couple of tests since these configurations were already supported - thanks for poining that out. Feel free to take another look.

internal/container/start_test.go
```diff
@@ -71,20 +71,23 @@ func TestFilterHostEnv(t *testing.T) {
 		"LOCALSTACK_DISABLE_EVENTS=1",
 		"LOCALSTACK_API_ENDPOINT=https://example.test",
 		"LOCALSTACK_AUTH_TOKEN=host-token",
+		"LOCALSTACK_PERSISTENCE=1",
 		"PATH=/usr/bin",
 		"HOME=/home/user",
 		"CI_PIPELINE=foo",
+		"PERSISTENCE=1",
+		"SNAPSHOT_SAVE_STRATEGY=ON_SHUTDOWN",
```
carole-lavillonniere: I suppose these are leftovers?

carole-lavillonniere (approved):

PR #196, opened 2026-04-23
Inline sink emit calls

carole-lavillonniere:
## Summary

Refactoring addressing suggested improvement #1 from @thrau in this comment https://github.com/localstack/lstk/pull/127#pullrequestreview-3972108878

The goal is to make reading the code easier, be more explicit and idiomatic go, and remove some lines of code.

## Changes

- Replace package-level `EmitXxx(sink, ...)` helpers with direct `sink.Emit(output.XxxEvent{...})` calls at all sites
- Change `Event` from a generic union constraint to a sealed marker interface (`sealedEvent()` unexported method), enabling an exported `Sink.Emit(event Event)` without generics
- Remove ~88 lines of boilerplate emit helpers; retain `SpinnerStart`/`SpinnerStop`/`SpinnerStartWithDuration` as value constructors (they encode the `MinDuration` default)


carole-lavillonniere:
@claude review once

thrau (approved):
This is a really nice refactor to move us towards a more object-oriented codebase! LGTM

internal/output/events.go
```diff
@@ -76,18 +76,29 @@ type ResourceSummaryEvent struct {
...
+func (SpinnerEvent) sealedEvent()          {}
+func (ErrorEvent) sealedEvent()            {}
+func (AuthEvent) sealedEvent()             {}
+func (InstanceInfoEvent) sealedEvent()     {}
+func (TableEvent) sealedEvent()            {}
+func (ResourceSummaryEvent) sealedEvent()  {}
+func (ContainerStatusEvent) sealedEvent()  {}
+func (ProgressEvent) sealedEvent()         {}
+func (UserInputRequestEvent) sealedEvent() {}
+func (LogLineEvent) sealedEvent()          {}
```
silv-io, line 93: As mentioned in-sync, I think we can remove that seal in the future and just define the `Event` interface more clearly and then move the Events to their domains.
But that's a future optimization.

silv-io (approved):
LGTM!

PR #197, opened 2026-04-23
Inline telemetry lifecycle event calls

carole-lavillonniere:
## Summary

Refactoring addressing suggested improvement #1 from @thrau in this comment https://github.com/localstack/lstk/pull/127#pullrequestreview-3972108878

The goal is to make reading the code easier, be more explicit and idiomatic go.

## Changes

Move nil safety for `*telemetry.Client` from call-site guards into nil receiver methods, and inline the two private helpers `emitEmulatorStartError` / `emitEmulatorStartSuccess`.

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -393,7 +393,13 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 						{Label: "Stop existing emulator:", Value: "lstk stop"},
 					},
 				})
-				emitEmulatorStartError(ctx, tel, c, telemetry.ErrCodePortConflict, fmt.Sprintf("running on port %s, configured port %s", found.BoundPort, c.Port))
+				tel.EmitEmulatorLifecycleEvent(ctx, telemetry.LifecycleEvent{
+					EventType: telemetry.LifecycleStartError,
+					Emulator:  c.EmulatorType,
+					Image:     c.Image,
+					ErrorCode: telemetry.ErrCodePortConflict,
+					ErrorMsg:  fmt.Sprintf("running on port %s, configured port %s", found.BoundPort, c.Port),
+				})
 				return nil, output.NewSilentError(fmt.Errorf("LocalStack already running on port %s", found.BoundPort))
 			}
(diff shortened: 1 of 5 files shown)


anisaoshafi (approved):
Valid changes IMO, looks good 💅🏼

internal/container/start.go
thrau: example where `tel` can be removed.

internal/telemetry/client.go
```diff
@@ -36,6 +36,9 @@ type Client struct {
 // SetAuthToken stores the resolved auth token for inclusion in telemetry events.
 // Call this once the token is known (e.g. after keyring resolution or interactive login).
 func (c *Client) SetAuthToken(token string) {
+	if c == nil {
+		return
+	}
```
thrau: q: under what circumstances could `c` be nil?

thrau (changes requested):
Thanks @carole-lavillonniere! Looks like a simple refactoring that makes everything a bit tidier! The code now constructs the event structs in-place at the call site, which is much more readable than the previous multi-argument helper functions :+1: 


Given the goal is to reduce functional over-engineering, one change we can still make in `start.go` to advance that goal is to remove `tel` as an argument in all functions that are passed `opts StartOptions` as arguments, given that the telemetry client is a value of `StartOptions`.

In the future (outside the scope of this PR), we should further encapsulate `Start`, `Stop` and similar container commands into a common struct (like a `ContainerService` class), and make stable dependencies (like the the ContainerRuntime, TelemetryClient, etc...) part of the struct.

internal/telemetry/client.go
carole-lavillonniere: Removed the guards. The idiom was not applied consistently here anyways.

internal/container/start.go
carole-lavillonniere: Good point, I forgot those :see_no_evil: 
Removed those unnecessary tel args.
e29bbe7f7791cdb9145365a0a94cdb2192da1eab

thrau (approved):
thanks for addressing my comments! LGTM 🚀

PR #198, opened 2026-04-23
Refactor command telemetry

carole-lavillonniere:
Refactoring addressing suggested improvement #2 from @thrau in this comment https://github.com/localstack/lstk/pull/127#pullrequestreview-3972108878

## Changes

- Replaces the `commandWithTelemetry` decorator and per-command manual telemetry with a single `instrumentCommands` function in `root.go` that walks the Cobra command tree and wraps every `RunE` automatically
- Command names are derived from `cmd.CommandPath()` (stripping the `lstk ` prefix) — no more hardcoded strings that can drift
- Removes `*telemetry.Client` from constructors that only needed it for the decorator (logout, status, logs, setup, config, volume, update, aws); kept where still needed for lifecycle events (start, stop, restart) or token updates (login)

## Deviation from the suggestion

> You can use Cobra's PersistentPreRun func to store data globally that you need (for example the startTime)

The PR instead captures `startTime` directly inside the `RunE` closure in `instrumentCommands`. This is simpler because local to the closure and there is no shared state needed.

> Instead, we could centralize this into root.go and remove the tel dependency out of the constructors completely.

`tel` must still be passed to constructors for commands that do more things with it: emitting lifecycle events, updating auth token. 

## Change in telemetry data

Before this change `lstk` (without subcommand) and `lstk start` both emitted `command="start"` and could not be differentiated in telemetry. Now `lstk` emits `"lstk"` instead.

### DIFF
--- a/.claude/skills/add-command/SKILL.md
+++ b/.claude/skills/add-command/SKILL.md
@@ -61,7 +61,7 @@ Create `internal/<package>/<name>.go` (use an existing package if it fits, or cr
 
 In `cmd/root.go`, add the new command to `root.AddCommand(...)`.
 
-If the command constructor needs dependencies (like `*env.Env` or `*telemetry.Client`), add them as parameters matching the existing pattern.
+If the command constructor needs dependencies (like `*env.Env`), add them as parameters matching the existing pattern.
 
 ## Step 4: Add new event types (if needed)
 
@@ -96,9 +96,7 @@ Create `test/integration/<name>_test.go` with:
 
 ## Telemetry
 
-Every new command must emit an `lstk_command` telemetry event. Wrap the command's `RunE` with `commandWithTelemetry(name, tel, fn)` — this handles timing, exit code, and error message automatically.
-
-Start and stop are exceptions: they emit `lstk_lifecycle` events in addition to `lstk_command`, so they manage their own telemetry manually instead of using `commandWithTelemetry`.
+`lstk_command` telemetry is emitted automatically for every command via `instrumentCommands` in `cmd/root.go` — no per-command wiring needed. Do NOT pass `*telemetry.Client` to a new command constructor unless the command emits additional lifecycle events (like `lstk_lifecycle` for start/stop/restart) or needs to update the auth token (like login).
 
 In the corresponding integration test, add an assertion that the `lstk_command` event was emitted.
 
--- a/cmd/aws.go
+++ b/cmd/aws.go
@@ -12,12 +12,11 @@ import (
 	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/output"
 	"github.com/localstack/lstk/internal/runtime"
-	"github.com/localstack/lstk/internal/telemetry"
 	"github.com/localstack/lstk/internal/terminal"
 	"github.com/spf13/cobra"
 )
 
-func newAWSCmd(cfg *env.Env, tel *telemetry.Client) *cobra.Command {
+func newAWSCmd(cfg *env.Env) *cobra.Command {
 	return &cobra.Command{
 		Use:   "aws [args...]",
 		Short: "Run AWS CLI commands against LocalStack",
@@ -35,7 +34,7 @@ Examples:
   lstk aws s3 mb s3://my-bucket`,
 		DisableFlagParsing: true,
 		PreRunE:            initConfig,
-		RunE: commandWithTelemetry("aws", tel, func(cmd *cobra.Command, args []string) error {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			rt, err := runtime.NewDockerRuntime(cfg.DockerHost)
 			if err != nil {
 				return err
@@ -93,6 +92,6 @@ Examples:
 			}
 
 			return awscli.Exec(cmd.Context(), "http://"+host, profileExists, stdout, stderr, args)
-		}),
+		},
 	}
 }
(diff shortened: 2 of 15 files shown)


anisaoshafi (approved):
LGTM, much simpler 🚀 

Question: I wasn't aware we want to track "lstk" differently from "lstk start". Why do we want to check them separately? I assume we need to adjust the dashboard queries so that start events count "lstk" in addition to "lstk start".

In `AGENTS.md` this part needs to be removed: "Wrap the command's `RunE` with `commandWithTelemetry(name, tel, fn)` — this handles timing, exit code, and error message automatically".

carole-lavillonniere:
> Question: I wasn't aware we want to track "lstk" differently from "lstk start". Why do we want to check them separately? I assume we need to adjust the dashboard queries so that start events count "lstk" in addition to "lstk start".

Good point. The different command name comes for free with the refactoring, and I assumed it would be better to have this added visibility over `lstk` versus `lstk start`. However if it's extra work with not much added value, I can bring back the previous command name. @thrau @mmaureenliu?


> In `AGENTS.md` this part needs to be removed: "Wrap the command's `RunE` with `commandWithTelemetry(name, tel, fn)` — this handles timing, exit code, and error message automatically".

I cannot find this string in AGENTS.md, is it Claude or myself hallucinating :upside_down_face:? 

@anisaoshafi

anisaoshafi:
> > In AGENTS.md this part needs to be removed: "Wrap the command's RunE with commandWithTelemetry(name, tel, fn) — this handles timing, exit code, and error message automatically".
> 
> I cannot find this string in AGENTS.md, is it Claude or myself hallucinating 🙃?

@carole-lavillonniere sorry, my bad. I meant in [add-commands/SKILL.md](https://github.com/localstack/lstk/blob/main/.claude/skills/add-command/SKILL.md#telemetry)

.claude/skills/add-command/SKILL.md
```diff
@@ -61,7 +61,7 @@ Create `internal/<package>/<name>.go` (use an existing package if it fits, or cr
 
 In `cmd/root.go`, add the new command to `root.AddCommand(...)`.
 
-If the command constructor needs dependencies (like `*env.Env` or `*telemetry.Client`), add them as parameters matching the existing pattern.
+If the command constructor needs dependencies (like `*env.Env`), add them as parameters matching the existing pattern.
```
thrau, line 64: praise: updating skills is sometimes forgotten! great that you keep it up to date with every change 🙌

thrau (approved):
Thank you for following up on this @carole-lavillonniere! To me this is much cleaner now and objectively better!

* Telemetry is now treated as a cross-cutting concern. By centralizing instrumentation in root.go, the individual commands (like aws, logs, or status) don't need to know that they are being timed or tracked.
* Adding a new command is now simpler. You don't have to worry about telemetry wiring, the framework handles it automatically.
* The use of recursive command wrapping ensures 100% coverage. Every command in the tree is now automatically instrumented with basic telemetry (execution time, exit code, flags), which makes analytics much more reliable.

I would consider adding a way to disable telemetry for individual commands that we may not want to track, once that need comes up.

mmaureenliu:
> Good point. The different command name comes for free with the refactoring, and I assumed it would be better to have this added visibility over `lstk` versus `lstk start`. However if it's extra work with not much added value, I can bring back the previous command name. @thrau @mmaureenliu?

There are two separate things we might want to understand from the `lstk` vs `lstk start` commands telemetry:
1. Intended operation, such as `start`, `stop` - here we don't care about the differentiation but need to make sure the logic (when counting `start` we also need to count `lstk`) is captured somewhere and applied correctly to stats without forcing the end data consumer to actively understand this nuance
2. User habits, i.e. do they prefer to use simply `lstk` or do they insist on `lstk start`. We may want to use this data to revise our assumption and to feed into future decision making on other tools. In this case we do need to differentiate between `lstk` and `lstk start`

In either case, we need to be careful with `lstk -h`, `lstk -v`, etc., which != `lstk start` (I guess you've taken care of this already?)

In terms of implementation, #1 is most important for me to track, so I don't mind whether differentiation is done but want to be sure counting is correct. #2 is nice to have.

carole-lavillonniere:
> > Good point. The different command name comes for free with the refactoring, and I assumed it would be better to have this added visibility over `lstk` versus `lstk start`. However if it's extra work with not much added value, I can bring back the previous command name. @thrau @mmaureenliu?
> 
> There are two separate things we might want to understand from the `lstk` vs `lstk start` commands telemetry:
> 
>     1. Intended operation, such as `start`, `stop` - here we don't care about the differentiation but need to make sure the logic (when counting `start` we also need to count `lstk`) is captured somewhere and applied correctly to stats without forcing the end data consumer to actively understand this nuance
> 
>     2. User habits, i.e. do they prefer to use simply `lstk` or do they insist on `lstk start`. We may want to use this data to revise our assumption and to feed into future decision making on other tools. In this case we do need to differentiate between `lstk` and `lstk start`
> 
> 
> In either case, we need to be careful with `lstk -h`, `lstk -v`, etc., which != `lstk start` (I guess you've taken care of this already?)
> 
> In terms of implementation, #1 is most important for me to track, so I don't mind whether differentiation is done but want to be sure counting is correct. #2 is nice to have.

Thanks @mmaureenliu, and @anisaoshafi for bringing attention to the matter.
I pushed this commit 34f0724561e3b871909b80492eac1aea2cb5d08e that will bring back the telemetry for `lstk`. It should now be unchanged and both `lstk` and `lstk start` emit `command="start"`

anisaoshafi (approved):
:shipit:

PR #205, opened 2026-04-27
Use ANSI red for error styles to match terminal profile

gtsiolis:
Switch `LogError` and `ErrorTitle` styles from hex `#C33820` to ANSI red (`"1"`) so each terminal renders error text using its own profile's red.

| BEFORE | AFTER |
|--------|--------|
| Frame 852839925 | Frame 852839926 |
| Frame 852839928 | Frame 852839927 |
| Frame 852839930 | Frame 852839929 |

### DIFF
--- a/internal/ui/styles/styles.go
+++ b/internal/ui/styles/styles.go
@@ -51,7 +51,7 @@ var (
 		Foreground(lipgloss.Color("214"))
 
 	LogError = lipgloss.NewStyle().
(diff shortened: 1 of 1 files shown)


carole-lavillonniere (approved):

gtsiolis:
Thanks for taking a look and merging, @carole-lavillonniere! 😁

PR #206, opened 2026-04-27
Fix Docker Desktop detection via ~/.docker/run/docker.sock

carole-lavillonniere:
## Motivation

Docker Desktop on macOS uses `~/.docker/run/docker.sock` when the `/var/run/docker.sock` symlink is absent. lstk only probed the standard path, causing false "Docker is not available" errors even with Docker Desktop running.

## Changes

- Add `~/.docker/run/docker.sock` to the socket probe list in `findDockerSocket()`
- Extract `vmSocketPaths()` as single source of truth used by both `findDockerSocket()` and `isVM()` — previously two separate lists that could drift
- Add Lima host socket (`~/.lima/docker/sock/docker.sock`) to `isVM()`, which was missing: containers bind-mount `/var/run/docker.sock` regardless of the host-side socket path
- Convert repetitive VM detection subtests to table-driven tests; adding a new runtime now requires only one entry

## Tests

Unit tests cover all VM socket paths for both discovery (`findDockerSocket`) and bind-mount path remapping (`isVM`/`SocketPath`).

Closes DRG-754

### DIFF
--- a/internal/runtime/docker.go
+++ b/internal/runtime/docker.go
@@ -57,20 +57,24 @@ func NewDockerRuntime(dockerHost string) (*DockerRuntime, error) {
 	return &DockerRuntime{client: cli}, nil
 }
 
+func vmSocketPaths(home string) []string {
+	return []string{
+		filepath.Join(home, ".docker", "run", "docker.sock"),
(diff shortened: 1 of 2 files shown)


anisaoshafi (approved):
Thank you 👏🏼

PR #208, opened 2026-04-28
Hide banner on first-run before auth

gtsiolis:
On first interactive run with no auth token, run the device-flow login before launching the start UI so the header doesn't render with (No license) while authentication is still pending. This will hide the banner during the login flow and render the header after login succeeds and the start UI takes over with a resolved license label.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-04-28 at 11 57 10 | imageimage |

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -159,23 +159,6 @@ func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *t
 		return fmt.Errorf("failed to get config: %w", err)
 	}
 
-	interactive := isInteractiveMode(cfg)
-
-	// On first run with no token, run the auth flow before showing the start UI
-	// so the header isn't rendered with "(No license)" before authentication completes.
-	if interactive && cfg.AuthToken == "" {
-		platformClient := api.NewPlatformClient(cfg.APIEndpoint, logger)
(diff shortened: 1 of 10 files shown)


carole-lavillonniere:
I would like to rework this because here we spin up two sequential TUI: one for login, then exit, and new TUI for start. That's not following the existing pattern (one TUI per command) and is a bit surprising.

gtsiolis (commented):
I'd expect the header to be injected in a linear way after the authentication is complete. Now the header is injected above the welcome message when the authentication completes which reminds me of layout shift in browsers.

| Before authentication | After authentication |
|--------|--------|
| Screenshot 2026-04-29 at 12 41 14 | Screenshot 2026-04-29 at 12 41 24 |

carole-lavillonniere (approved):

carole-lavillonniere:
> I'd expect the header to be injected in a linear way after the authentication is complete. Now the header is injected above the welcome message when the authentication completes which reminds me of layout shift in browsers.
> Before authentication 	After authentication
> Screenshot 2026-04-29 at 12 41 14 	Screenshot 2026-04-29 at 12 41 24

Updated to "clear" the output after auth and show auth success+header only upon auth success.

PR #209, opened 2026-04-28
Integration tests parallelism - part 1

carole-lavillonniere:
## Summary

First pass at parallelizing the integration test suite, only addresses the low-hanging fruits. Docker-dependent tests are left serial for a follow-up since they share container name, ports and the host's Docker daemon.
Execution time goes from about 12min to 7min.

towards DRG-665

### Changes

- `t.Parallel()` added to tests that are safe to run concurrently
- Isolate some tests from the developer's real config, changed values of `HOME` / `XDG_CONFIG_HOME`
- Drop redundant `ContainerStop` calls before `ContainerRemove(Force: true)` in cleanup because Force-remove already SIGKILLs; an explicit stop first would add the default 10s SIGTERM grace period for nothing.

### DIFF
--- a/test/integration/logging_test.go
+++ b/test/integration/logging_test.go
@@ -3,18 +3,35 @@ package integration_test
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 
+	"github.com/localstack/lstk/test/integration/env"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestLogging_NonTTY_WritesToLogFile(t *testing.T) {
-	logPath := filepath.Join(configDir(), "lstk.log")
-	_ = os.Remove(logPath)
+	t.Parallel()
+
+	tmpHome := t.TempDir()
+	var logPath string
+	var e env.Environ
+	if runtime.GOOS == "windows" {
+		appData := filepath.Join(tmpHome, "AppData", "Roaming")
+		e = env.Without("HOME", "XDG_CONFIG_HOME", "APPDATA", "USERPROFILE", "HOMEDRIVE", "HOMEPATH").
+			With(env.Home, tmpHome).
+			With("USERPROFILE", tmpHome).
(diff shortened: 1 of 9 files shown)


anisaoshafi (approved):
great improvement already 12 -> 7 min 🏃🏼‍♀️

test/integration/logs_test.go
```diff
@@ -13,6 +15,21 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// writeAwsConfig writes a minimal aws-only config so tests don't inherit the
+// developer's real ~/.config/lstk/config.toml (which may target a different
+// emulator / running container).
+func writeAwsConfig(t *testing.T) string {
```
anisaoshafi, line 21: thanks for fixing this 🙏🏼

test/integration/logging_test.go
```diff
@@ -3,18 +3,35 @@ package integration_test
...
+	var e env.Environ
+	if runtime.GOOS == "windows" {
+		appData := filepath.Join(tmpHome, "AppData", "Roaming")
+		e = env.Without("HOME", "XDG_CONFIG_HOME", "APPDATA", "USERPROFILE", "HOMEDRIVE", "HOMEPATH").
+			With(env.Home, tmpHome).
+			With("USERPROFILE", tmpHome).
+			With("APPDATA", appData)
+		logPath = filepath.Join(appData, "lstk", "lstk.log")
+	} else {
+		require.NoError(t, os.MkdirAll(filepath.Join(tmpHome, ".config"), 0755))
```
anisaoshafi, line 28: sorry, a bit late, but noticed we have [this](https://github.com/localstack/lstk/blob/main/test/integration/config_test.go#L241) func `testEnvWithHome` that could probably be reused since it already handles the env for different OS.

carole-lavillonniere: Good point! Addressed here https://github.com/localstack/lstk/pull/222

PR #210, opened 2026-04-28
Add snapshot save command (local)

anisaoshafi:
Added a new snapshot subcommand with an initial save action that saves the current LocalStack state to a snapshot file.
Scope for this PR is saving a snapshot **_locally_**.
Saving **_remotely_** coming next.


cmd/snapshot.go
```diff
@@ -0,0 +1,76 @@
...
+
+			appConfig, err := config.Get()
+			if err != nil {
+				return fmt.Errorf("failed to get config: %w", err)
+			}
+			if len(appConfig.Containers) == 0 {
+				return fmt.Errorf("no emulator configured")
+			}
+
+			c := appConfig.Containers[0]
```
carole-lavillonniere: issue: this could potentially panic. Snapshots are only for aws AFAIK so we could do this instead:
` awsContainer := config.ContainerConfig{Type: config.EmulatorAWS, Port: config.DefaultAWSPort}`

internal/snapshot/destination.go
```diff
@@ -0,0 +1,52 @@
...
+	String() string
+}
+
+// ParseDestination returns a Destination for the user-supplied path, or an error for cloud/bare names.
+func ParseDestination(dest string) (Destination, error) {
+	if dest == "" {
+		dest = "ls-state-export"
+	} else if strings.Contains(dest, "://") {
+		return nil, fmt.Errorf("cloud destinations are not yet supported — use a file path like ./my-snapshot")
+	} else if !strings.HasPrefix(dest, ".") && !strings.HasPrefix(dest, "/") && !strings.HasPrefix(dest, "~") && !strings.Contains(dest, "/") {
```
carole-lavillonniere: issue: this looks like it would not work on Windows? Can we make unit-tests more OS-agnostic to verify? Instead of defining the paths as strings we could use `filepath.Join("subdir", "state")`

internal/snapshot/client.go
```diff
@@ -0,0 +1,44 @@
...
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("connect to LocalStack: %w", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		_ = resp.Body.Close()
+		return nil, fmt.Errorf("LocalStack returned status %d", resp.StatusCode)
+	}
+	return resp.Body, nil
+}
```
carole-lavillonniere: question: we already have `internal/snapshot/client.go` for the aws emulator http api, can we add the new method there instead? Keeping a dedicated interface `StateExporter` just for the endpoints related to snapshots make sense to me though

internal/snapshot/client.go
```diff
@@ -0,0 +1,44 @@
...
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("connect to LocalStack: %w", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		_ = resp.Body.Close()
+		return nil, fmt.Errorf("LocalStack returned status %d", resp.StatusCode)
+	}
+	return resp.Body, nil
```
carole-lavillonniere: question: instead of returning` io.ReadCloser` can we instead pass the `io.Writer` destination as argument?
This way we're sure the caller won't forget to close the returned body.

internal/snapshot/destination.go
```diff
@@ -0,0 +1,52 @@
...
+	return LocalDestination{Path: abs}, nil
+}
+
+// LocalDestination writes snapshot state to a local file.
+type LocalDestination struct {
+	Path string
+}
+
+func (d LocalDestination) Writer() (io.WriteCloser, error) {
+	return os.Create(d.Path)
```
carole-lavillonniere: issue: this abstraction won't work once we introduce pods because in the case of pods we are not saving to a file so we don't have a `io.WriteCloser`. We could just drop the interface for now?

gtsiolis:
I've posted a comment[[1](https://linear.app/localstack/issue/PRO-212/add-snapshot-command#comment-7a343a78)] in PRO-212 to discuss some potential changes, comments welcome!

cmd/snapshot.go
anisaoshafi: attempted to address here: https://github.com/localstack/lstk/pull/210/commits/f5a1f1249ef22d45f3d15ce04ba6b063b2fb52c5

internal/snapshot/client.go
carole-lavillonniere: Sorry I meant to say `internal/emulator/aws/client.go`

internal/snapshot/save_test.go
```diff
@@ -0,0 +1,200 @@
...
+	body []byte
+	err  error
+}
+
+func (f *fakeExporter) ExportState(_ context.Context, _ string) (io.ReadCloser, error) {
+	if f.err != nil {
+		return nil, f.err
+	}
+	return io.NopCloser(bytes.NewReader(f.body)), nil
+}
```
carole-lavillonniere: nit: can we use gomock to generate this mock exporter?

internal/snapshot/client.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"context"
+	"io"
+)
+
+// StateExporter retrieves state from the running LocalStack instance.
+type StateExporter interface {
+	ExportState(ctx context.Context, host string) (io.ReadCloser, error)
+}
```
carole-lavillonniere: nit: the go convention is rather to define an interface where it is used, that would be directly in` save.go`

internal/snapshot/client.go
carole-lavillonniere: @anisaoshafi were you able to explore this idea?

internal/snapshot/destination_test.go
```diff
@@ -0,0 +1,195 @@
...
+func TestParseDestinationDefault(t *testing.T) {
+	t.Parallel()
+	wd, err := os.Getwd()
+	require.NoError(t, err)
+
+	now := time.Date(2026, 5, 11, 21, 4, 32, 0, time.UTC)
+	got, err := snapshot.ParseDestination("", now)
+	require.NoError(t, err)
+	assert.Equal(t, filepath.Join(wd, "snapshot-2026-05-11T21-04-32.zip"), got)
+}
```
carole-lavillonniere, line 83: nit: can't this test case go as a new row in `TestParseDestination`?

internal/snapshot/destination_test.go
```diff
@@ -0,0 +1,195 @@
...
+)
+
+func TestParseDestinationRejectsDirectory(t *testing.T) {
+	t.Parallel()
+	dir := t.TempDir()
+	now := time.Date(2026, 5, 11, 21, 4, 32, 0, time.UTC)
+	_, err := snapshot.ParseDestination(dir, now)
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "is a directory")
+}
```
carole-lavillonniere: question: isn't this test case already covered in `TestParseDestination`?

cmd/snapshot.go
```diff
@@ -0,0 +1,85 @@
...
+
+func newSnapshotSaveCmd(cfg *env.Env) *cobra.Command {
+	return &cobra.Command{
+		Use:   "save [destination]",
+		Short: "Save a snapshot of the emulator state",
+		Long: `Save a snapshot of the running emulator's state.
+
+Pass [destination] as an absolute or relative path for the exported file:
+
+  lstk snapshot save                    # saves to ./snapshot-<YYYY-MM-DDTHH-mm-ss>.zip
```
carole-lavillonniere: question: spec says:
> Omitting REF generates snapshot-<timestamp>-<short-hash> and writes to CWD.

So it looks like we're missing a hash?
Not sure if it makes sense since it is unlikely to conflict with the timestamp being precise to the second.

anisaoshafi: @gtsiolis we agreed to remove the hash, but forgot to remove one reference in the spec. Is that correct?

internal/snapshot/save_test.go
anisaoshafi: Done here: https://github.com/localstack/lstk/pull/210/commits/b16938c79f3fa7de5b94667c6fd6c9415bee8834

internal/snapshot/client.go
anisaoshafi: TIL, done here: https://github.com/localstack/lstk/pull/210/commits/8bdf86e9baa4e87ea33315ac8f45045798af9666

internal/snapshot/destination_test.go
anisaoshafi: Good catch, addressed here: https://github.com/localstack/lstk/pull/210/commits/832bbf8258f6c1d488de5bd2989a0030e483ee54

internal/snapshot/destination_test.go
anisaoshafi: addressed here: https://github.com/localstack/lstk/pull/210/commits/832bbf8258f6c1d488de5bd2989a0030e483ee54

internal/snapshot/client.go
anisaoshafi: Sorry, I missed this. I gave it a try here 🤞🏼 https://github.com/localstack/lstk/pull/210/changes/2bf9e135118b9ab3481134511f2c17617e2ba1ec

anisaoshafi:
@gtsiolis would really appreciate your UX input when you have time to test this 🙏🏼

cmd/snapshot.go
gtsiolis: Yes, I think we can safely skip it, the timestamp should be quite unique on every generation, right? Plus, system will place a second file next to each other in the rare case you try to generate one in the same second, correct?

> snapshot-2026-05-12T20-07-31.zip
> snapshot-2026-05-12T20-07-31 (1).zip

gtsiolis: I see we don't handle the double saving on the same second and overwrite the existing one, but sounds like an edge case and ok to leave as is and skip the over-engineering. Could we continue skipping the hash and easily append the (1) on the filename?

gtsiolis:
Looking at this now! 👀

internal/snapshot/save.go
```diff
@@ -0,0 +1,63 @@
...
+			},
+		})
+		return output.NewSilentError(fmt.Errorf("LocalStack is not running"))
+	}
+
+	sink.Emit(output.SpinnerStart("Saving snapshot..."))
+	defer func() {
+		sink.Emit(output.SpinnerStop())
+		if retErr == nil {
+			sink.Emit(output.MessageEvent{Severity: output.SeveritySuccess, Text: fmt.Sprintf("Snapshot saved to %s", dest)})
```
gtsiolis: **praise**: Nice! 🤩  Adding some screenshots for future reference.

| Saving Snapshot | Saved Snapshot |
|-|-|
| Image | Image |

gtsiolis: **question**: There's some hard text wrapping on long file path. Any quick wins to make this render better on small width terminals? Like ...
- If it's under CWD → ./snapshot-….zip
- If it's under $HOME → ~/snapshot-….zip
- Otherwise leave the absolute path

| Text wrapping for saved snapshot on a small terminal width | 
|-|
| Image |

gtsiolis (approved):
UX LGTM on first pass. 🏁

internal/snapshot/destination.go
```diff
@@ -0,0 +1,73 @@
...
+	// ErrRemoteNotSupported is returned for known remote schemes (s3://, oras://, cloud:).
+	ErrRemoteNotSupported = errors.New("remote destinations are not yet supported — coming soon")
+	// ErrUnknownScheme is returned for unrecognized URL schemes.
+	ErrUnknownScheme = errors.New("unrecognized destination scheme")
+)
+
+// ParseDestination resolves the user-supplied path to an absolute local path.
+// When dest is empty, a default name based on now (UTC) is used, e.g.
+// "snapshot-2026-05-11T21-04-32.zip", saved in the current working directory.
+// The returned path always has a .zip extension.
```
gtsiolis, line 39: **question**: Can we drop the `.zip` filename later, following the [relevant discussion](https://www.notion.so/CLI-snapshots-352fc2a234318036b3fafc79fa15f98b?d=35dfc2a2343180c89360001cc70a0be2&source=copy_link#35dfc2a234318045af89e9910e3ee203)?

gtsiolis (commented):
Missed one important question before.

cmd/snapshot.go
anisaoshafi: @gtsiolis I think in this case it's better to add the 3-char hash in this case instead of the (1) append. In the spec there is this line `Overwrites an existing snapshot at the same REF without prompting`, in order to be consistent when a REF is specified. WDYT?

gtsiolis: @anisaoshafi Here's what I had in mind:
- Overwriting when using a name makes sense. **Named saves can overwrite (user's responsibility).** 
- Saving without a name is a bit trickier. **Unnamed saves need to be unique (system's responsibility).**

If trivial, I'd say let's go with the 3-char hash on auto-named snapshots. Added [a comment in the spec](https://www.notion.so/CLI-snapshots-352fc2a234318036b3fafc79fa15f98b?d=35ffc2a2343180a9a291001c7481a82c&source=copy_link#35dfc2a23431804b9d61fcd9acb2cf90), pushbacks on this are welcome if it's something we can ship later in a follow-up PR.

anisaoshafi: thanks for clarifying @gtsiolis, agreed. Will adjust to the hash when saving without name, that's trivial 🙏🏼

gtsiolis: Updated spec[[1](https://www.notion.so/localstack/CLI-snapshots-352fc2a234318036b3fafc79fa15f98b?source=copy_link#35ffc2a234318019a634df5ad7fdc35a)] to reflect this decision.

anisaoshafi: Done in https://github.com/localstack/lstk/pull/210/commits/658341deddd8a665a73cb462284f52076fbc157b 
image

internal/snapshot/save.go
anisaoshafi: Good idea, gave it a try https://github.com/localstack/lstk/pull/210/commits/c7c3f05e868f97139c6060a79f4c5c57299ec2ac:
Before | After
-------| ------
image | image

internal/snapshot/destination.go
anisaoshafi: Thanks for the heads up. I have this in mind. 
I'm checking if the emulator actually requires the `.zip` format before actually dropping it.

gtsiolis: Awesome. Just posted a [proposal](https://www.notion.so/CLI-snapshots-352fc2a234318036b3fafc79fa15f98b?d=35dfc2a2343180c89360001cc70a0be2&source=copy_link#35dfc2a234318045af89e9910e3ee203) to replace `.zip` with `.snapshot`, but possibly also ok to change later?

anisaoshafi: Agreed, this can be changed in a follow up PR 🙌🏼

internal/snapshot/save.go
gtsiolis: Smooth. 🧈 

| Saving in current directory, home, and other folder |
|--------|
| Screenshot 2026-05-13 at 14 25 51 |

test/integration/snapshot_save_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+	ctx := testContext(t)
+	startTestContainer(t, ctx)
+	srv := mockStateServer(t)
+	dir := t.TempDir()
+
+	stdout, stderr, err := runLstk(t, ctx, dir,
+		env.With(env.LocalStackHost, lsHost(srv)),
+		"--non-interactive", "snapshot", "save",
+	)
```
carole-lavillonniere, line 60: issue:

> - Never let an integration test inherit the developer's real `$HOME`. Pass an isolated env via `testEnvWithHome(t.TempDir(), "")` (or build on top of it with `env.With(...)`) instead of `nil` or `os.Environ()`. Inheriting HOME pollutes the user's `~/.config/lstk/`, `~/.aws/`, and `~/.cache/lstk/`, and makes parallel runs interfere through shared `lstk.log`, license cache, and file-keyring fallback.

Pasted from CLAUDE.md. There are a few tests affected.

cmd/snapshot.go
```diff
@@ -0,0 +1,85 @@
...
+			if err != nil {
+				return err
+			}
+
+			appConfig, err := config.Get()
+			if err != nil {
+				return fmt.Errorf("failed to get config: %w", err)
+			}
+
+			awsContainer := config.ContainerConfig{Type: config.EmulatorAWS, Port: config.DefaultAWSPort}
```
carole-lavillonniere:
```suggestion
			var awsContainer config.ContainerConfig
```
(since we never use the default value)

carole-lavillonniere (approved):
Added 2 comments, but good to go from my side after these were addressed!

test/integration/snapshot_save_test.go
anisaoshafi: Sorry for missing this 😬, addressed here 🙏🏼 https://github.com/localstack/lstk/pull/210/commits/60503f4f85c3f92254edd8c0397a0d3327e75b96

gtsiolis:
That's one small step for snapshots, one giant leap for LocalStack. 💿

PR #211, opened 2026-04-28
Integration tests parallelism - docker-in-docker

carole-lavillonniere:


carole-lavillonniere:
Closing for now in favor of CI sharding https://github.com/localstack/lstk/pull/213. We already went from 12+min to ~3min to run the integration tests on CI.

This PR was trying to implement docker-in-docker (dind). Why? If running all tests in parallel, localstack containers conflict with each over because of names and bound ports. With dind each test gets its own `docker:dind` container, which is a full docker daemon running inside a container. Tests then point lstk at that daemon via `DOCKER_HOST=tcp://127.0.0.1:<random-port>`.

Blocker: we were hitting the resources limit on CI, since each test needs to pull` localstack/localstack-pro` (~2GB)

PR #212, opened 2026-04-28
Hide header when already logged-in

gtsiolis:
Hide header/banner when already logged in (env var or stored token), only emits a colored note and exits without showing the header.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-04-28 at 16 47 58 | Screenshot 2026-04-28 at 16 44 14 |

### DIFF
--- a/internal/output/plain_format.go
+++ b/internal/output/plain_format.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 	"time"
+
+	"github.com/localstack/lstk/internal/ui/styles"
 )
 
 // FormatEventLine converts an output event into a single display line.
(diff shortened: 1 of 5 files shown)


gtsiolis:
Cc @carole-lavillonniere in case this relates to https://github.com/localstack/lstk/pull/208#issuecomment-4334077504.

internal/output/plain_format.go
```diff
@@ -118,13 +120,13 @@ func formatAuthEvent(e AuthEvent) string {
...
+		return styles.Success.Render(SuccessMarker()) + " " + e.Text
 	case SeverityNote:
-		return "> Note: " + e.Text
+		return styles.Secondary.Render(">") + " " + styles.Note.Render("Note:") + " " + e.Text
 	case SeverityWarning:
-		return "> Warning: " + e.Text
+		return styles.Secondary.Render(">") + " " + styles.Warning.Render("Warning:") + " " + e.Text
 	case SeveritySecondary:
-		return e.Text
+		return styles.SecondaryMessage.Render(e.Text)
```
carole-lavillonniere: issue: the convention so far is to not do any styling in non-TTY.
What about keeping the TUI in this case but just hide the header, as we do in a few places already: https://github.com/localstack/lstk/blob/60b92a085176610d0a9234f7055ad01b3bfdc8c2/internal/ui/run_awsconfig.go#L29

gtsiolis: Nice catch, added 4193d24aa8c9e4d1fabb4fffcdd76742641d1cca. Let me know what you think!

carole-lavillonniere (approved):
Pushed a small clean-up commit, added 2 integration tests and updated the PR title and description.

PR #213, opened 2026-04-28
 Integration tests parallelism - CI sharding

carole-lavillonniere:
## Summary

Shards Ubuntu integration tests across 4 GitHub Actions runners. macOS and Windows stay single-runner.
Slowest CI job overall before: 6m40s, after (ubuntu shard 4/4): 3m07s.

## Changes

- **`.github/workflows/ci.yml`** — replace flat `os` matrix with `include:` entries adding `shard` / `shard_total` for Ubuntu only.
- **`Makefile`** — `test-integration` now delegates to `scripts/test-integration.sh`.
- **`scripts/test-integration.sh`** (new) — when `SHARD_TOTAL` is set, runs `go test -list` once, partitions top-level test names round-robin by sorted name, and passes the resulting `-run` regex to gotestsum. When unset, behaves exactly like the previous Makefile invocation. `RUN=` still works for `make test-integration RUN=...` locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,10 +86,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Ubuntu integration tests are sharded across 4 runners — they dominate
-        # wall time (~7min unsharded). macOS and Windows stay single-runner;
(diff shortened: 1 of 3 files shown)


anisaoshafi (approved):
LGTM, this is going to save us so much time 👏🏼

PR #215, opened 2026-04-30
Skip stale Docker sockets when probing for the daemon

gtsiolis:
## Problem
When Docker Desktop is installed alongside another Docker runtime (e.g., Colima or OrbStack), Docker Desktop can leave behind a stale Unix socket at `~/.docker/run/docker.sock` even when it is not running. This causes lstk to incorrectly probe the stale socket and fail to connect, even though a working Docker daemon is available through the other runtime.

This was reported by a user who had both Docker Desktop and Colima installed. Docker Desktop had been used previously but was only \"stopped\" (not fully exited), leaving its socket file in place. Colima was running and functional, but lstk failed because it tried the stale Docker Desktop socket first.

## Solution
When probing candidate Docker sockets, attempt a \"hello\" ping to each socket and skip any that return `connection refused` or `no such file` errors. The first responsive socket is used. This ensures that stale or orphaned sockets are ignored and a live daemon is found.

## Changes
- Update socket probing loop to test connectivity before selecting a socket
- Add test coverage for stale socket fallback behavior

See DES-212 and [relevant discussion](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1777550172961019?thread_ts=1773771243.122349&cid=C0A1857LGTU) for more context.

### DIFF
--- a/internal/runtime/docker.go
+++ b/internal/runtime/docker.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


gtsiolis:
Possibly not needed, as this bug only occurs when running Docker and Colima side-by-side.

carole-lavillonniere (approved):
Looks good! Can we add a PR description to explain when/how this can happen though?

gtsiolis:
Updated description. Thanks for taking a look, @carole-lavillonniere! 🍵

PR #217, opened 2026-04-30
Select default emulator on first run

anisaoshafi:
First-time users are prompted to choose their emulator (AWS or Snowflake) before lstk starts (to simulate this, rename or delete config.toml). The choice is saved to the config file immediately.
- When run with `--non-interactive` / no TTY it silently uses the AWS by default.

Screenshot 2026-04-30 at 16 59 20
Note to user they can change to the other emulator:
Screenshot 2026-04-30 at 16 59 43


internal/config/switch.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil
+	}
+
+	if err := os.WriteFile(path, []byte(updated), 0644); err != nil {
+		return fmt.Errorf("failed to write config file: %w", err)
+	}
+	return loadConfig(path)
+}
+
+func switchEmulatorContent(content string, to EmulatorType) (updated string, changed bool, err error) {
```
carole-lavillonniere: nit: we can remove the err in the returned values since we never return an error

cmd/root.go
```diff
@@ -173,32 +194,72 @@ func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *t
...
+			},
 		})
 	}
 
 	sink := output.NewPlainSink(os.Stdout)
+	if firstRun && requestedEmulator == "" && len(appConfig.Containers) > 0 {
+		emName := appConfig.Containers[0].Type.DisplayName()
+		sink.Emit(output.MessageEvent{
+			Severity: output.SeverityNote,
+			Text:     fmt.Sprintf("No emulator configured; defaulting to %s. Use --emulator to change this.", emName),
```
carole-lavillonniere: question: isn't the message slightly misleading since the default emulator is written to the config?
suggestion: `Configured with default emulator AWS. Pass --emulator to change.`

internal/config/switch.go
```diff
@@ -0,0 +1,175 @@
...
+	for _, line := range lines {
+		effective := strings.TrimSpace(line)
+		if isCommented {
+			effective = strings.TrimSpace(strings.TrimPrefix(effective, "#"))
+		}
+		if m := typeLineRe.FindStringSubmatch(effective); m != nil {
+			return EmulatorType(strings.ToLower(m[1]))
+		}
+	}
+	return ""
```
carole-lavillonniere: nit: 
If the user has such config and requests "snowflake", it would match (it would ignore the fact that line is commented out)
```
[[containers]]
# type = "snowflake" 
type = "aws"
```

internal/config/switch.go
```diff
@@ -0,0 +1,175 @@
...
+			end:         end,
+			emulType:    detectBlockType(lines[i:end], isCommented),
+			isCommented: isCommented,
+		})
+		i = end - 1
+	}
+	return blocks
+}
+
+var typeLineRe = regexp.MustCompile(`type\s*=\s*"(\w+)"`)
```
carole-lavillonniere: nit: this regex only matches double quotes, can we update it so it also matches single quotes? e.g.
```
[[containers]]
type = 'aws'
```

and

```
[[containers]]
type = "aws"
```

test/integration/emulator_select_test.go
```diff
@@ -0,0 +1,103 @@
...
+	require.NoError(t, os.WriteFile(configPath, []byte(`[[containers]]
+type = "aws"
+tag  = "latest"
+port = "4566"
+`), 0644))
+
+	ctx := testContext(t)
+	// The process will fail at container.Start (no Docker / no real auth), but the
+	// config switch happens earlier so the file should already be updated.
+	_, _, _ = runLstk(t, ctx, "", e.With(env.AuthToken, "test-token"), "--emulator", "snowflake", "--non-interactive")
```
carole-lavillonniere: nit: can we still check returned error/add assertions for the expected failures, to make sure we pass for the right reason?

cmd/root.go
```diff
@@ -173,32 +194,72 @@ func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *t
...
+	switch config.EmulatorType(strings.ToLower(s)) {
+	case config.EmulatorAWS:
+		return config.EmulatorAWS, nil
+	case config.EmulatorSnowflake:
+		return config.EmulatorSnowflake, nil
+	default:
+		return "", fmt.Errorf("unsupported emulator %q: must be 'aws' or 'snowflake'", s)
+	}
+}
+
```
carole-lavillonniere: nit: can we move this function next to `EmulatorType` in` internal/config/containers.go` to keep it in its domain?

cmd/root.go
```diff
@@ -152,13 +158,28 @@ func buildStartOptions(cfg *env.Env, appConfig *config.Config, logger log.Logger
...
-func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *telemetry.Client, logger log.Logger, persist bool) error {
-
+func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *telemetry.Client, logger log.Logger, persist bool, firstRun bool, requestedEmulator string) error {
 	appConfig, err := config.Get()
 	if err != nil {
 		return fmt.Errorf("failed to get config: %w", err)
 	}
 
+	if requestedEmulator != "" {
+		emType, err := parseEmulatorType(requestedEmulator)
```
carole-lavillonniere: nit: can we parse the emulator type earlier, right after reading the flag in `cmd/root.go` line 40, to fail early and strong-type it when passing it down. I would also make it a pointer `*config.EmulatorType` so that an unset value is clear (nil), rather than relying on empty string.

internal/config/switch.go
```diff
@@ -0,0 +1,175 @@
...
+port = "4566"
+# volume = ""    # Host directory for persistent state (default: OS cache dir)
+# env = []       # Named environment profiles to apply (see [env.*] sections below)`
+
+const snowflakeContainerBlock = `[[containers]]
+type = "snowflake"
+tag  = "latest"
+port = "4566"
+# volume = ""    # Host directory for persistent state (default: OS cache dir)
+# env = []       # Named environment profiles to apply (see [env.*] sections below)`
```
carole-lavillonniere: question: 
At the moment those blocks are duplicated and there is a chance of drift.
How hard it is to parse those blocks from `default_config.toml`?
Another solution could be to use go templates in `default_config.toml` so the container blocks could come from another file.

carole-lavillonniere:
@anisaoshafi @gtsiolis it feels a bit odd to me that the `--emulator` flag overwrites the config. The convention for flags as far as I know is to be transient/ephemeral.

Users can still manually edit their config if they want to.

If we want to provide a way to reset the emulator in the config from command line, what about something like `lstk start --reset-config` or `lstk config emulator` that would prompt again `Which emulator would you like to use`?

`lstk start --emulator <type>` could override the config, but only temporarily.

internal/config/emulator_type_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
```
carole-lavillonniere, line 12: nit: these tests seem to be a good fit for table-driven tests, WDYT?

anisaoshafi:
> @anisaoshafi @gtsiolis it feels a bit odd to me that the `--emulator` flag overwrites the config. The convention for flags as far as I know is to be transient/ephemeral.
> 
> Users can still manually edit their config if they want to.
> 
> If we want to provide a way to reset the emulator in the config from command line, what about something like `lstk start --reset-config` or `lstk config emulator` that would prompt again `Which emulator would you like to use`?
> 
> `lstk start --emulator <type>` could override the config, but only temporarily.

Good point, can be too much of a responsibility for a flag. 

I like `lstk config emulator` option. Does the same but in a more controlled way since the user have to confirm the option, or cancel if that's not what they intended. 
Will wait for @gtsiolis's opinion before making the changes 🙏🏼

cmd/root.go
```diff
@@ -173,32 +194,72 @@ func startEmulator(ctx context.Context, rt runtime.Runtime, cfg *env.Env, tel *t
...
-			LabelCh:       labelCh,
+			Runtime:                rt,
+			Version:                version.Version(),
+			StartOptions:           opts,
+			NotifyOptions:          notifyOpts,
+			ConfigPath:             configPath,
+			EmulatorLabel:          config.CachedPlanLabel(),
+			LabelCh:                labelCh,
+			NeedsEmulatorSelection: needsEmulatorSelection,
+			OnEmulatorSelected: func(emType config.EmulatorType) ([]config.ContainerConfig, error) {
```
carole-lavillonniere: issue: such callback is not very "go-like". It's not clear to me why we need it, don't we have all dependencies in `ui/run.go` to execute what we must execute after selection?

carole-lavillonniere:
> > @anisaoshafi @gtsiolis it feels a bit odd to me that the `--emulator` flag overwrites the config. The convention for flags as far as I know is to be transient/ephemeral.
> > Users can still manually edit their config if they want to.
> > If we want to provide a way to reset the emulator in the config from command line, what about something like `lstk start --reset-config` or `lstk config emulator` that would prompt again `Which emulator would you like to use`?
> > `lstk start --emulator <type>` could override the config, but only temporarily.
> 
> Good point, can be too much of a responsibility for a flag.
> 
> I like `lstk config emulator` option. Does the same but in a more controlled way since the user have to confirm the option, or cancel if that's not what they intended. Will wait for @gtsiolis's opinion before making the changes 🙏🏼

If we decide for another command, I think we can leave it out of the scope of this PR/create another issue for it to keep this PR small.

internal/config/switch.go
anisaoshafi: Tackled here:https://github.com/localstack/lstk/pull/217/changes/16841526292b95dc13a5774d64aed45d65fd3f5c

cmd/root.go
anisaoshafi: Good point, tackled here: https://github.com/localstack/lstk/pull/217/changes/16841526292b95dc13a5774d64aed45d65fd3f5c

internal/config/switch.go
anisaoshafi: thanks for the heads up, didn't properly test it. Handled it here: https://github.com/localstack/lstk/pull/217/changes/99272768aa9e0dbc5771fea70424caf9dc2cade3

internal/config/switch.go
anisaoshafi: thoughtless of me 😬 thanks for raising. Addressed it here: https://github.com/localstack/lstk/pull/217/changes/99272768aa9e0dbc5771fea70424caf9dc2cade3

test/integration/emulator_select_test.go
anisaoshafi: Yes, addressed here: https://github.com/localstack/lstk/pull/217/changes/3706dc869bcffbc6bc803d84e05615e1e494012a

cmd/root.go
anisaoshafi: Moved it here, good point: https://github.com/localstack/lstk/pull/217/changes/b85896ca7f40e672cdb001e2c501b59d319c1e55

internal/config/emulator_type_test.go
anisaoshafi: great observation, done: https://github.com/localstack/lstk/pull/217/changes/abc89ac99f20b48f4a9be6f2dddf00622283c1fb

cmd/root.go
anisaoshafi: Attempted here: https://github.com/localstack/lstk/pull/217/changes/aa2efc936f75829f6b3dccd4bbaf7ff51ab87d81

gtsiolis:
Looking at this now! 👀

internal/config/switch.go
```diff
@@ -0,0 +1,174 @@
...
+type = "aws"
+tag  = "latest"
+port = "4566"
+# volume = ""    # Host directory for persistent state (default: OS cache dir)
+# env = []       # Named environment profiles to apply (see [env.*] sections below)`
+
+const snowflakeContainerBlock = `[[containers]]
+type = "snowflake"
+tag  = "latest"
+port = "4566"
```
gtsiolis: **issue**: Similar to what @carole-lavillonniere [mentioned above](https://github.com/localstack/lstk/pull/217/changes#r3186817626), this leads to content duplication.

```sh

# [[containers]]
# type = "aws"     # Emulator type. Currently supported: "aws", "snowflake"
# tag  = "latest"  # Docker image tag, e.g. "latest", "2026.03"
# port = "4566"    # Host port the emulator will be accessible on
# # volume = ""    # Host directory for persistent state (default: OS cache dir)
# # env = []       # Named environment profiles to apply (see [env.*] sections below)

[...]

[[containers]]
type = "snowflake"
tag  = "latest"
port = "4566"
```

I'd expect if I select SNO on start, to simply use snowflake in the first [[containers]] section.

```sh
[...]
[[containers]]
type = "snowflake"     # Emulator type. Currently supported: "aws", "snowflake"
tag  = "latest"  # Docker image tag, e.g. "latest", "2026.03"
volume = ""    # Host directory for persistent state (default: OS cache dir)
env = []       # Named environment profiles to apply (see [env.*] sections below)
[...]
```

cmd/root.go
carole-lavillonniere: Looks good! Are we now able to inline `ParseEmulatorType` in `ParseOptionalEmulatorType` (if not used anywhere else)?

cmd/root.go
anisaoshafi: Attempted to refactor here, let me know if I'm doing sth terrible 🙏🏼https://github.com/localstack/lstk/pull/217/commits/5e7723a1e86789ebc074fb258b4a37be565e8398

cmd/root.go
anisaoshafi: Good idea, tried here: https://github.com/localstack/lstk/pull/217/commits/fe5fa045dbebe8c337828b33519d26cfa4f14858

internal/ui/run.go
```diff
@@ -96,6 +115,63 @@ func Run(parentCtx context.Context, runOpts RunOptions) error {
...
+	ctx context.Context,
+	sink output.Sink,
+	configPath string,
+) ([]config.ContainerConfig, error) {
+	responseCh := make(chan output.InputResponse, 1)
+	sink.Emit(output.UserInputRequestEvent{
+		Prompt: "Which emulator would you like to use?",
+		Options: []output.InputOption{
+			{Key: "a", Label: "AWS [A]"},
+			{Key: "s", Label: "Snowflake [S]"},
```
gtsiolis: **suggestion**: Thinking once we add AZU emulation this could break unless we add [Z] for AZU. Alternatively, we could go with a boring [1], [2], [3] ... Thoughts?

internal/ui/run.go
```diff
@@ -96,6 +115,63 @@ func Run(parentCtx context.Context, runOpts RunOptions) error {
...
+		return nil, fmt.Errorf("failed to switch emulator: %w", err)
+	}
+	newCfg, err := config.Get()
+	if err != nil {
+		return nil, err
+	}
+
+	msg := selected.DisplayName() + " emulator selected."
+	if configPath != "" {
+		msg += fmt.Sprintf(" You can change this anytime in %s.", configPath)
```
gtsiolis: **thought**: Yes, you can, but this would break LocalStack in most cases where 90% of users have access to only one of the emulators at the moment, right? Maybe a softer copy like:


```suggestion
		msg += fmt.Sprintf(" Change configuration in %s.", configPath)
```

internal/ui/run.go
gtsiolis: **issue(non-blocking)**: Here's what most users with no access to SNO emulation will see. I guess this comes from the emulator, right? Could we mask now or later this error?

Image

SNO users in trial see no error currently. 

Have to check what SNO-only users in production see.

internal/ui/run.go
gtsiolis: **issue**: The message here wraps early on line 2, can we make the message wrap normally in two lines, I think we do so for other messages, right?

Image

internal/ui/run.go
```diff
@@ -96,6 +115,63 @@ func Run(parentCtx context.Context, runOpts RunOptions) error {
...
 
+func resolveAndCacheLabel(ctx context.Context, opts container.StartOptions, labelCh chan<- string) {
+	label, ok := container.ResolveEmulatorLabel(ctx, opts.PlatformClient, opts.Containers, opts.AuthToken, opts.Logger)
+	if ok {
+		config.CachePlanLabel(label)
+	}
+	labelCh <- label
+}
+
+func selectEmulatorInTUI(
```
gtsiolis: **praise**: Thanks for reusing this UX!

internal/ui/run.go
anisaoshafi: good point, hadn't thought about the clashing `A`, but `Z` is going to be fine I think. But if you think it's more user friendly I'm fine with the numeric alternative.

internal/config/switch.go
anisaoshafi: I'm going to tackle this 💯

internal/ui/run.go
anisaoshafi: Handled here: [eec8f75](https://github.com/localstack/lstk/pull/217/commits/eec8f75d0215ee9caed58873e821b1673042d176)

internal/ui/run.go
gtsiolis: I'd either skip the keyboard shortcuts, or keep them as is for now and remove later. Your call.

anisaoshafi: Done here: https://github.com/localstack/lstk/pull/217/commits/b4e27eaed261dca68f75f7ef826792807ca988db

anisaoshafi: yes, comes from emulator side. I created a follow up task: DRG-818 for handling this error in a nicer way since it's out of the scope for this PR 🙏🏼

internal/config/switch.go
anisaoshafi: Addressed here 🙏🏼 https://github.com/localstack/lstk/pull/217/commits/0aeca91421fd718b8231916fa655fd82a23efcbf

internal/config/switch.go
anisaoshafi: Addressed here 🙏🏼 https://github.com/localstack/lstk/pull/217/commits/0aeca91421fd718b8231916fa655fd82a23efcbf

internal/ui/run.go
anisaoshafi: @gtsiolis Does this look a bit better? Also changed the message as you proposed in the other comment image

gtsiolis: What do you think of using the same success message as the AWS profile creation? This will avoid ending up with multiple NOTE messages when the emulator is running (although edge case) and make the output more functional upon selection.


| 1/2 | 2/2 |
| -- | -- |
|  |  |

anisaoshafi: > same success message as the AWS profile creation

What does this mean when the user selects snowflake on first try?

@gtsiolis sorry, I'm not following. Can you rephrase your suggestion? 
Did you mean `Change configuration in ~/.config/lstk/config.toml` should have a green tick in front?

anisaoshafi:
Apologies for the back and forth and the huge PR. I removed the `--emulator` flag from the scope of this PR (discussing it in linear), should have removed that before tackling the review comments which were mostly around the switching between emulators, but a bit too late now.
This simplifies the PR. Thanks for your patience @carole-lavillonniere @gtsiolis 🙏🏼

internal/ui/run.go
```diff
@@ -98,6 +117,62 @@ func Run(parentCtx context.Context, runOpts RunOptions) error {
...
+	}
+
+	sink.Emit(output.MessageEvent{Severity: output.SeveritySuccess, Text: selected.DisplayName() + " emulator selected."})
+	if configPath != "" {
+		sink.Emit(output.MessageEvent{Severity: output.SeveritySecondary, Text: "Change configuration in " + configPath + "."})
+	}
+
+	return newCfg.Containers, nil
+}
+
```
carole-lavillonniere: issue: now I think we should just move `selectEmulatorInTUI` and `resolveAndCacheLabel` out of `internal/ui` and we should be good. These are domain code.

`AGENTS.md` is kind of documenting this with:
> - Do not pass UI callbacks like `onProgress func(...)` through domain layers; prefer typed output events.

> 2. Keep UI as presentation/orchestration only; business logic stays in domain packages.

domain packages => everything under `internal/` except `internal/ui/`.
A test for "is this domain code?": if we wired a different kind of frontend, domain code should still work. 

`AGENTS.md` does not explicitly say where the code executed upon user input lives, though. I think it'd be good to add a line about this.

internal/config/containers.go
```diff
@@ -25,6 +25,12 @@ var emulatorDisplayNames = map[EmulatorType]string{
 	EmulatorAzure:     "Azure",
 }
 
+func (e EmulatorType) DisplayName() string {
+	if name, ok := emulatorDisplayNames[e]; ok {
+		return name
+	}
+	return string(e)
+}
```
carole-lavillonniere, line 41: question: we already have `DisplayNameForType` line 77.
Can we 
- refactor to make them one?
- or rename to make more explicit?
- and/or change the signatures so that they are both methods?

internal/ui/run.go
```diff
@@ -98,6 +117,62 @@ func Run(parentCtx context.Context, runOpts RunOptions) error {
...
+	ctx context.Context,
+	sink output.Sink,
+	configPath string,
+) ([]config.ContainerConfig, error) {
+	responseCh := make(chan output.InputResponse, 1)
+	sink.Emit(output.UserInputRequestEvent{
+		Prompt: "Which emulator would you like to use?",
+		Options: []output.InputOption{
+			{Key: "a", Label: "AWS"},
+			{Key: "s", Label: "Snowflake"},
```
carole-lavillonniere: nit: possible to build the options from `emulatorDisplayNames` or another map, to make it more dynamic?

cmd/root.go
```diff
@@ -296,5 +294,20 @@ func initConfig(cmd *cobra.Command, _ []string) error {
...
+		path, err := cmd.Flags().GetString("config")
+		if err != nil {
+			return err
+		}
+		if path != "" {
+			return config.InitFromPath(path)
+		}
+		*firstRun, err = config.Init()
+		return err
+	}
```
carole-lavillonniere, line 303: question: do we really need 2 functions `initConfig` and `initConfigCapturingFirstRun`? They look quite similar

test/integration/emulator_select_test.go
```diff
@@ -0,0 +1,76 @@
...
+		close(outputCh)
+	}()
+
+	require.Eventually(t, func() bool {
+		return bytes.Contains(out.Bytes(), []byte("Which emulator would you like to use?"))
+	}, 10*time.Second, 100*time.Millisecond, "emulator selection prompt should appear on first run")
+
+	cancel()
+	<-outputCh
+}
```
carole-lavillonniere, line 115: suggestion: what about completing this test with an actual user input, making sure we persist the choice in the config and start that chosen emulator?

carole-lavillonniere: suggestion: we could also add a simple test to make sure we don't prompt when the config is already created?

internal/ui/run.go
anisaoshafi: Attempted to address here: https://github.com/localstack/lstk/pull/217/changes/fe9352933b1f4ae7c9a15934149228a91c21fac5

internal/config/containers.go
anisaoshafi: Tried here: https://github.com/localstack/lstk/pull/217/changes/62926ca1f6ee07c6fdbfb17e5434e74bd87c1cef

internal/ui/run.go
anisaoshafi: Tried here: https://github.com/localstack/lstk/pull/217/changes/b4fc4cf797438093eda44aaa766ed8e167a972d1

cmd/root.go
anisaoshafi: Good point, tried here: https://github.com/localstack/lstk/pull/217/changes/6ced3fe963d008d160315f3525333c9c3b3c72bf

test/integration/emulator_select_test.go
anisaoshafi: Good point: tried [here](https://github.com/localstack/lstk/pull/217/changes/238b97acbaded9b158cdba99f7e131bb8873c784) and [here](https://github.com/localstack/lstk/pull/217/changes/3d88700f0ad5f2be2a1a3ba0e9ac643a716e6eb3) 🙏🏼

carole-lavillonniere (approved):
🥳

PR #219, opened 2026-05-04
Add --persist flag to enable persistence

anisaoshafi:
Add flag `--persist` to start LocalStack with local persistence enabled (similar to setting LOCALSTACK_PERSISTENCE=1 env var).

### DIFF
--- a/cmd/restart.go
+++ b/cmd/restart.go
@@ -16,7 +16,7 @@ import (
 )
 
 func newRestartCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.Command {
-	return &cobra.Command{
+	cmd := &cobra.Command{
 		Use:     "restart",
 		Short:   "Restart emulator",
 		Long:    "Stop and restart emulator and services.",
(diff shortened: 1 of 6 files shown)


cmd/restart.go
```diff
@@ -35,7 +35,7 @@ func newRestartCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobr
 			stopOpts := container.StopOptions{
 				Telemetry: tel,
 			}
-			startOpts := buildStartOptions(cfg, appConfig, logger, tel)
+			startOpts := buildStartOptions(cfg, appConfig, logger, tel, false)
```
carole-lavillonniere: suggestion: should we also add the flag to restart command? Otherwise it might be confusing when users started with `--persist` and then restart. cc @gtsiolis

carole-lavillonniere (approved):
I added one suggestion, but it looks good!

cmd/restart.go
anisaoshafi: Done, thanks! https://github.com/localstack/lstk/pull/219/commits/a02347e5d2c5ee253b5e5840e3ff32ed7f3905eb

PR #225, opened 2026-05-05
Report emulator type mismatch

carole-lavillonniere:
## Motivation

`lstk` would emit a confusing version-mismatch warning when a config targeted a different emulator type than the one already running on the port (e.g. snowflake config, AWS instance running). User-facing messages also varied between commands: `lstk status` showed `LocalStack Snowflake Emulator is not running` while `lstk stop` only said `LocalStack is not running`.

## Changes

- Detect emulator type mismatch in `selectContainersToStart` by widening the running-container scan to all known LocalStack image repos and comparing emulator types.
- Replace generic `LocalStack ...` strings in start/stop/logout flows with `DisplayName()` so messages match `lstk status` (`LocalStack AWS Emulator is not running`).
- Replace `AnyRunning` with `RunningEmulators` returning the running configs; add `StillRunningMessage` helper that handles singular/plural grammar.
- Narrow `resolveRunningContainerName` scan to repos matching the configured emulator type via new `config.KnownImageReposForType(t)` so `stop`/`logout` don't falsely report a different emulator as the configured one.
- Document the Docker image naming convention (full image / image repo / product name) in `CLAUDE.md`.

Closes DRG-806

## Example with aws running, snowflake in the config:

image

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -396,7 +396,7 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 			foundType := config.EmulatorTypeForImage(found.Image)
 			if foundType != "" && foundType != c.EmulatorType {
 				sink.Emit(output.ErrorEvent{
-					Title:   fmt.Sprintf("%s is already running on port %s", config.DisplayNameForType(foundType), found.BoundPort),
+					Title:   fmt.Sprintf("%s is running on port %s", config.DisplayNameForType(foundType), found.BoundPort),
 					Summary: fmt.Sprintf("Your config specifies the %s. Only one emulator can run on a port at a time.", config.DisplayNameForType(c.EmulatorType)),
 					Actions: []output.ErrorAction{
 						{Label: "Stop the running emulator:", Value: fmt.Sprintf("docker stop %s", found.Name)},
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -217,7 +217,7 @@ func runPostStartSetups(ctx context.Context, sink output.Sink, containers []conf
 }
 
 func emitAlreadyRunning(sink output.Sink, c runtime.ContainerConfig, localStackHost, webAppURL string) {
-	sink.Emit(output.MessageEvent{Severity: output.SeverityNote, Text: "LocalStack is already running"})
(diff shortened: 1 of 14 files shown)


internal/container/start.go
```diff
@@ -388,15 +388,32 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
-		found, err := rt.FindRunningByImage(ctx, []string{imageRepo, "localstack/localstack"}, c.ContainerPort)
+		found, err := rt.FindRunningByImage(ctx, config.KnownImageRepos(), c.ContainerPort)
 		if err != nil {
 			return nil, fmt.Errorf("failed to scan for running containers: %w", err)
 		}
 		if found != nil {
+			foundType := config.EmulatorTypeForImage(found.Image)
+			if foundType != "" && foundType != c.EmulatorType {
+				sink.Emit(output.ErrorEvent{
+					Title:   fmt.Sprintf("%s is already running on port %s", config.DisplayNameForType(foundType), found.BoundPort),
```
anisaoshafi: small nit: I'd omit the word `already`. When you `lstk start` with **aws**, then change config to **snowflake** and then `lstk start` again this message `LocalStack AWS Emulator is already running on port 4566` might feel like the user wanted to start aws, not snowflake. The port is already busy, but the emulator the user intended is not.

internal/container/start.go
```diff
@@ -388,15 +388,32 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
...
+					Summary: fmt.Sprintf("Your config specifies the %s. Only one emulator can run on a port at a time.", config.DisplayNameForType(c.EmulatorType)),
+					Actions: []output.ErrorAction{
+						{Label: "Stop the running emulator:", Value: fmt.Sprintf("docker stop %s", found.Name)},
```
anisaoshafi, line 402: thought for later:  we could as well replace `docker stop` with`lstk stop`. Though with stop's current behaviour it would find nothing to stop. But we could adjust it to stop any running emulators it finds.

anisaoshafi (approved):
LGTM, thanks 👏🏼

internal/container/start.go
carole-lavillonniere: Yes good point :+1:

internal/container/start.go
carole-lavillonniere: Makes more sense, updated :+1:

PR #226, opened 2026-05-06
Set LOCALSTACK_HOST to fix custom port errors

carole-lavillonniere:
## Motivation

When starting an emulator on a non-default port (e.g. `port = "4567"`), LocalStack rejects browser requests to `https://snowflake.localhost.localstack.cloud:4567/` with 403s and CORS-blocked errors. Root cause: lstk only set `GATEWAY_LISTEN=:4566,:443` and never told LocalStack about the external port, so the gateway's host validation and CORS allowlist treated the configured external port as unknown.

Towards DRG-805

## Changes

- Set `LOCALSTACK_HOST=localhost.localstack.cloud:<port>` in container env so LocalStack advertises the correct external host:port and accepts requests / origins on it. It worked for default port because the default value is `localhost.localstack.cloud:4566`

## Tests

- `TestStartCommandSucceedsWithNonDefaultPort` now asserts `LOCALSTACK_HOST=localhost.localstack.cloud:4567`. Fails before the fix, passes after.
- `TestStartCommandSetsUpContainerCorrectly` asserts the default-port equivalent for symmetry.
- Manually verified with custom port 4567 + snowflake `2026.3` (latest has an issue), loaded browser at https://snowflake.localhost.localstack.cloud:4567, it rendered without error:

image

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -100,6 +100,7 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, opts Start
 			"LOCALSTACK_AUTH_TOKEN="+token,
 			"GATEWAY_LISTEN=:4566,:443",
 			"MAIN_CONTAINER_NAME="+containerName,
+			"LOCALSTACK_HOST="+endpoint.Hostname+":"+c.Port,
 		)
 
 		env = append(env, hostEnv...)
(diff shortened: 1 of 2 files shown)


anisaoshafi (approved):
Nice finding 🕵🏼

PR #227, opened 2026-05-06
Add Snowflake post-start hints

carole-lavillonniere:
## Motivation

Snowflake users see no post-start hints today (endpoint, tip). The bare `• Endpoint:` line printed for AWS is misleading for Snowflake — Snowflake clients need the `snowflake.<host>` form to route through the gateway.

## Changes

- New `internal/emulator/snowflake` package with `Endpoint(resolvedHost)` that builds `http://snowflake.<host>`; returns `""` when the host is an IP (subdomain on an IP is invalid).
- For Snowflake, replace `• Endpoint:` with `• Snowflake endpoint: http://snowflake.<host>`; fall back to the bare endpoint when the snowflake-prefixed form is invalid (DNS rebind / IP host).
- Show a randomized `> Tip:` line for Snowflake too (previously AWS only).
- `runPostStartSetups` now emits post-start pointers for every unique emulator type, not only types with a registered setup function — so Snowflake gets pointers without needing a setup hook.

Towards DRG-805

image

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -19,6 +19,7 @@ import (
 	"github.com/localstack/lstk/internal/auth"
 	"github.com/localstack/lstk/internal/awsconfig"
 	"github.com/localstack/lstk/internal/config"
+	"github.com/localstack/lstk/internal/emulator/snowflake"
 	"github.com/localstack/lstk/internal/endpoint"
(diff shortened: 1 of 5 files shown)


anisaoshafi (approved):
Tested and looks good 🚀

PR #228, opened 2026-05-06
Snowflake endpoint in lstk status output

carole-lavillonniere:
## Motivation

`lstk status` for Snowflake currently prints the AWS endpoint  `localhost.localstack.cloud:4566`. Snowflake clients need the `snowflake.<host>` form (the same fix already applied to the post-start hint in #227).

Towards DRG-805

image


anisaoshafi (approved):
👏🏼

PR #229, opened 2026-05-06
Snowflake version in lstk status output

carole-lavillonniere:
## Motivation

`lstk status` displays the snowflake version (just like it does for aws)

Closes DRG-805

## Changes

- New `internal/emulator/snowflake/Client` mirroring the AWS client; `FetchVersion` hits `/_localstack/health` and parses the `version` field. `FetchResources` is a no-op since Snowflake does not expose AWS-style resources.
- `container.Status` now takes `clients map[config.EmulatorType]emulator.Client` instead of a single AWS client.

image

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -6,7 +6,9 @@ import (
 
 	"github.com/localstack/lstk/internal/config"
 	"github.com/localstack/lstk/internal/container"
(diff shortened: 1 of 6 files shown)


anisaoshafi (approved):
neat 👏🏼

PR #230, opened 2026-05-06
Improve error for users with no snowflake license

anisaoshafi:
Follow up from https://github.com/localstack/lstk/pull/217#discussion_r3190134661
Improve Snowflake access error for users that don't have it in their licence, and guide them on how to continue.

| Before  | After |
| Screenshot 2026-05-06 at 18 32 40 | Screenshot 2026-05-06 at 19 09 57

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -376,7 +376,8 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink, 
 				sink.Emit(output.ErrorEvent{
 					Title: "Your license does not include the Snowflake emulator.",
 					Actions: []output.ErrorAction{
-						{Label: "Start a free Snowflake trial:", Value: "https://www.localstack.cloud/pricing?tab=snowflake"},
+						{Label: "Sign up for a free trial:", Value: "https://app.localstack.cloud/sign-up"},
+						{Label: "Contact our team:", Value: "https://www.localstack.cloud/demo"},
 					},
 				})
 				err = output.NewSilentError(err)
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -369,11 +369,23 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink, 
 		sink.Emit(output.ContainerStatusEvent{Phase: "waiting", Container: c.Name})
 		healthURL := fmt.Sprintf("http://localhost:%s%s", c.Port, c.HealthPath)
 		if err := awaitStartup(ctx, rt, sink, containerID, "LocalStack", healthURL); err != nil {
+			errCode := telemetry.ErrCodeStartFailed
+			var licErr *licenseNotCoveredError
+			if errors.As(err, &licErr) && c.EmulatorType == config.EmulatorSnowflake {
+				errCode = telemetry.ErrCodeLicenseInvalid
+				sink.Emit(output.ErrorEvent{
+					Title: "Your license does not include the Snowflake emulator.",
+					Actions: []output.ErrorAction{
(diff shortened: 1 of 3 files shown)


internal/container/start.go
```diff
@@ -369,11 +369,23 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink,
 		sink.Emit(output.ContainerStatusEvent{Phase: "waiting", Container: c.Name})
 		healthURL := fmt.Sprintf("http://localhost:%s%s", c.Port, c.HealthPath)
 		if err := awaitStartup(ctx, rt, sink, containerID, "LocalStack", healthURL); err != nil {
+			errCode := telemetry.ErrCodeStartFailed
+			var licErr *licenseNotCoveredError
+			if errors.As(err, &licErr) && c.EmulatorType == config.EmulatorSnowflake {
+				errCode = telemetry.ErrCodeLicenseInvalid
+				sink.Emit(output.ErrorEvent{
+					Title: "Your license does not include the Snowflake emulator.",
```
gtsiolis, line 377: **praise**: Short and sweet.

internal/container/start.go
```diff
@@ -369,11 +369,23 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink,
...
+					Actions: []output.ErrorAction{
+						{Label: "Start a free Snowflake trial:", Value: "https://www.localstack.cloud/pricing?tab=snowflake"},
```
gtsiolis: **suggestion**: URL design for URLs like this are quite bad. Until we fix that, what do you think of using the dual (primary, secondary) pattern we have on other errors? This will also avoid trimming the URL on small width terminal windows which currently breaks the URL.

```diff
diff --git a/internal/container/start.go b/internal/container/start.go
index bdab74c..a9600f3 100644
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -376,7 +376,8 @@ func startContainers(ctx context.Context, rt runtime.Runtime, sink output.Sink,
                                sink.Emit(output.ErrorEvent{
                                        Title: "Your license does not include the Snowflake emulator.",
                                        Actions: []output.ErrorAction{
-                                               {Label: "Start a free Snowflake trial:", Value: "https://www.localstack.cloud/pricing?tab=snowflake"},
+                                               {Label: "Sign up for a free trial:", Value: "https://app.localstack.cloud/sign-up"},
+                                               {Label: "Contact our team:", Value: "https://www.localstack.cloud/demo"},
                                        },
                                })
                                err = output.NewSilentError(err)
```

anisaoshafi: Good point, thanks. Looks like this now:
Though I had a free trial but not for snowflake (freemium license), I mean the sign-up link is not entirely accurate or actionable.

image

gtsiolis: Correct, the sign up implies you need to a new account, which is unfortunatelly the case for now until we have better product packaging. Open to removing the sign up call to action, or merge as is. Your call.

gtsiolis (approved):

internal/container/start.go
anisaoshafi: Are you aware of any piece of documentation we have that explains the snowflake users need a new account? Wondering if we could link that instead, but I assume we don't have one 🙏🏼

gtsiolis: I don't think we have one, and the goal is to not require a separate account soon. I'd think it's ok to merge as is for now.

anisaoshafi: thanks for checking George 🙌🏼

carole-lavillonniere (approved):

PR #231, opened 2026-05-07
Improve OTel tracing for analytics and aws cli

carole-lavillonniere:
## Motivation

Traces emitted under `LSTK_OTEL=1` were not fine-grained enough around the `lstk aws` wrapper. Investigation revealed several blind spots: the telemetry call, the Docker daemon-ID lookup used to derive `machine_id`, and the `aws` CLI subprocess itself were either uninstrumented or producing orphaned/dropped spans.

## Changes

- Instrument the telemetry HTTP client with `otelhttp.NewTransport`
- Instrument the Docker client used by `LoadOrCreateMachineID`
- Wrap `awscli.Exec` in a span (`aws cli`)
- Reorder defers in `cmd/root.go:Execute` so `tel.Close()` runs before the tracer-provider shutdown (otherwise the analytics POST span is created on a noop tracer after shutdown and never reaches the exporter)
- Make machine-ID resolution lazy (in `GetEnvironment(ctx)`) so the Docker `info` / `_ping` spans are parented to the active command span instead of orphaned at process init.

toward DRG-809

## Result of running `lstk aws  sts get-caller-identity`:

image

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -84,8 +84,6 @@ func NewRootCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.C
 
 func Execute(ctx context.Context) error {
 	cfg := env.Init()
-	tel := telemetry.New(cfg.AnalyticsEndpoint, cfg.DisableEvents)
-	defer tel.Close()
 
 	logger, cleanup, err := newLogger()
 	if err != nil {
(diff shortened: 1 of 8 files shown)


anisaoshafi (approved):
LGTM 👏🏼

anisaoshafi:
Question: there is a difference in time for `lstk aws  sts get-caller-identity` from before to after `700 ms -> 1.98 s`. Is that perhaps related to the improvement itself or just random?

carole-lavillonniere:
> Question: there is a difference in time for `lstk aws sts get-caller-identity` from before to after `700 ms -> 1.98 s`. Is that perhaps related to the improvement itself or just random?

Good observation! I did some research and testing, and figured I was pointing to aws cli v1 which is in Python (v2 is in go hence a single binary). aws cli v1 is much slower and needs to do I/O when cache is not warm. The slowness does not seem to be due to anything added in this PR.

PR #232, opened 2026-05-07
Detect externally-named containers in AWS proxy commmnds

gtsiolis:
`lstk aws ...` reported `LocalStack AWS Emulator is not running` whenever LocalStack was started outside lstk under a non-default container name (e.g. via the `docker compose` template, where the container is named `localstack-main`). Other commands — `lstk` (start), `lstk status`, `lstk stop` — already detect the running container in this case, so the inconsistency was confusing.

Root cause: `cmd/aws.go` only checked Docker for the configured container name (`rt.IsRunning(awsContainer.Name())`) instead of falling back to image-based discovery the way the other commands do via `resolveRunningContainerName`.

Closes DES-220

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-05-07 at 14 03 04 | Screenshot 2026-05-07 at 14 02 49 |

### DIFF
--- a/cmd/aws.go
+++ b/cmd/aws.go
@@ -8,6 +8,7 @@ import (
 	"github.com/localstack/lstk/internal/awscli"
 	"github.com/localstack/lstk/internal/awsconfig"
 	"github.com/localstack/lstk/internal/config"
+	"github.com/localstack/lstk/internal/container"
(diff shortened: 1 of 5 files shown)


anisaoshafi (approved):
Good catch, thanks for the prompt fix 👏🏼

PR #233, opened 2026-05-07
Add spans for ResolveHost and ProfileExists

carole-lavillonniere:
## Motivation

Trace of `lstk aws` shows ~100ms gap between Docker container check and `aws cli` span start. See https://github.com/localstack/lstk/pull/231
Untraced work in that window: DNS lookup in `endpoint.ResolveHost` plus `~/.aws/config` reads in `awsconfig.ProfileExists`. Adding spans makes the gap visible.

Towards DRG-809

image


anisaoshafi (approved):
Good job! already using otel to investigate performance issues🏅

PR #235, opened 2026-05-07
Set dependabot gomod versioning strategy to increase-if-necessary

anisaoshafi:
Dependabot's default behavior for Go modules skips major version bumps, so security fixes that land in a new major release (e.g. a CVE fixed in v29.x when we're on v28.x) never generate a PR.

Added `versioning-strategy: increase-if-necessary` to the `gomod` ecosystem in `.github/dependabot.yml`

ℹ️ Alternatively we can use `semver-major`

Closes DRG-826

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,7 @@ updates:
       interval: "daily"
       time: "09:00"
       timezone: "Etc/UTC"
+    versioning-strategy: increase-if-necessary
     open-pull-requests-limit: 10
     labels:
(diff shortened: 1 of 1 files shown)


carole-lavillonniere:
> Dependabot's default behavior for Go modules skips major version bumps

@anisaoshafi I am not convinced about and I'm afraid this PR does not change the behavior but the issue lies somewhere else. We could try migrating to `github.com/moby/moby` but bump the version manually in the meantime.

PR #236, opened 2026-05-08
Hide header on start when Docker is not detected

gtsiolis:
Running lstk start without a Docker daemon shows the header with the "LocalStack" loading-dots animation frozen at one or two dots, then the unavailable-Docker error:
```
     ▟████▖   lstk (0.7.1)
    ▟██▙█▙█▟  LocalStack ..
      ▀▛▀▛▀   ~/.config/lstk/config.toml
```

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-05-08 at 15 06 42 |  Screenshot 2026-05-08 at 15 06 30| 

Fix DES-223

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -204,6 +204,10 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 		return a, nil
 	case output.ErrorEvent:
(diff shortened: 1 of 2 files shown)


anisaoshafi (approved):
Valid change, looks good 💅🏼

PR #240, opened 2026-05-11
Update docker compose using Jaeger v2

joe4dev:
## Motivation

The previously used `jaegertracing/all-in-one:latest` image (Jaeger v1) is deprecated and now fails to start, so `make otel` is broken.
Screenshot 2026-05-08 at 11 57 43

## Changes

Upgrades `docker-compose.tracing.yaml` to Jaeger v2 (`jaegertracing/jaeger:latest`).

Screenshot 2026-05-08 at 12 10 25

- Switches to the Jaeger v2 image with Badger embedded storage (7-day span TTL)
- Adds a named Docker volume (`jaeger-data`) so traces persist across container restarts
- Adds `restart: unless-stopped` and an explicit `container_name`

### DIFF
--- a/docker-compose.tracing.yaml
+++ b/docker-compose.tracing.yaml
@@ -1,17 +1,37 @@
-# Starts Jaeger for local trace collection and visualization.
+# Starts Jaeger v2 (all-in-one) for local trace collection and visualization.
 # UI:        http://localhost:16686
-# OTLP HTTP: http://localhost:4318   (used by lstk)
 # OTLP gRPC: localhost:4317
+# OTLP HTTP: http://localhost:4318
+# GitHub: https://github.com/jaegertracing/jaeger
 #
 # Usage:
-#   docker compose -f docker-compose.tracing.yaml up -d
+#   docker compose -f otel/jaeger/docker-compose.yaml up -d
+#
+# Traces are persisted in the jaeger-data Docker volume via Badger embedded storage.
+# The volume is mounted at /tmp so UID 10001 (the Jaeger process user) can write to it.
+# Badger data is stored at /tmp/badger inside the volume.
+
+name: jaeger-standalone
 
 services:
   jaeger:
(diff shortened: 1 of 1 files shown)


joe4dev:
I'd love some feedback on https://github.com/localstack/localstack-pro/pull/6934, introducing `OTEL=1` in LocalStack together with this fix 🙏 . You can check out `feature/otel-flag` and test Docker mode together with `lstk`.


FYI: I provide three tracing tool alternatives in this PR https://github.com/localstack/appinspector-workspace/pull/5

docker-compose.tracing.yaml
```diff
@@ -1,17 +1,37 @@
...
+# Starts Jaeger v2 (all-in-one) for local trace collection and visualization.
 # UI:        http://localhost:16686
-# OTLP HTTP: http://localhost:4318   (used by lstk)
 # OTLP gRPC: localhost:4317
+# OTLP HTTP: http://localhost:4318   (used by lstk)
+# GitHub: https://github.com/jaegertracing/jaeger
 #
 # Usage:
-#   docker compose -f docker-compose.tracing.yaml up -d
+#   docker compose -f otel/jaeger/docker-compose.yaml up -d
```
carole-lavillonniere: issue: path to yaml is wrong

carole-lavillonniere (approved):
Thanks @joe4dev  for the fix and improvement 👏 
I just added one minor comment.

I am thinking we should also handle `LOCALSTACK_OTEL` and treat it as an alias to `LSTK_OTEL` since it is the name of the env var the emulator is using, so it can ultimately work across all LS products when set in the shell (not in the scope of your PR)

Will test together with your PR soon.

docker-compose.tracing.yaml
joe4dev: thanks for pointing this out. Fixed in ae9f436

PR #243, opened 2026-05-12
Make status command output easier to read

gtsiolis:
| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-05-13 at 00 31 29 | Screenshot 2026-05-13 at 00 31 18 |

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -296,9 +296,14 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return a, nil
 	case output.InstanceInfoEvent:
 		if line, ok := output.FormatEventLine(msg); ok {
-			line = strings.Replace(line, output.SuccessMarker(), styles.Success.Render(output.SuccessMarker()), 1)
-			for _, part := range strings.Split(line, "\n") {
-				a.addLine(styledLine{text: part})
+			parts := strings.Split(line, "\n")
+			for i, part := range parts {
+				if i == 0 {
+					part = strings.Replace(part, output.SuccessMarker(), styles.Success.Render(output.SuccessMarker()), 1)
+					a.addLine(styledLine{text: part})
+				} else {
+					a.addLine(styledLine{text: part, secondary: true})
+				}
 			}
 		}
 		return a, nil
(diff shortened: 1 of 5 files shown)


internal/ui/app.go
```diff
@@ -295,11 +295,18 @@ func (a App) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			a.addLine(styledLine{text: "• Endpoint: " + msg.Host, secondary: true})
+		}
+		if msg.ContainerName != "" {
+			a.addLine(styledLine{text: "• Container: " + msg.ContainerName, secondary: true})
+		}
+		if msg.Version != "" {
+			a.addLine(styledLine{text: "• Version: " + msg.Version, secondary: true})
+		}
+		if msg.Uptime > 0 {
+			a.addLine(styledLine{text: "• Uptime: " + output.FormatUptime(msg.Uptime), secondary: true})
```
anisaoshafi: This code is mostly duplicate across `app.go` and `plain_format.go`. If we add more properties in the future, we risk diverging.
Could you try to bring back the `FormatEventLine` and make sure the field list lives in exactly one place, not sure if sth like this would work?
```suggestion
               if line, ok := output.FormatEventLine(msg); ok {
                       parts := strings.Split(line, "\n")
                       for i, part := range parts {
                               if i == 0 {
                                       part = strings.Replace(part, output.SuccessMarker(), styles.Success.Render(output.SuccessMarker()), 1)
                                       a.addLine(styledLine{text: part})
                               } else {
                                       a.addLine(styledLine{text: part, secondary: true})
                               }
                       }

```

anisaoshafi (commented):
This looks much neater ✨ thank you!

internal/ui/app.go
gtsiolis: Added in ~~9a244eb~~, 98141b5c7869ba2bdf4692ad236548a943e4848f.

gtsiolis: There's a risk with this change as it creates an implicit contract between `formatInstanceInfo` and the TUI: the TUI assumes line 1 of the formatted output is the header with the success marker and lines 2+ are secondary key:value bullets. Checking if this breaks anything right now... 👀

anisaoshafi: still works well for me 🥳 
image

anisaoshafi (approved):
:shipit:

internal/ui/app.go
gtsiolis: Thanks for checking, merging!

PR #244, opened 2026-05-13
Surface persistence in start and status commands

gtsiolis:
Adds a `Persistence: Enabled` bullet to the start info block and the status output when the running instance has `LOCALSTACK_PERSISTENCE=1` set — either via `--persist` or by inheriting the env var from the host shell. restart benefits transitively since it dispatches through start.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-05-13 at 16 39 33 | Frame 852839935 |
| Screenshot 2026-05-13 at 16 39 38 | Frame 852839936 |

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -244,7 +244,7 @@ func emitPostStartPointers(sink output.Sink, emulatorType config.EmulatorType, r
 	} else {
 		sink.Emit(output.MessageEvent{Severity: output.SeveritySecondary, Text: fmt.Sprintf("• Endpoint: %s", resolvedHost)})
 	}
-	if persist {
+	if persist && emulatorType == config.EmulatorAWS {
 		sink.Emit(output.MessageEvent{Severity: output.SeveritySecondary, Text: "• Persistence: Enabled"})
 	}
 	if webAppURL != "" {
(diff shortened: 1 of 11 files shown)


carole-lavillonniere:
@gtsiolis can you open separate PRs for the other changes? There seem to be some changes about the displayed host and some snowflake header changes? Not sure. It's a little hard to read at the moment.

gtsiolis:
Makes sense. I had a feeling I would have to split this one, will split later. ⏳

gtsiolis:
Updated PR to include only the Persistence metadata in the header without additional cleanup, plus show the metadata only for AWS emulation. Curious what you think!

anisaoshafi (commented):
Thanks George, this is useful 🙌🏼 

PR description mentions this should work when `LOCALSTACK_PERSISTENCE=1` is set in shell, but doesn't seem to.

Same when `LOCALSTACK_PERSISTENCE` is set in the config profile.

Image

gtsiolis:
Great catch, @anisaoshafi! Added 082c7e1 to make this work also with the inline environment variable. Thoughts?

anisaoshafi (approved):
LGTM, thanks for addressing the comments 🚀

PR #245, opened 2026-05-14
Hash auth token in telemetry events

gtsiolis:
The auth token was previously sent in the `auth_token_id` field of every telemetry event. We don't need the raw value to correlate events per user, but a stable fingerprint is enough. Switching to a hash keeps the credential out of telemetry payloads while preserving the analytics signal.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -114,7 +114,7 @@ func Execute(ctx context.Context) error {
 
 	logger.Info("lstk %s starting", version.Version())
 
-	// Resolve auth token for telemetry: keyring first, then env var.
+	// Resolve auth token for telemetry correlation: keyring first, then env var.
 	resolvedToken := cfg.AuthToken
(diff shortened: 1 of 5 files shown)


anisaoshafi (commented):
Thanks for raising this @gtsiolis 🔐 
I wanted to mitigate this similarly, but my understanding is that it breaks the data contract
I'll defer to @thrau for his data expertise 🙏🏼

thrau (commented):
in principle sending a stable hash instead of the actual key would be a great idea! unfortunately, all our data pipelines that join usage data with user and subscription data currently rely on the actual auth token.

so tthis change would require a significant refactor of our data pipelines. i don't have a complete picture of what this would mean right now. is there anything i might be missing?

gtsiolis:
Thanks, @thrau! Closing for now, we may get to it later if needed.

PR #248, opened 2026-05-15
Add DOCKER_FLAGS support for passing extra flags to the emulator container

anisaoshafi:
## Problem

lstk had no way to pass arbitrary Docker flags when starting an emulator if you needed to inject environment variables (-e SERVICES=s3,sqs), or extra volume mounts.

The root cause was structural: `runtime.ContainerConfig` only modeled the options lstk explicitly needed (port bindings, known env vars, the volume mount for persistence). 

 ## Fix
Added support for a `DOCKER_FLAGS` environment variable and a docker_flags field in the `[[containers]] `config.toml block. 
Both accept a raw docker-run-style string (e.g. "-e SERVICES=s3,sqs -v /tmp/data:/data"). 
The two sources are merged: env var flags apply globally across all containers, config flags apply per-container.

Supported flags: `-e/--env, -v/--volume`.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -150,6 +150,7 @@ func buildStartOptions(cfg *env.Env, appConfig *config.Config, logger log.Logger
 		LocalStackHost:   cfg.LocalStackHost,
 		Containers:       appConfig.Containers,
 		Env:              appConfig.Env,
+		DockerFlags:      cfg.DockerFlags,
 		Persist:          persist,
 		Logger:           logger,
 		Telemetry:        tel,
--- a/internal/config/containers.go
+++ b/internal/config/containers.go
@@ -94,6 +94,9 @@ type ContainerConfig struct {
 	Volume string       `mapstructure:"volume"`
 	// Env is a list of named environment references defined in the top-level [env.*] config sections.
 	Env []string `mapstructure:"env"`
+	// DockerFlags is a raw docker-run-style flag string (e.g. "-e FOO=bar -v /tmp:/data").
+	// Supported flags: -e/--env, -v/--volume.
+	DockerFlags string `mapstructure:"docker_flags"`
 }
 
 // VolumeDir returns the host directory to mount into the container for persistence/caching.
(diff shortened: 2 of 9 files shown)


carole-lavillonniere:
@anisaoshafi I don't fully get why we need this new feature.

- ` -e SERVICES=s3,sqs`: this should be possible using the current config format:

```

  [env.aws_services]
  SERVICES = "s3,sqs"

  [[containers]]
  type = "aws"
  port = "4566"
  env = ["aws_services"]
```

-  extra volume mounts: not supported by config yet, but it could:
```
[[containers]]
type = "aws"
volume = "/tmp/data:/data"
```

could become:

```
[[containers]]
type = "aws"
volumes = ["/tmp/data:/data", "/home/me/init:/etc/localstack/init/ready.d:ro"]
```

Am I missing something?

anisaoshafi:
@carole-lavillonniere that's true, there are workarounds for env vars, though not directly for volumes.

In the [docs](https://docs.localstack.cloud/aws/capabilities/config/configuration/#docker), we support customer `DOCKER_FLAGS`, and Maureen was expecting [this](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1778776181506749?thread_ts=1778693596.466979&cid=C0A1857LGTU) to work out of the box since it works with the old CLI. I took the liberty to support this, but probably should have asked for more input.
@mmaureenliu what do you think?

mmaureenliu:
> @carole-lavillonniere that's true, there are workarounds for env vars, though not directly for volumes.
> 
> In the [docs](https://docs.localstack.cloud/aws/capabilities/config/configuration/#docker), we support customer `DOCKER_FLAGS`, and Maureen was expecting [this](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1778776181506749?thread_ts=1778693596.466979&cid=C0A1857LGTU) to work out of the box since it works with the old CLI. I took the liberty to support this, but probably should have asked for more input. @mmaureenliu what do you think?

I like Carole's suggestion if it works. My initial intuition was also to look at the volume config but then realised it was only used for persistence.

It would also be much cleaner than DOCKER_FLAGS - current `localstack` implementation apparently struggles with space in path

anisaoshafi:
closing for now, will revisit at another point

PR #251, opened 2026-05-19
`lstk az` Azure CLI proxy

carole-lavillonniere:
## Motivation

Bring Azure CLI integration into lstk so users can run `az` commands against the LocalStack Azure emulator. Uses the **wrap** model (isolated `AZURE_CONFIG_DIR`), not `start-interception` global mutation — it mirrors `lstk aws` and leaves the user's global `~/.azure` untouched.

Depends on #262 (`Support starting the Azure emulator`) to actually launch the emulator via `lstk start`.

## Approach

The Azure CLI has no `--endpoint-url`/`--profile`; the only isolation knob is `AZURE_CONFIG_DIR`. Inside that isolated dir we register a custom Azure cloud (`LocalStack`) whose endpoints point at the LocalStack Azure emulator, activate it, disable instance discovery, and perform a one-time dummy service-principal login:

- `lstk setup azure` runs (against the isolated `AZURE_CONFIG_DIR`):
  - `az cloud register|update --name LocalStack --cloud-config '{…}'`
  - `az cloud set --name LocalStack`
  - `az config set core.instance_discovery=false`
  - `az login --service-principal -u any-app -p any-pass --tenant anytenant`
- `lstk az <args>` runs `az <args>` with just `AZURE_CONFIG_DIR` set — no HTTP(S) forward proxy. `az` talks directly to LocalStack for Azure service URLs and to the real internet for everything else (extension downloads, AAD static content, etc.).

Closes FLC-656

 ## Error when not setup first
az-setup-first

## Setup
az-setup

## Proxy
az-proxy

## Not affecting global az config
az-not-global

az-real

## Trace
image

### DIFF
--- a/cmd/az.go
+++ b/cmd/az.go
@@ -97,8 +97,8 @@ Examples:
 				sink.Emit(output.ErrorEvent{
 					Title: "DNS resolution required for 'lstk az'",
 					Actions: []output.ErrorAction{
-						{Label: "Note:", Value: endpoint.DNSRebindNote},
-						{Label: "Why:", Value: "LocalStack routes Azure requests by Host header"},
+						{Label: "Note:", Value: "Could not resolve *." + endpoint.Hostname + " to 127.0.0.1."},
+						{Label: "Why:", Value: "the Azure emulator serves endpoints under *." + endpoint.Hostname + ", which the Azure CLI must be able to resolve"},
 						{Label: "Fix:", Value: "configure DNS or set LOCALSTACK_HOST"},
 					},
 				})
--- a/cmd/az.go
+++ b/cmd/az.go
@@ -1,0 +1,122 @@
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"time"
+
+	"github.com/localstack/lstk/internal/azurecli"
+	"github.com/localstack/lstk/internal/azureconfig"
+	"github.com/localstack/lstk/internal/config"
+	"github.com/localstack/lstk/internal/container"
+	"github.com/localstack/lstk/internal/endpoint"
+	"github.com/localstack/lstk/internal/env"
+	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/runtime"
+	"github.com/localstack/lstk/internal/terminal"
+	"github.com/spf13/cobra"
+)
+
+func newAzCmd(cfg *env.Env) *cobra.Command {
+	return &cobra.Command{
+		Use:   "az [args...]",
+		Short: "Run Azure CLI commands against LocalStack",
+		Long: `Run Azure CLI commands against the LocalStack Azure emulator.
+
+Runs 'az <args>' with an isolated AZURE_CONFIG_DIR in which a custom Azure cloud
+is registered against LocalStack's endpoints, so your global ~/.azure
+configuration is left untouched and plain 'az' commands keep talking to real
+Azure.
+
+Run 'lstk setup azure' once before using this command.
+
+Examples:
+  lstk az group list
+  lstk az storage account list`,
+		DisableFlagParsing: true,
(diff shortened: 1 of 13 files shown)


cmd/az.go
```diff
@@ -0,0 +1,122 @@
...
+				return output.NewSilentError(fmt.Errorf("%s is not running", azureContainer.Name()))
+			}
+
+			_, dnsOK := endpoint.ResolveHost(cmd.Context(), azureContainer.Port, cfg.LocalStackHost)
+			if !dnsOK {
+				sink.Emit(output.ErrorEvent{
+					Title: "DNS resolution required for 'lstk az'",
+					Actions: []output.ErrorAction{
+						{Label: "Note:", Value: endpoint.DNSRebindNote},
+						{Label: "Why:", Value: "LocalStack routes Azure requests by Host header"},
```
bblommers: **nitpick**: This does not seem quite correct - should that be something like `Emulator not reachable on 127.0.0.1`?

cmd/setup.go
```diff
@@ -39,3 +40,29 @@ func newSetupAWSCmd(cfg *env.Env) *cobra.Command {
...
+		PreRunE: initConfig(nil),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			appConfig, err := config.Get()
+			if err != nil {
+				return fmt.Errorf("failed to get config: %w", err)
+			}
+
+			if !isInteractiveMode(cfg) {
+				return fmt.Errorf("setup azure requires an interactive terminal")
+			}
```
bblommers: **question**: I'm not fluent enough in Go to figure this out out - which parts of the setup are interactive?

Is it possible to also make it non-interactive (even in a second pass)? Ideally this can be used in the CI as well

bblommers (approved):
Awesome - thanks @carole-lavillonniere! 

Note that I have not actually tested this out, but the Azure-specific parts LGTM 🙏

cmd/az.go
carole-lavillonniere: Good point, addressed here f986533. WDYT?

cmd/setup.go
carole-lavillonniere: You're totally right! This must also work in non-interactive mode. Fixed here 4904457
Thanks dear human for your valuable review.

cmd/az.go
bblommers: Makes sense! Thanks for addressing this 👍

cmd/setup.go
bblommers: 🙏

internal/azurecli/exec.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// ErrNotInstalled is returned when the `az` binary cannot be found on PATH.
+var ErrNotInstalled = errors.New("az CLI not found in PATH — install it from https://learn.microsoft.com/cli/azure/install-azure-cli")
+
+// CheckInstalled returns ErrNotInstalled if the `az` binary is not on PATH.
+// Callers should use this before performing setup work to avoid leaving partial state.
+func CheckInstalled() error {
+	if _, err := exec.LookPath("az"); err != nil {
+		return ErrNotInstalled
```
anisaoshafi, line 24: small nit: when az cli is not installed the error is a bit redundant with warning and error

Image

anisaoshafi (approved):
✨

anisaoshafi:
@bblommers Not sure if it's a known issue, but when trying az login, I get this error which believe it's on the emulator or az cli config:

lstk:
image
emulator logs:
image

internal/azurecli/exec.go
carole-lavillonniere: image

Thank you! Fixed here 0cdf9e0

bblommers:
Thanks for letting me know @anisaoshafi! Will look into that ASAP :eyes:

PR #256, opened 2026-05-20
Validate image tag format in config and improve error message

anisaoshafi:
### Motivation

We don't validate the version tag in config.toml so users discover later in the flow some tags are not possible.
Also the error message was not pointing to the real issue.
  
### Changes  
- Reject tags that don't conform to Docker's tag format rules at config load (invalid chars, leading ./-, >128 chars). Fails fast before any network or Docker operation.  
- Normalize tags with zero-padded calendar months (2026.04 → suggest 2026.4) before license validation
- The suggested valid tag in error messages is computed from the previous calendar month, so it's a reasonable recent version.
- License API error improvement: when the license API returns 400 with a `licensing.license.format` body, the error now says "unsupported image tag" instead of the generic token/subscription message, which is more actionable for the user.

a) _Before // inaccurate message_ 
image

b) _Now: tag="3.99.99" // valid tag regex, but does't exist: license server complains_
image

c) _Now: tag="2026.04" // valid tag, previously failed with error a)_
image

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -573,8 +573,13 @@ func validateLicense(ctx context.Context, sink output.Sink, opts StartOptions, c
 	if err != nil {
 		sink.Emit(output.SpinnerStop())
 		var licErr *api.LicenseError
-		if errors.As(err, &licErr) && licErr.Detail != "" {
-			opts.Logger.Error("license server response (HTTP %d): %s", licErr.Status, licErr.Detail)
+		if errors.As(err, &licErr) {
+			if licErr.Detail != "" {
+				opts.Logger.Error("license server response (HTTP %d): %s", licErr.Status, licErr.Detail)
+			}
+			if licErr.IsUnsupportedTag {
+				licErr.Message = config.UnsupportedTagMessage()
+			}
 		}
 		opts.Telemetry.EmitEmulatorLifecycleEvent(ctx, telemetry.LifecycleEvent{
 			EventType: telemetry.LifecycleStartError,
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -557,7 +557,7 @@ func validateLicense(ctx context.Context, sink output.Sink, opts StartOptions, c
 	licenseReq := &api.LicenseRequest{
 		Product: api.ProductInfo{
 			Name:    containerConfig.ProductName,
-			Version: version,
+			Version: config.NormalizeTag(version),
 		},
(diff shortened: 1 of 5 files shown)


gtsiolis:
Cross-posting [relevant decision](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1779284662874089?thread_ts=1779201043.297249&cid=C0A1857LGTU) to redirect this to support zero padding.

anisaoshafi:
> Cross-posting [relevant decision](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1779284662874089?thread_ts=1779201043.297249&cid=C0A1857LGTU) to redirect this to support zero padding.

Adjusted according to discussion [[1](https://github.com/localstack/lstk/pull/256/commits/64158268d4dcb88a9cadf8ba4aa57e032b372692)] 🙏🏼

gtsiolis (approved):
UX LGTM, works as expected. 💯

internal/container/start.go
```diff
@@ -570,8 +570,13 @@ func validateLicense(ctx context.Context, sink output.Sink, opts StartOptions, c
...
 	if err != nil {
 		var licErr *api.LicenseError
-		if errors.As(err, &licErr) && licErr.Detail != "" {
-			opts.Logger.Error("license server response (HTTP %d): %s", licErr.Status, licErr.Detail)
+		if errors.As(err, &licErr) {
+			if licErr.Detail != "" {
+				opts.Logger.Error("license server response (HTTP %d): %s", licErr.Status, licErr.Detail)
+			}
+			if licErr.IsUnsupportedTag {
+				licErr.Message = config.UnsupportedTagMessage()
```
carole-lavillonniere: issue: it's unusual to mutate an error message. Can we wrap the error instead?

carole-lavillonniere (approved):
👏 good one! Added just one small concern, good to go after it's addressed

internal/container/start.go
anisaoshafi: went with this approach to avoid a future circular dependancy: https://github.com/localstack/lstk/pull/256/commits/0830aa97401e2c3f75760c107c79f8d7d6594174. will address now. Thanks for checking :pray:

anisaoshafi: Addressed here: https://github.com/localstack/lstk/pull/256/changes/463b35854c08f7b36f8b7a534d53f547eeef5108

PR #261, opened 2026-05-26
Unit tests do not use real $HOME 

carole-lavillonniere:
## Motivation

Under Nix's sandboxed build, `$HOME` is set to `/homeless-shelter` (a deliberately non-existent placeholder), so `TestParseDestination` fails because it calls `os.UserHomeDir()` which does not exist.

## Changes

- Add a `home` parameter to `snapshot.ParseDestination` 
- Switch `TestParseDestination` to use `t.TempDir()` as home so it is hermetic
- Add `TestParseDestinationTildeWithoutHome` to test the Nix scenario

## Tests

- `make test` — all 520 unit tests pass.
- Verified the Nix-sandbox scenario locally: `HOME=/homeless-shelter go test ./internal/snapshot/...` passes (and failed before this change).

Closes FEED-67

### DIFF
--- a/cmd/snapshot.go
+++ b/cmd/snapshot.go
@@ -48,7 +48,8 @@ To save to a remote pod on the LocalStack platform, use the pod: prefix:
 				destArg = args[0]
 			}
 
-			dest, err := snapshot.ParseDestination(destArg, time.Now())
+			home, _ := os.UserHomeDir()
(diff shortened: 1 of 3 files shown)


anisaoshafi (approved):

purcell:
Nice, thanks!

PR #262, opened 2026-05-27
Support starting the Azure emulator

carole-lavillonniere:
## Motivation

Minimal implementation so that LocalStack Azure emulator (`localstack/localstack-azure-alpha`) can be started.

## Changes

- Register the Azure emulator in `internal/config`: `knownImages` (`localstack/localstack-azure`), `emulatorHealthPaths`, and `ContainerPort` — so `Image()`/`ProductName()`/`HealthPath()`/`ContainerPort()` resolve for Azure.
- Add Azure to `SelectableEmulatorTypes` so it appears in the first-run emulator picker. Replaced the first-character `SelectionKey()` with an explicit unique-key map (`a`/`s`/`z`) — `aws` and `azure` both start with "a", so without this, selecting Azure resolved to AWS.
- Skip lstk's pre-flight platform license check for Azure, mirroring Snowflake. Introduced `EmulatorType.SelfValidatesLicense()` (Snowflake + Azure) and used it in `start.go` (pre-pull and post-pull validation) and `label.go`. The Azure-alpha container activates its own license against the licensing server; lstk's platform `GetLicense` as no catalog entry for it and would otherwise abort the start.

1. Select
select

2. Started
az-started

3. Status
az-status

4. Stop
az-stop

## TODO before merging
- [x] Change the image name to `localstack-azure`, see SMF-642

Towards FLC-679


anisaoshafi (approved):
🚀

PR #265, opened 2026-05-27
Remove license catalog check for version resolution

anisaoshafi:
### Motivation
The catalog endpoint has gone stale multiple times and caused outdated versions of localstack to be downloaded. 
Now that "latest" refers to the stable monthly release which means pulls from latest are less frequent, we can get rid of the catalog lookup.

## Changes
- "latest" images now always use post-pull validation. lstk pulls the image, reads `LOCALSTACK_BUILD_VERSION` from the image config, and uses that version for the license request. 
- Pinned tags continue to validate pre-pull unchanged. 
- UI header label populated after startup rather than via a concurrent catalog call.


carole-lavillonniere (approved):
LGTM! 
One small thing: Claude tells me that "a logged-in user on first lstk run against an externally-started container would see the TUI hang". Maybe worth adding an integration test case (and fixing in case that's true).

PR #270, opened 2026-05-28
Prompt for login before emulator selection on first run

gtsiolis:
On a fresh install with no stored credentials, `lstk` currently shows the emulator picker first and only then starts the login flow. That order is backwards:
- Login is a hard gate. Until the user authenticates, the picked emulator can't actually start, so any input we collect before login is friction the user may end up throwing away if they bail at the auth step.
- The natural mental model is identify-then-configure: "log me in, then ask me what I want to run."

This flips the order so first-run users see the login screen first, then the emulator picker.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-05-28 at 13 44 31 | Screenshot 2026-05-28 at 13 43 13 | 

Cc @anisaoshafi @carole-lavillonniere because https://github.com/localstack/lstk/pull/217


internal/ui/run_test.go
```diff
@@ -0,0 +1,81 @@
...
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	t.Setenv("XDG_CONFIG_HOME", filepath.Join(home, ".config"))
+
+	configPath := filepath.Join(home, "config.toml")
+	require.NoError(t, os.WriteFile(configPath, []byte(`[[containers]]
+type = "aws"
+port = "4566"
+`), 0644))
+	require.NoError(t, config.InitFromPath(configPath))
```
anisaoshafi: Perhaps this could lead to a test interference because `config.InitFromPath` writes to a global Viper instance if another test runs in parallel and also calls config.InitFromPath or viper.Reset().

Could you try set config path in `_, err := container.SelectEmulator(ctx, sink, configPath)` instead?

internal/ui/run_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestResolveAuthThenSelectEmulator_OrdersAuthFirst verifies that on first run
+// with no token, the login prompt fires before the emulator-selection prompt.
+// This is the property that lets users authenticate before configuring an
+// emulator instead of the other way around.
+func TestResolveAuthThenSelectEmulator_OrdersAuthFirst(t *testing.T) {
```
anisaoshafi: We are already setting `aws` in the config file (Line 36). Then we trigger `container.SelectEmulator(ctx, sink, "")` without configuration. Was that intended?
The order of login before emulator selection is always ensured in the test, even if we change the order again in `internal/ui/run.go` so test wouldn't catch regressions.
Can you either rename the test to the scope you're trying to test or avoid writing the config file beforehand so it's more like an integration test?

anisaoshafi (commented):
Good catch 🎣  George, this is a better approach!
Only have a concern about the test.

internal/ui/run_test.go
```diff
@@ -0,0 +1,81 @@
...
+	require.NoError(t, err)
+
+	mu.Lock()
+	defer mu.Unlock()
+	require.GreaterOrEqual(t, len(prompts), 2, "expected at least the auth and emulator prompts")
+	assert.True(t, strings.Contains(strings.ToLower(prompts[0]), "authorization"),
+		"first prompt should be the login flow, got %q", prompts[0])
+	assert.True(t, strings.Contains(strings.ToLower(prompts[1]), "emulator"),
+		"second prompt should be emulator selection, got %q", prompts[1])
+}
```
carole-lavillonniere: Could we test this with an integration test instead? Claude's being stubborn at favoring unit over integration tests.

gtsiolis:
Thanks both, resolved comments in 7000c71. Please, take another look when you're free! 🍬 🍬

anisaoshafi (approved):

anisaoshafi:
<s>ℹ️ Some tests are failing intermittently, likely due to planned maintenance on https://www.dockerstatus.com/ </s> All green now after few retries ✅

PR #272, opened 2026-05-28
Implement --dry-run flag for snapshot load

anisaoshafi:
Added `--dry-run` to lstk snapshot load. 
When passed with a pod ref, it calls `GET /_localstack/pods/{name}/diff` instead of loading, and shows a per-service preview of what would be added or modified, without touching running state. 

Local snapshot refs return an error since the emulator has no diff endpoint for them. 

The `--merge` flag is also accepted to contextualize the output.

| event | output |
| loading | Screenshot 2026-05-28 at 19 41 21 |
| success | image| 
| error | image |

ℹ️ Deletions are not tracked, because it's not implemented in cloud pods' backend (in emulator).
ℹ️  During `lstk snapshot load --dry-run` (pod refs only), `--merge` is ignored because `GET /_localstack/pods/{name}/diff` endpoint which is computed by the emulator doesn't support  a strategy param, so it has zero effect on the diff computation itself.


skyrpex (approved):
Looks good to me but I wonder: would it make sense to have a `lstk snapshot diff` command instead? 🤔

anisaoshafi:
> Looks good to me but I wonder: would it make sense to have a `lstk snapshot diff` command instead? 🤔

Thanks @skyrpex 🙌🏼  
@gtsiolis what do you think about an `lstk snapshot diff` alias that does the same? Could it be confused with sth different? Thanks 🙏🏼

anisaoshafi:
Merging for now, open to later feedback.

PR #273, opened 2026-05-29
Honor --non-interactive in proxy commands

gtsiolis:
## Motivation

Non-interactive consumers (CI, scripts, agents) need clean, deterministic output. `lstk aws` ignored `--non-interactive` — and because the command uses `DisableFlagParsing`, the flag was also forwarded to the `aws` binary, which rejects it as an unknown option.

## Changes

- Strip and honor `--non-interactive` in proxy commands so the spinner is suppressed and the flag isn't passed through to `aws`.

## Tests

- Integration `TestAWSCommandSuppressesSpinnerInNonInteractiveMode` (PTY; fails before, passes after) — existing slow/fast spinner tests still pass.
- Unit `TestStripNonInteractiveFlag`.

DES-240

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #274, opened 2026-06-01
Honor persistent flags in proxy commands

gtsiolis:
Generalizes #273. Proxy commands set `DisableFlagParsing`, so lstk's persistent flags were forwarded to the wrapped binary (rejected as unknown options) and their effect silently dropped. `--config` is broken on `lstk aws` the same way `--non-interactive` was — it's ignored *and* passed through to `aws`.

## Changes

- Add `stripGlobalFlags`, which extracts `--non-interactive` and `--config` (both `--flag value` and `--flag=value`, in any position) from a proxy command's args. Reusable by future proxy commands; replaces the one-off `stripNonInteractiveFlag`.
- `lstk aws` now sets `cfg.NonInteractive` so interactivity is consistent across the codebase, and routes `--config` back through Cobra so `initConfig` resolves the overridden path instead of forwarding it to `aws`.

## Tests

- Unit `TestStripGlobalFlags` (both flags, both forms, interleaving, edge cases).
- Integration `TestAWSCommandStripsGlobalFlagsFromPassthrough` — asserts both flags are consumed by lstk, not forwarded to the `aws` binary.


gtsiolis:
Follow up from https://github.com/localstack/lstk/pull/273#pullrequestreview-4400944981. Cc @carole-lavillonniere 

Possibly worth waiting for https://github.com/localstack/lstk/pull/251 to land first.

carole-lavillonniere (approved):

PR #276, opened 2026-06-02
Add lstk terraform command

peter-smith-phd:
## Summary

Adds a new `lstk terraform` / `lstk tf` command that proxies Terraform against a running LocalStack instance, mirroring the existing `lstk aws` pattern. The approach is based on the well-tested `tflocal`, with one significant improvement: the `endpoints {}` configuration fields are determined dynamically rather than hard-coded.

This PR now includes both the OpenSpec planning artifacts **and the full implementation** (the PR originally landed as specs-only).

- Auto-resolves the LocalStack endpoint via lstk's container discovery + host resolution (no host/port env vars).
- Generates a Terraform provider-override file (`localstack_providers_override.tf`) pointing every `aws` provider/alias block at LocalStack, then removes it after the run.
- Discovers the `endpoints {}` keys **dynamically** from `terraform providers schema -json` (AWS provider 4.0+), with **no fallback list** — fails with a "run `terraform init` first" message when the schema is unavailable.
- Special-cases the `s3` endpoint host (`s3.` prefix) for virtual-host addressing, gated by the same check that drives `s3_use_path_style`.
- Supports a reduced, `tflocal`-compatible env-var surface and `--region`/`--account` flags.

Part of [DPX-507](https://linear.app/localstack/issue/DPX-507/add-terraform-cli-proxy-command)

## Contents

**Implementation:**

- `cmd/terraform.go` — Cobra wiring for the `terraform` / `tf` command
- `internal/iac/terraform/cli/` — domain logic: endpoint/credential override generation, dynamic schema discovery (`schema.go`), provider-override file management (`override.go`), env-var handling (`env.go`), account/defaults resolution, and Terraform exec wrapping

**OpenSpec artifacts:**

- `proposal.md` — why / what changes / impact
- `design.md` — decisions, trade-offs, resolved/open questions
- `specs/terraform-proxy/spec.md` — requirements & scenarios
- `tasks.md` — implementation task breakdown

## Testing

- Unit tests for override, schema, and account logic.
- Integration tests (`cmd` + e2e) exercising the `single-bucket`, `explicit-endpoint`, `multiple-aliases`, `provider-v4`, and `submodule` sample configs.

Backend state / `terraform_remote_state` support is explicitly deferred to a follow-up.


carole-lavillonniere:
Switched the semver label to `minor` and documented them here: https://github.com/localstack/lstk/pull/278

openspec/changes/add-terraform-command/specs/terraform-proxy/spec.md
```diff
@@ -0,0 +1,246 @@
...
+
+#### Scenario: Dry run generates override but does not run terraform
+
+- **WHEN** `DRY_RUN` is enabled and a user runs `lstk terraform plan`
+- **THEN** the override file is generated and left in place for inspection
+- **AND** `terraform` is not invoked
+
+### Requirement: Configurable Terraform binary
+
+The system SHALL invoke the binary named by the `TF_CMD` environment variable when set (for example `tofu`), defaulting to `terraform` otherwise.
```
carole-lavillonniere: suggestion: I think that all env vars that are invented by us (and not AWS of terraform standards) should be prefixed with `LSTK_`: `TF_CMD`, `LS_PROVIDERS_FILE`, `DRY_RUN`, `SKIP_ALIASES`.   
Especially something like `DRY_RUN` sounds so generic that it could already be set in the env for another reason.
We've prioritized consistency and user-friendliness of the new experience rather than parity with old CLI.

openspec/changes/add-terraform-command/design.md
```diff
@@ -0,0 +1,120 @@
...
+
+### No spinner; subprocess (not exec-replace)
+Per the requirement, no spinner wraps terraform. Unlike `aws.go`, we do **not** wrap stdout/stderr in `StopOnWriteWriter`. We still run terraform as a managed child process (`exec.CommandContext`) rather than `syscall.Exec`, because the deferred override-file cleanup must run after terraform exits — `USE_EXEC`-style process replacement would skip it.
+
+### Override-file safety
+Before writing, check whether the target override file already exists. If it does and lstk did not create it, fail with a clear error (mirrors `tflocal`'s `check_override_file`) so we never delete a user's file during cleanup. Cleanup is a `defer` that removes only the paths lstk wrote.
```
carole-lavillonniere: question: how do we know if the file was created by lstk or not?

openspec/changes/add-terraform-command/tasks.md
```diff
@@ -0,0 +1,60 @@
...
+- [ ] 7.4 Integration test: `DRY_RUN` generates the override file and skips terraform; assert resolved `region`/`access_key` are encoded in the override
+- [ ] 7.5 Integration test: pre-existing override file causes a clear failure and is not deleted
+- [ ] 7.6 Integration test: leading `--region`/`--account` are stripped from forwarded args and encoded into the override; invalid `--account` fails before terraform is invoked
+- [ ] 7.7 Integration test: positional rules — `--region`/`--account` placed after the terraform action are forwarded to terraform unchanged (not consumed by lstk), and placing them before the `terraform` subcommand (`lstk --account … terraform`) is rejected
+- [ ] 7.8 Integration test (Docker-gated): override file is generated and removed after a real `terraform plan` against a running LocalStack, with endpoint keys sourced from the provider schema
+- [ ] 7.9 Integration test: when the provider schema is unavailable (no `terraform init` / provider not installed), the command fails with the specific "run `terraform init` first" message and does not invoke terraform or generate an override
```
carole-lavillonniere: nit: we could add a test case for tofu

openspec/changes/add-terraform-command/tasks.md
```diff
@@ -0,0 +1,60 @@
...
+- [ ] 7.2 Integration test: fails with "not running" error and does not invoke terraform when no emulator is running
+- [ ] 7.3 Integration test: `fmt`/`validate`/`version` run without generating an override file (including with `--region`/`--account` present, which are stripped and ignored)
+- [ ] 7.4 Integration test: `DRY_RUN` generates the override file and skips terraform; assert resolved `region`/`access_key` are encoded in the override
+- [ ] 7.5 Integration test: pre-existing override file causes a clear failure and is not deleted
+- [ ] 7.6 Integration test: leading `--region`/`--account` are stripped from forwarded args and encoded into the override; invalid `--account` fails before terraform is invoked
+- [ ] 7.7 Integration test: positional rules — `--region`/`--account` placed after the terraform action are forwarded to terraform unchanged (not consumed by lstk), and placing them before the `terraform` subcommand (`lstk --account … terraform`) is rejected
+- [ ] 7.8 Integration test (Docker-gated): override file is generated and removed after a real `terraform plan` against a running LocalStack, with endpoint keys sourced from the provider schema
```
carole-lavillonniere: thought: this looks like it's the only test that's requiring an actual terraform CLI + aws provider. It's good to test the full integration, I'm curious to see how CI behaves with the required downloads.

openspec/changes/add-terraform-command/specs/terraform-proxy/spec.md
peter-smith-phd: Good idea - I was just continuing to use the existing environment variables from `tflocal`, but that doesn't really make sense in this case. We should feel free to rename them to more meaningful names.

I'd propose:
* `LSTK_TF_CMD`
* `LSTK_TF_OVERRIDE_FILE_NAME`
* `LSTK_TF_DRY_RUN`
* and I'm going to complete remove `SKIP_ALIASES` since I don't understand what value it provides.

openspec/changes/add-terraform-command/design.md
peter-smith-phd: Yes, that's an odd wording in the design doc. The `and lstk did not create it` doesn't make sense, since we don't have any persistent record of whether `lstk` created it. We should just assume that if it already exists, then it's either because it was manually created (by somebody else), or it was orphaned by a previous `lstk tf` run. In either case, failing with an error and letting the user fix the problem is the correct solution.

openspec/changes/add-terraform-command/tasks.md
peter-smith-phd: I can actually think of numerous test cases in this area — I've been planning to execute them manually, but I might see if Claude can automate them efficiently. I'm curious about how long they take to run, and what their download dependencies are (much of the time, it's the `terraform init` that does the download).

Cases (all using `terraform plan`):
1. Testing with only a top-level Terraform project (all `.tf` in the same directory)
2. Testing with sub-modules (`.tf` in sub-directories), with and without their own AWS provider blocks.
3. Testing with a single AWS provider block
5. Testing with multiple aliased AWS provider blocks (in addition to the main AWS provider block)
6. Explicitly setting the endpoint in a provider block (some customers might do this if they need FIPS endpoints), then ensuring it's overridden.
7. Testing with v4.0 of the AWS provider (really old), and testing with the latest version of the AWS Provider.
8. Testing with `tofu`, simply because the behaviour might be subtly different in all the above test cases.

carole-lavillonniere: I suppose all of these will need docker (since they'll need the emulator running), so unfortunately they can't be run in parallel. Let's see how it looks, we could potentially add another parallel CI worker if the suite is getting too long.

carole-lavillonniere (approved):

cmd/terraform.go
```diff
@@ -0,0 +1,260 @@
...
+
+func newTerraformCmd(cfg *env.Env, logger log.Logger) *cobra.Command {
+	// DisableFlagParsing means Cobra won't strip lstk's own flags; PreRunE does
+	// that and stashes the remaining args here for RunE to forward to terraform.
+	var passthrough []string
+	return &cobra.Command{
+		Use:     "terraform [args...]",
+		Aliases: []string{"tf"},
+		Short:   "Run Terraform against LocalStack",
+		Long: `Proxy Terraform commands to LocalStack, using LocalStack endpoints as
```
carole-lavillonniere: issue: there should not be a line break here, otherwise:

Image

carole-lavillonniere:
Pushed 2 small commits to make error handling consistent 1b2a5135e0d0fd983e5e1e427aa7b77242664069 and 3b5d98b. Also updated `CLAUDE.md` to document this architecture decision since it is the standard for the code base.

## Before:
before

## After:
after

internal/iac/terraform/cli/exec.go
```diff
@@ -0,0 +1,125 @@
...
+		span.SetStatus(codes.Error, err.Error())
+		return fmt.Errorf("%s not found in PATH — install Terraform and ensure it is on your PATH", tfCmd())
+	}
+	span.SetAttributes(attribute.StringSlice("terraform.args", args), attribute.Bool("terraform.unproxied", IsUnproxied(args)))
+
+	if IsUnproxied(args) {
+		return runTerraform(ctx, span, tfBin, args)
+	}
+
+	workdir, err := os.Getwd()
```
carole-lavillonniere, line 57: question: shouldn't we handle a custom directory set with `-chdir=...`? https://developer.hashicorp.com/terraform/cli/commands#switching-working-directory-with-chdir

carole-lavillonniere:
nit: can we document the new command in `README.md` and `CLAUDE.md`?

carole-lavillonniere (approved):
I added a couple of minor issues/nits, but overall it looks good to me and it's ready to go from my point of view, once addressed! I think it'd be good if @gtsiolis also gives it a try before it lands on main and is made available to users this Thursday.

peter-smith-phd:
> Pushed 2 small commits to make error handling consistent [1b2a513]

Thanks for doing that! You've encouraged me to learn more about error reporting in `lstk`, since I completely missed that part.

internal/iac/terraform/cli/exec.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+		installURL := "https://developer.hashicorp.com/terraform/install"
+		if tfCmd() == "tofu" {
+			installURL = "https://opentofu.org/docs/intro/install/"
+		}
+		sink.Emit(output.ErrorEvent{
+			Title:   fmt.Sprintf("%s not found in PATH", tfCmd()),
+			Actions: []output.ErrorAction{{Label: "Install it and ensure it is on your PATH:", Value: installURL}},
```
gtsiolis, line 47: **praise**: Thanks for reusing this nudge with a better structure!

cmd/terraform.go
```diff
@@ -0,0 +1,282 @@
...
+
+var accountIDRe = regexp.MustCompile(`^\d{12}$`)
+
+func newTerraformCmd(cfg *env.Env, logger log.Logger) *cobra.Command {
+	// DisableFlagParsing means Cobra won't strip lstk's own flags; PreRunE does
+	// that and stashes the remaining args here for RunE to forward to terraform.
+	var passthrough []string
+	return &cobra.Command{
+		Use:     "terraform [args...]",
+		Aliases: []string{"tf"},
```
gtsiolis, line 29: **thought**: Terraform CLI ships as a single executable named `terraform`, there's no official `tf`, though it's a common alias people set themselves from what I understand. The proxy commands are meant to just prefix your existing command with `lstk`, whereas this alias is really about shorter typing. I think it's fine to leave as is, but wanted to highlight it.

internal/iac/terraform/cli/override_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/localstack/lstk/internal/log"
+)
+
+func baseOpts(workdir string) overrideOptions {
+	return overrideOptions{
+		workdir:      workdir,
+		fileName:     defaultOverrideFileName,
+		endpointURL:  "http://127.0.0.1:4566",
```
dmacvicar: Is there a reasoning for not using localhost.localstack.cloud here?

peter-smith-phd: No obvious reason. I've updated it to be more readable.

gtsiolis (commented):
Works as advertised. 💯 

Leaving two comments retroactively.

PR #280, opened 2026-06-03
Check runtime availability before login and emulator selection

gtsiolis:
The Docker health check ran inside `container.Start`, after login and emulator selection — so a user without Docker would log in, pick an emulator (with the choice persisted), and only then hit `Docker is not available`. This moves it to a pre-flight gate that runs before the login and emulator-selection prompts, so a missing runtime is reported up front.

Closes [PRO-302](https://linear.app/localstack/issue/PRO-302/check-runtime-availability-before-login-and-emulator-selection)

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

| BEFORE/AFTER |
| -- |
|  |

### DIFF
--- a/internal/ui/run.go
+++ b/internal/ui/run.go
@@ -70,6 +70,14 @@ func Run(parentCtx context.Context, runOpts RunOptions) error {
 			p.Send(runDoneMsg{})
 			return
 		}
+		if healthErr := runOpts.Runtime.IsHealthy(ctx); healthErr != nil {
+			if errors.Is(healthErr, context.Canceled) {
+				return
+			}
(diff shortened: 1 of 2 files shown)


anisaoshafi:
@gtsiolis the change looks valid, thanks for looking into this 🐳 

There are 3 tests that are failing on macos. Could you take a look at them before I approve? 
Let me know if you need help!

gtsiolis:
Thanks for taking a look, @anisaoshafi!

Added 39f32d4 to fix the checks and added a visualization of the change in the PR description.

anisaoshafi (approved):
LGTM! :shipit:

PR #281, opened 2026-06-03
Implement snapshot remove

anisaoshafi:
Implemented snapshot remove. Works for cloud pods only, and throws an error in local for safety reason. It requires confirmation (Y/n) and in non-interactive mode can be bypassed by using `--force` flag.

image
602367388-62f127ad-65a7-45ac-8f79-e322daac7c83
image
image
image

### DIFF
--- a/internal/snapshot/destination.go
+++ b/internal/snapshot/destination.go
@@ -39,6 +39,21 @@ type Destination struct {
 	Value string
 }
 
+// ParseRemovable parses a ref for snapshot remove. Only cloud (pod:) refs are accepted;
+// local file paths are rejected because the CLI cannot delete local files.
+// cwd and home are used to produce a human-readable path in error messages.
+func ParseRemovable(ref, cwd, home string) (Destination, error) {
+	lower := strings.ToLower(ref)
+	if !strings.HasPrefix(lower, "pod:") && !strings.Contains(lower, "://") {
+		abs, _ := filepath.Abs(ref)
+		if !strings.EqualFold(filepath.Ext(abs), ".zip") {
+			abs += ".zip"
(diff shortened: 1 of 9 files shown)


internal/snapshot/remove.go
```diff
@@ -0,0 +1,88 @@
+package snapshot
+
+//go:generate mockgen -source=remove.go -destination=mock_pod_remover_test.go -package=snapshot_test
```
carole-lavillonniere: nit: the mock was not generated and is also not used, so this line could be removed

carole-lavillonniere (approved):

internal/snapshot/destination.go
```diff
@@ -39,6 +39,21 @@ type Destination struct {
...
+// local file paths are rejected because the CLI cannot delete local files.
+// cwd and home are used to produce a human-readable path in error messages.
+func ParseRemovable(ref, cwd, home string) (Destination, error) {
+	lower := strings.ToLower(ref)
+	if !strings.HasPrefix(lower, "pod:") && !strings.Contains(lower, "://") {
+		abs, _ := filepath.Abs(ref)
+		if !strings.EqualFold(filepath.Ext(abs), ".zip") {
+			abs += ".zip"
+		}
+		return Destination{}, fmt.Errorf("'%s' resolves to a local file (%s); CLI cannot delete local files", ref, displayPath(abs, cwd, home))
```
carole-lavillonniere, line 52: issue: the error does not show as an error in the TUI

anisaoshafi: Thank you! Now looks like this, let me know if I missed sth:  image

PR #282, opened 2026-06-03
Improve help text wrapping

gtsiolis:
Command help text was hard-wrapped in the Go source (e.g. `lstk start`'s `Long` broke mid-sentence between two string lines). That froze the break at a single width: a stubby first line on wide terminals, ragged re-wrapping on narrow ones, and — because `lstk docs` generates man/markdown straight from the raw `Long` — the hard break leaked into the generated docs too. This grew out of #277, which added the `LOCALSTACK_` env-var forwarding note and hit exactly this problem.

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-06-04 at 00 09 07 | Screenshot 2026-06-04 at 00 31 57 | 
| Screenshot 2026-06-04 at 00 32 26 | Screenshot 2026-06-04 at 00 32 13 |

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ SERVICES = "s3,sqs"
 EAGER_SERVICE_LOADING = "1"
 ```
 
-In addition, host environment variables prefixed with `LOCALSTACK_` (and the `CI` variable) are automatically forwarded to the emulator container.
+Host environment variables prefixed with `LOCALSTACK_` are also forwarded to the container.
 
 ## Interactive And Non-Interactive Mode
 
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ SERVICES = "s3,sqs"
 EAGER_SERVICE_LOADING = "1"
 ```
 
-Host environment variables prefixed with `LOCALSTACK_` are also forwarded to the container.
+Host environment variables prefixed with `LOCALSTACK_` are also forwarded to the emulator.
 
 ## Interactive And Non-Interactive Mode
 
--- a/cmd/help.go
+++ b/cmd/help.go
@@ -41,11 +41,8 @@ const helpTemplate = `{{if not .HasParent}}{{if or .Runnable .HasSubCommands}}{{
 {{end}}{{if or .Runnable .HasSubCommands}}{{.UsageString}}{{end}}{{end}}
 `
 
-func init() {
-	cobra.AddTemplateFunc("wrapText", wrapText)
-}
-
 func configureHelp(cmd *cobra.Command) {
+	cobra.AddTemplateFunc("wrapText", wrapText)
 	cmd.InitDefaultHelpFlag()
 	cmd.Flags().Lookup("help").Usage = "Show help"
 	cmd.SetUsageTemplate(usageTemplate)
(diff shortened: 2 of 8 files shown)


cmd/start.go
```diff
@@ -15,8 +15,7 @@ func newStartCmd(cfg *env.Env, tel *telemetry.Client, logger log.Logger) *cobra.
 		Short: "Start emulator",
 		Long: `Start emulator and services.
 
-Host environment variables prefixed with LOCALSTACK_ (and the CI variable) are
-forwarded to the emulator container.`,
+Host environment variables prefixed with LOCALSTACK_ are forwarded to the emulator.`,
```
anisaoshafi, line 18: `CI` var is also forwarded. I understand it's a bit ugly to document every single variable, apart from the pattern. Did you remove it on purpose?

README.md
```diff
@@ -143,7 +143,7 @@ SERVICES = "s3,sqs"
 EAGER_SERVICE_LOADING = "1"
 ```
 
-In addition, host environment variables prefixed with `LOCALSTACK_` (and the `CI` variable) are automatically forwarded to the emulator container.
+Host environment variables prefixed with `LOCALSTACK_` are also forwarded to the container.
```
anisaoshafi: [Same](https://github.com/localstack/lstk/pull/282/changes#r3356065289)

cmd/help.go
```diff
@@ -30,14 +36,69 @@ Global Options:
...
   {{rpad .CommandPath .CommandPathPadding}} {{.Short}}{{end}}{{end}}{{end}}`
 
-const helpTemplate = `{{if not .HasParent}}{{if or .Runnable .HasSubCommands}}{{.UsageString}}{{end}}{{else}}{{with (or .Long .Short)}}{{. | trimTrailingWhitespaces}}
+const helpTemplate = `{{if not .HasParent}}{{if or .Runnable .HasSubCommands}}{{.UsageString}}{{end}}{{else}}{{with (or .Long .Short)}}{{wrapText . | trimTrailingWhitespaces}}
 
 {{end}}{{if or .Runnable .HasSubCommands}}{{.UsageString}}{{end}}{{end}}
 `
 
+func init() {
+	cobra.AddTemplateFunc("wrapText", wrapText)
```
anisaoshafi: suggestion: could you move `cobra.AddTemplateFunc("wrapText", wrapText)` inside `configureHelp(` and delete `init()` func instead, since it's meant to be consumed by ConfigureHelp?

anisaoshafi (approved):
Looks good to me, thanks ✨  
just two small comments to tackle

cmd/help.go
gtsiolis: Good catch! ⚾ 

Moved in 1530c9d.

cmd/start.go
gtsiolis: Yes, removed this on purpose. The `CI` variable isn't part of the configuration contract, we forward it as an internal mechanism for CI detection, etc.

README.md
gtsiolis: See https://github.com/localstack/lstk/pull/282#discussion_r3356362839.

anisaoshafi (approved):
📦

PR #285, opened 2026-06-05
Implement snapshot list

anisaoshafi:
Implements snapshot list. Currently lists cloud pods.
image
image

### DIFF
--- a/internal/snapshot/list.go
+++ b/internal/snapshot/list.go
@@ -1,0 +1,66 @@
+package snapshot
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/localstack/lstk/internal/config"
+	"github.com/localstack/lstk/internal/container"
+	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/runtime"
+)
+
+// PodSnapshot holds metadata for a single Cloud Pod returned by the list endpoint.
+type PodSnapshot struct {
+	Name        string
+	Version     int
+	LastChanged *time.Time
+}
+
+// PodLister retrieves available Cloud Pods from the running LocalStack instance.
+type PodLister interface {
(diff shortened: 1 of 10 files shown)


internal/snapshot/list.go
```diff
@@ -0,0 +1,66 @@
...
+
+	entries := make([]output.PodSnapshotEntry, len(pods))
+	for i, p := range pods {
+		entries[i] = output.PodSnapshotEntry{
+			Name:        p.Name,
+			Version:     p.Version,
+			LastChanged: p.LastChanged,
+		}
+	}
+	sink.Emit(output.SnapshotListEvent{Pods: entries})
```
carole-lavillonniere, line 64: question: weren't we able to reuse `TableEvent` from the status command rather than introduce a new `SnapshotListEvent` event?

carole-lavillonniere (approved):
Looks good, just one question/suggestion.

internal/snapshot/list.go
anisaoshafi: Hm, technically possible. In this case makes more sense to me to keep it separate since we also put a small summary line e.g ("~ 2 snapshots"), which is the same unit as the table itself. 
Also using TableEvent, it would render inline in the TUI buffer, which looks broken for a long list (according to claude). 
If you have a strong opinion happy to change it :)

carole-lavillonniere: I think it's worth giving it a try if you get the chance. We could get rid of extra code.
Status also has a small summary at the top which is rendered separately:

image
Not sure what is broken with a long list, but if that is actually the case it would also be an issue for the status results, right?

anisaoshafi: Agreed, thanks for clarifying. I'll follow up in another PR, ticket DEVX-907.

PR #292, opened 2026-06-09
Show emulator selector until one is selected

gtsiolis:
Running `lstk start` shows a "Which emulator?" picker on first run. It was tied to "config file just created" — but the config is written before the Docker check, so a first run with no Docker created the config and used up the picker. Start Docker, rerun, and the picker never came back: you were stuck on the AWS default.

Now config creation is deferred until the run actually does something. A run that fails early (e.g. no Docker) writes nothing, so the next run still shows the picker.

Closes PRO-305.


gtsiolis:
Cc @anisaoshafi @carole-lavillonniere because https://github.com/localstack/lstk/pull/217.

carole-lavillonniere (approved):

PR #293, opened 2026-06-09
Add input validation against malformed inputs

gtsiolis:
## Motivation

lstk is increasingly invoked by AI agents and scripts, which can produce malformed or hostile input (control characters, path traversal, percent-encoding, shell metacharacters) in ways humans rarely do. Pod names, the auth token, and `[env.*]` config values were only loosely checked (or not at all), so bad input surfaced as a confusing error deep inside a Docker call or platform API request instead of failing fast with a clear reason.

## Changes

- Add `internal/validate`, reusable deterministic validators (`ResourceName`, `EnvVarName`, `NoControlChars`, `AuthToken`) that return a machine-classifiable `*Error` (stable `Rule` field) instead of a generic message.
- Wire `validate.AuthToken` into `cmd/root.go`'s token resolution (env/keyring): trims whitespace, then rejects control chars, embedded whitespace, or implausibly long tokens.
- Wire `validate.EnvVarName`/`NoControlChars` into `internal/config/config.go` to reject malformed `[env.*]` keys/values before they're injected into a container's environment.
- Replace `internal/snapshot`'s bare pod-name regex with `validate.ResourceName`, adding deny-checks for percent-encoding, path traversal, embedded path/query characters, and shell metacharacters.
- Rebase conflict resolution: current `main` had independently grown its own pod-name validator (`ValidatePodName`, from the S3 remote-storage PR #344) that permits underscores and is also called directly for S3-remote pod names in `cmd/snapshot.go`. `ValidatePodName` now delegates to `validate.ResourceName`, whose charset was widened to keep underscores working — the S3-remote pod-name path now gets the same deny-checks as `pod:` refs, which it didn't before.
- Document the new package in the root `CLAUDE.md`, and encode its usage as a convention in the agent instructions so future input surfaces route through it instead of growing parallel validators: a Code Style rule in `CLAUDE.md` (route by value class — identifiers → `ResourceName`; opaque secrets → loose `AuthToken`-style checks; paths/URLs → their existing parsers; extend `internal/validate` rather than inlining a regexp), an input question + wiring/test steps in the `add-command` skill, and a checklist item in the `review-pr` skill.

## Review

Human review advised — changes user-facing error behavior for auth tokens, pod names, and `[env.*]` config (previously-accepted values are now rejected with new/different messages), and this rebase merges two independently-evolved pod-name validators.

## Tests

New unit tests in `internal/validate/validate_test.go` (all validators, deny-check ordering, rule codes) and `internal/config/env_validation_test.go`; extended `internal/snapshot/destination_test.go` with malformed pod-name cases (percent-encoding, embedded query, shell metacharacters) plus a regression case confirming underscore-named pods still pass. Verified after rebase: `make build`, `go vet ./...`, `make test` (full suite), `make lint` (both modules) all green.

Closes PRO-306


gtsiolis:
Updated PR description. I'd appreciate a human review following the suggestion from the PR description. 🤖

gtsiolis:
:fyi: the last commit (f2a04f9) is instructions only, no runtime change.

It turns the validation approach into a documented convention: a Code Style rule in CLAUDE.md routing values by class (identifiers through `validate.ResourceName`, opaque secrets through loose `AuthToken`-style checks, paths and URLs through their existing parsers), plus matching scaffold and test steps in the `add-command` skill and a checklist item in the `review-pr` skill.

Why: the pod name rules had already forked once while this branch was open (#344 grew its own validator), so this points new input surfaces at `internal/validate` instead of letting parallel regexps drift again.

skyrpex (approved):
Looks good to me! The only thing that triggers my spider-sense is maybe building over-specific validation for generic inputs, for example:

Are we certain that a ResourceName will **never** contain `..`, `/`, `?` or `#`?

gtsiolis:
Good spider-sense! 🕸️

Did some investigation on this, and confirmed that AWS has no universal resource-name grammar. S3 object keys can contain `/`, `?`, and valid `..` segments, and CloudWatch log group names can even contain `#`, so a generic validator can't be right for those.

The validator in this PR only guards Cloud Pod names, and naming it `ResourceName` could have caused confusion. I renamed it to `PodName` and updated the conventions to route other identifiers to their owning API's contract. 


For pod names we can be certain since the platform enforces `^[a-zA-Z0-9_-]+$` (see `verify_pod_name` in localstack-platform, same pattern in the emulator's pods bootstrap), and the name is embedded in the URL path (`/v1/cloudpods/<name>`), so `..`, `/`, `?`, `#` can't appear in a valid name.

I aligned the allow-list to that exact pattern, which also fixes a real over-restriction: we previously rejected leading `_`/`-` names the platform accepts.

I'd appreciate a second round when you have some time, @skyrpex! 🙏

gtsiolis:
Pushed one more commit to fix the pipeline.

skyrpex (approved):
I looks better now, thanks!

gtsiolis:
Awesome, merging. 🏀

PR #295, opened 2026-06-10
List snapshots without requiring a running emulator

gtsiolis:
In https://github.com/localstack/lstk/pull/285 we added `snapshot list` to list snapshots.

The previous implementation routed the request through the container's `GET /_localstack/pods`, so it required Docker to be up *and* an emulator running just to read remote metadata. 

That inverts the normal flow and blocks a common use case (browse snapshots, then load one).

| BEFORE (without runtime) | BEFORE (without instance) | AFTER |
|--------|--------|--------|
| Screenshot 2026-06-10 at 15 11 49 | Screenshot 2026-06-10 at 15 11 58 | Screenshot 2026-06-10 at 14 15 41 |

### DIFF
--- a/cmd/snapshot.go
+++ b/cmd/snapshot.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/localstack/lstk/internal/api"
 	"github.com/localstack/lstk/internal/config"
 	"github.com/localstack/lstk/internal/container"
 	"github.com/localstack/lstk/internal/emulator/aws"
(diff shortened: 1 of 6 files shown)


gtsiolis:
Cc @anisaoshafi @carole-lavillonniere because https://github.com/localstack/lstk/pull/285

anisaoshafi:
on it 🔍

anisaoshafi (approved):
I knew this was an option, but went for the lazy approach relying on the emulator as the old CLI did 😬 
Thanks for improving the user experience, no need for coupling this logic to the emulator 🧹

gtsiolis:
Thanks for taking a look, @anisaoshafi!

PR #296, opened 2026-06-10
Classify agents on every command invocation

gtsiolis:
Tags every command invocation in CLI telemetry so we can answer "are agents using lstk?". Detection is environment-based (the strongest signal) with a TTY fallback for humans, and is attached to every telemetry event. A detected agent is also forwarded into the emulator container as `AI_AGENT`.

Agent and CI are tracked as independent axes — an agent can run inside CI, so each is recorded separately rather than collapsed into one mutually exclusive value. Every event carries `is_ci` plus `agent_identity` / `ci_identity` when detected, and a derived `caller_type` (agent/ci/human) for single-label segmentation. This mirrors the legacy CLI (localstack/localstack-cli-standalone#19), where `is_ci` is its own field independent of agent detection.

See [PRO-276](https://linear.app/localstack/issue/PRO-276/classify-agents-on-every-command-invocation) for more context.

### Notes

- Orthogonal axes — an agent running inside CI is recorded on both (`agent_identity` and `is_ci`/`ci_identity`); `caller_type` derives to `agent` (agent > ci > human) so segmentation stays single-label without discarding the CI signal.
- `is_ci` is sourced from the single caller classifier (14 known CI providers), not a separate bare `CI` env lookup.
- `AI_AGENT` on the container is session-origin — it reflects who started the container, not who makes later calls into it, and is forwarded whenever an agent is detected (including inside CI).
- Detection-only, no free-text override, so no user-supplied value reaches telemetry and no sanitization is needed.

### DIFF
--- a/internal/caller/caller.go
+++ b/internal/caller/caller.go
@@ -6,25 +6,40 @@ import (
 	"golang.org/x/term"
 )
 
-type Type string
-
-const (
-	TypeHuman Type = "human"
-	TypeAgent Type = "agent"
-	TypeCI    Type = "ci"
-)
-
-const (
-	MethodAgentEnv = "agent_env"
-	MethodCIEnv    = "ci_env"
-	MethodTTY      = "tty"
-	MethodNoTTY    = "no_tty"
-)
-
 type Classification struct {
-	Type     Type
-	Identity string
-	Method   string
+	AgentIdentity string
+	CIIdentity    string
+	Interactive   bool
+}
+
+func (c Classification) IsAgent() bool { return c.AgentIdentity != "" }
+
+func (c Classification) IsCI() bool { return c.CIIdentity != "" }
+
+func (c Classification) IsHuman() bool { return c.AgentIdentity == "" && c.CIIdentity == "" }
+
+func (c Classification) CallerType() string {
+	switch {
+	case c.AgentIdentity != "":
+		return "agent"
+	case c.CIIdentity != "":
+		return "ci"
+	default:
+		return "human"
+	}
+}
+
+func (c Classification) DetectionMethod() string {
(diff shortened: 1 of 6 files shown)


gtsiolis:
Cc @carole-lavillonniere @anisaoshafi because https://github.com/localstack/lstk/pull/58, https://github.com/localstack/lstk/pull/127, etc.
Cc @andrewrjones @purcell @whummer because https://github.com/localstack/localstack-cli-standalone/pull/19

internal/caller/caller.go
```diff
@@ -0,0 +1,118 @@
...
+	"golang.org/x/term"
+)
+
+type Type string
+
+const (
+	TypeHuman Type = "human"
+	TypeAgent Type = "agent"
+	TypeCI    Type = "ci"
+)
```
carole-lavillonniere: question: why do we have a type with mutually exclusive values (ci can also be agent but we can't represent this here), when instead we could have 2 binary fields: `agent=yes/no ci=yes/no`? Data model could become something like:
```
  is_agent / agent_identity   (cursor, claude-code, …)
  is_ci    / ci_identity      (github-actions, jenkins, …)
  detection_method
```

Then we could also remove the other lookup in` internal/telemetry/client.go` line 77 but instead use the value from` internal/caller/caller.go`

gtsiolis: Nice catch, @carole-lavillonniere! ⚾ 

I was trying to mirror the approach from localstack/localstack-cli-standalone#19, but you're right! Made some changes in 27b8021, and updated PR description. Could you take another look and take it over from here? 🙏

carole-lavillonniere (approved):

PR #297, opened 2026-06-10
Fix silently ignored error when resolving license file path on logout

gtsiolis:
When `os.UserCacheDir()` fails (e.g. in a minimal container environment), the license file path would silently fall back to an empty string, meaning logout couldn't clean up the cached license file.

This change propagates the error to the caller instead, so the failure is surfaced with a clear message rather than ignored.


carole-lavillonniere (approved):
Good catch!

gtsiolis:
Thanks for taking a look, @carole-lavillonniere! 🏀

PR #302, opened 2026-06-11
feat: Add `lstk cdk` proxy command

peter-smith-phd:
## Summary

Adds `lstk cdk`, a proxy that runs the real AWS CDK CLI against a running LocalStack emulator — mirroring `lstk aws` / `lstk terraform`. This replaces the existing `cdklocal` command, with the one caveat that using and endpoint of `http://127.0.01:4566` or `http://localhost:4566` will _NOT_ work correctly in some cases. A future solution is still required for this edge case.

## Implementation

- Most of this implementation was a direct copy from the existing `cdklocal` tool. See `openspec/changes/add-cdk-proxy-command/` for the full proposal, design, and spec.

- **Endpoint via env vars** (`internal/iac/cdk/cli/`): `BuildEnv` sets `AWS_ENDPOINT_URL`, the derived `AWS_ENDPOINT_URL_S3`, mock creds, and region on the cdk subprocess.
- **Safety — strip ambient AWS config**: `AWS_PROFILE`/`AWS_DEFAULT_PROFILE`/`AWS_SESSION_TOKEN` are removed so a user's real credentials can't redirect a deploy at real AWS. This is the safety-critical core, mirroring `cdklocal`.
- **Version floor 2.177.0**: older CDK ignores the endpoint env vars and would silently hit real AWS, so `Run` checks `cdk --version` up front and fails with an actionable error. The caveat is that users wanting to use an older version of CDK must resort to `cdklocal` instead.
- **S3 endpoint addressing** (`internal/endpoint`): shared `S3Addressing` derives the `s3.`-prefixed virtual-host endpoint for `*.localstack.cloud` hosts. Path-style on the `127.0.0.1` DNS-rebind fallback is a known limitation (CDK exposes `forcePathStyle` only as a code-level arg, unreachable from a subprocess); lstk warns and the fix is out of scope for v1.
- **Shared IaC helpers → `cmd/iac.go`**: the command-boundary helpers (`--region`/`--account` parsing, `requireRunningAWSEmulator`, `resolveAWSContainer`, validation) moved out of `cmd/terraform.go` into a neutral file so cdk and terraform share them without one looking terraform-owned.

## Testing

- **`cdk_cmd_test.go`** — stub-`cdk` integration tests: arg/exit-code passthrough, env construction, emulator gating, version floor, flag handling, `LSTK_CDK_CMD`.
- **`cdk_e2e_test.go`** — real cdk + real LocalStack running in Docker.

Resolves [DPX-505](https://linear.app/localstack/issue/DPX-505/aws-cdk-cli-proxy-command).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,14 @@ jobs:
         with:
           tofu_wrapper: false
 
+      # Install the AWS CDK CLI so the `lstk cdk` end-to-end tests
+      # (cdk_e2e_test.go) run on the Docker-capable Linux shards. They skip
+      # automatically wherever cdk is absent (macOS/Windows). lstk requires
+      # CDK >= 2.177.0; the latest release satisfies that.
+      - name: Install AWS CDK
+        if: matrix.os == 'ubuntu-latest'
+        run: npm install -g aws-cdk
+
       - name: Run integration tests
         run: make test-integration
         env:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,7 +43,7 @@ Note: Integration tests require `LOCALSTACK_AUTH_TOKEN` environment variable for
   - `ui/` - Bubble Tea views for interactive output
   - `update/` - Self-update logic: version check via GitHub API, binary/Homebrew/npm update paths, archive extraction
   - `log/` - Internal diagnostic logging (not for user-facing output — use `output/` for that)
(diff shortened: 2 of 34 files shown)


anisaoshafi (approved):
LGTM, briefly tested it and seems to work fine 👏🏼 

small nit: would be good to update the [iac section](https://github.com/localstack/lstk/blob/main/CLAUDE.md?plain=1#L46) that cdk is now supported

PR #303, opened 2026-06-12
Include LocalStack version in already-running note

gtsiolis:
When `lstk start` detects an already-running emulator, the note now includes the running version. The version is fetched from `/_localstack/info` (and any build suffix is stripped); if the lookup fails, we fall back to the previous bare note.

Before:

`> Note: LocalStack AWS Emulator is already running`

After:

`> Note: LocalStack AWS Emulator 2026.5.3 is already running`

The change is in `emitAlreadyRunning`, which covers both detection paths (named container via `IsRunning` and externally-managed containers found via `FindRunningByImage`).

Added unit tests covering the version-bearing note and the fallback when the version is unavailable.


anisaoshafi (approved):
Thank you 🦅 👁️

PR #304, opened 2026-06-12
Replace the local snapshot extension .zip to .snapshot

anisaoshafi:
Use `.snapshot` extension for saving snapshots locally.
image
cc @gtsiolis

### DIFF
--- a/cmd/snapshot.go
+++ b/cmd/snapshot.go
@@ -41,9 +41,9 @@ const snapshotSaveLong = `Save a snapshot of the running emulator's state.
 
 Pass [destination] as an absolute or relative path for the exported file:
 
-  lstk snapshot save                    # saves to ./snapshot-<YYYY-MM-DDTHH-mm-ss>-<hex>.zip
-  lstk snapshot save ./my-snapshot.zip  # saves to ./my-snapshot.zip
-  lstk snapshot save /tmp/my-state      # saves to /tmp/my-state.zip
+  lstk snapshot save                         # saves to ./snapshot-<YYYY-MM-DDTHH-mm-ss>-<hex>.snapshot
+  lstk snapshot save ./my-snapshot.snapshot  # saves to ./my-snapshot.snapshot
+  lstk snapshot save /tmp/my-state           # saves to /tmp/my-state.snapshot
 
 To save to a remote pod on the LocalStack platform, use the pod: prefix:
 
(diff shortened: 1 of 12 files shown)


internal/snapshot/destination.go
```diff
@@ -23,6 +23,20 @@ var (
 	validPodName = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9-]*$`)
 )
 
+const (
+	snapshotExt       = ".snapshot" // user-facing extension for local snapshots
+	legacySnapshotExt = ".zip"      // accepted on load for backward compatibility
```
gtsiolis, line 28: **praise**: Thanks for adding this! 🙏

cmd/snapshot.go
```diff
@@ -55,9 +55,9 @@ const snapshotLoadLong = `Load a snapshot into the running emulator, starting it
 
 REF identifies the snapshot to load:
 
-  lstk snapshot load my-baseline           # loads ./my-baseline or ./my-baseline.zip
-  lstk snapshot load ./checkpoint.zip      # loads from explicit path
-  lstk snapshot load pod:my-baseline       # loads from LocalStack Cloud
+  lstk snapshot load my-baseline             # loads ./my-baseline or ./my-baseline.snapshot
```
gtsiolis, line 58: **question:** Couldn't find a better line to add this comment, but loading an invalid file returns an error. Could we update that and possibly remove the file extension reference?

Image

gtsiolis, line 58: **question**: We don't ship any manifest or a way to detect whether it's a valid snapshot file, right?

Image

gtsiolis (commented):
Left some comments below we discuss async or sync!

cmd/snapshot.go
anisaoshafi: Not afaik. cc @giograno to correct me in case there is a way

anisaoshafi: Good catch, will change 🙏🏼

giograno: Correct, we do not!

gtsiolis: Thanks for confirming, @giograno! Leaving this out of scope for now then.

anisaoshafi: fyi, this error was returned from emulator in [this](https://github.com/localstack/localstack-pro/blob/main/localstack-pro-core/localstack/pro/core/persistence/pods/endpoints.py#L245) line, we only prefix it with `load failed for service:`.
I modified it to look like this. let me know if it needs more tweaking. thank you!
image

gtsiolis: Perfect, this matches the error pattern and style we introduced for other cases, too. 💯

test/integration/snapshot_load_test.go
```diff
@@ -169,6 +187,52 @@ func TestSnapshotLoadLocalBareNameFallback(t *testing.T) {
 	assert.Contains(t, stdout, "Snapshot loaded")
 }
 
+// TestSnapshotLoadLocalLegacyZipFallback verifies that snapshots saved as .zip by
+// older lstk versions still load by bare name.
+func TestSnapshotLoadLocalLegacyZipFallback(t *testing.T) {
```
carole-lavillonniere, line 192: praise: good call testing the legacy behavior

carole-lavillonniere (approved):

PR #315, opened 2026-06-16
Implement snapshot show

anisaoshafi:
Added support for `lstk snapshot show`
image
image


carole-lavillonniere:
issue: it seems like some special characters are not well handled. This pod exist in the LS org: 
image

anisaoshafi:
> issue: it seems like some special characters are not well handled. This pod exist in the LS org: image

Thanks for testing it, will adjust it 🙏🏼 
_update_: fixed in https://github.com/localstack/lstk/pull/315/commits/b4be3ef27f5342bce51620ae70f515c520cca7e5, thanks @carole-lavillonniere

carole-lavillonniere (approved):

PR #320, opened 2026-06-18
Stop integration tests from sending telemetry to production

carole-lavillonniere:
## Motivation

Integration tests run the real `lstk` binary, whose analytics endpoint defaults to production (`https://analytics.localstack.cloud`). Aany test that doesn't explicitly redirect analytics to a mock sends real `lstk_command`/`lstk_lifecycle` events to the production backend, polluting analytics.

## Changes

- Default `LSTK_ANALYTICS_ENDPOINT` to an unreachable local address (`http://127.0.0.1:1`) in the integration test env helpers (`env.With` / `env.Without`), so the binary under test can never reach production.
- Tests that exercise telemetry are unaffected: they redirect to a mock server, and the explicit value wins because `exec` dedups duplicate env keys to the last value.

Closes DEVX-928

### DIFF
--- a/test/integration/env/env.go
+++ b/test/integration/env/env.go
@@ -22,6 +22,14 @@ const (
 	OtelEndpoint      Key = "OTEL_EXPORTER_OTLP_ENDPOINT"
 )
 
+// UnreachableAnalyticsEndpoint is a closed local port used as the default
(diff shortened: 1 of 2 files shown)


anisaoshafi (approved):
Great catch 🎣

PR #321, opened 2026-06-18
Set SF_S3_ENDPOINT from Snowflake emulator port config

gtsiolis:
The Python Snowflake emulator funnels all S3 access through a single `SF_S3_ENDPOINT` that defaults to port `4566`. When the Snowflake emulator runs on a custom port, internal-stage operations like `COPY INTO` fail because S3 access still targets `4566`.

This sets `SF_S3_ENDPOINT` on startup to match the configured Snowflake port (`s3.localhost.localstack.cloud:<port>`), unless the user already provided their own value via an env profile. AWS + Snowflake multi-emulator setups remain unsupported in this flow for now — users who need both should use a single container image.

Tests: a unit test for `snowflake.S3Endpoint` and an integration test asserting `SF_S3_ENDPOINT` matches a custom configured port.

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -108,6 +108,15 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, opts Start
 			"LOCALSTACK_HOST="+endpoint.Hostname+":"+c.Port,
 		)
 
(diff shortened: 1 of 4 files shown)


anisaoshafi (approved):
Thanks for fixing this 👏🏼

PR #325, opened 2026-06-18
feat: support custom container images via config

gtsiolis:
## What

Adds an optional per-container `image` field so users can override the default Docker Hub image — e.g. point at an internal registry mirror or a locally loaded offline image.

`ContainerConfig.Image()`:
- returns `image` as-is when it already carries a tag,
- otherwise appends the configured `tag` (or `latest`),
- falls back to the default `localstack/<product>:<tag>` when `image` is unset.

## Scope

Split out of the original DEVX-703 work to keep reviews small. **This PR is image override only.** Rebased on `main` (which now includes #339).

The offline / slow-network handling that originally motivated this branch is **stacked on top of this PR in #337** (graceful degradation when the registry or license server is unreachable). The success-path integration test raised in review — a configured image found locally ⇒ no pull and no CLI license check ⇒ container starts — lives on #337, since it depends on the start-path logic this PR doesn't touch.

Related follow-ups (tracked under PRO-219):

- **PRO-323** — reuse local image when already present → #339 (merged)
- **PRO-324** — ESC to abort an image pull, falling back to the local image
- **PRO-325** — smart refresh: `pull_policy` config + `--pull` flag
- **PRO-326** — daily non-blocking "newer image available" notice

## Tests

- Unit: `internal/config` `Image()` resolution cases (with/without tag, registry port, digest, default).
- Integration: `TestStartWithCustomImageFailsClearlyWhenUnavailable` — a configured custom image is honored and fails clearly when it can be neither pulled nor found locally.

## Known limitations

- Running-container discovery (`FindRunningByImage`) still matches only the default `localstack/*` repos, so a container started from a *custom* `image` is invisible to the image-based scans used for port-conflict / cross-type detection and stop/status reuse. The common re-run path is unaffected (it matches by container name). Extending discovery to custom images is tracked as follow-up.

Refs DEVX-703

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,6 +66,8 @@ When adding a new command that depends on configuration, wire config initializat
 
 Created automatically on first run with defaults. Supports emulator types: `aws`, `snowflake`, and `azure`.
 
+Each `[[containers]]` block may set an optional `image` to override the default Docker Hub image (e.g. an internal registry mirror or a locally loaded offline image). `ContainerConfig.Image()` returns `image` as-is when it already carries a tag, otherwise it appends `tag` (or `latest`); the default `localstack/<product>:<tag>` is used when `image` is unset.
+
 # Emulator Setup Commands
 
 Use `lstk setup <emulator>` to set up CLI integration for an emulator type:
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,7 +66,7 @@ When adding a new command that depends on configuration, wire config initializat
 
 Created automatically on first run with defaults. Supports emulator types: `aws`, `snowflake`, and `azure`.
 
-Each `[[containers]]` block may set an optional `image` to override the default Docker Hub image (e.g. an internal registry mirror or a locally loaded offline image). `ContainerConfig.Image()` returns `image` as-is when it already carries a tag, otherwise it appends `tag` (or `latest`); the default `localstack/<product>:<tag>` is used when `image` is unset.
+Each `[[containers]]` block may set an optional `image` to override the default Docker Hub image (e.g. an internal registry mirror or a locally loaded offline image). `ContainerConfig.Image()` returns `image` as-is when it already carries a tag (so the separately-configured `tag` is dropped in that case), otherwise it appends `tag` (or `latest`); the default `localstack/<product>:<tag>` is used when `image` is unset.
 
 # Emulator Setup Commands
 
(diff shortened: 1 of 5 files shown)


carole-lavillonniere:
I would ship these 2 features in 2 PRs to keep them small and easy to review: 
- override of image 
- offline handling

For the latter, we are missing the most important integration tests that make sure the CLI works properly when offline.

gtsiolis:
Agree, will break in two PRs later. :loading:

gtsiolis:
Could you take another look, @carole-lavillonniere? 🍸

internal/config/containers.go
```diff
@@ -202,17 +207,32 @@ func (c *ContainerConfig) ResolvedEnv(namedEnvs map[string]map[string]string) ([
...
 
+// imageHasTag reports whether a Docker image reference already includes a tag.
+// A colon only counts as a tag separator when it appears in the final path
+// segment, so "my-registry:5000/localstack-pro" (registry port, no tag) is
+// correctly treated as untagged.
+func imageHasTag(image string) bool {
+	lastSegment := image[strings.LastIndex(image, "/")+1:]
+	return strings.Contains(lastSegment, ":")
+}
+
```
carole-lavillonniere, line 238: nit: can we use dep `github.com/distribution/reference ` instead of implementing ourselves? It would just need to be promoted to direct dependency.

carole-lavillonniere:
I think it would be important to write an integration test for the success path:
- image is set in config
- it is found locally and started
- no image is pulled at all
- no license check is done at all from CLI

carole-lavillonniere (commented):
edited

gtsiolis:
@carole-lavillonniere, thoughts?

_UPDATE_: Addressed https://github.com/localstack/lstk/pull/325#issuecomment-4799187476 in the separate PR in https://github.com/localstack/lstk/pull/337 following up from https://github.com/localstack/lstk/pull/325#issuecomment-4768144094.

CLAUDE.md
```diff
@@ -66,6 +66,8 @@ When adding a new command that depends on configuration, wire config initializat
 
 Created automatically on first run with defaults. Supports emulator types: `aws`, `snowflake`, and `azure`.
 
+Each `[[containers]]` block may set an optional `image` to override the default Docker Hub image (e.g. an internal registry mirror or a locally loaded offline image). `ContainerConfig.Image()` returns `image` as-is when it already carries a tag, otherwise it appends `tag` (or `latest`); the default `localstack/<product>:<tag>` is used when `image` is unset.
```
anisaoshafi: Can you also mention that value from `tag` is dropped when tag included in the image value?

anisaoshafi (approved):
Looks good! 
There is an integration test for when the image is not available, could you add an integration test for the happy path (image available)?

carole-lavillonniere (approved):

CLAUDE.md
gtsiolis: Added in 88b66d1.

gtsiolis:
Thanks, @anisaoshafi! The happy path test for image being available is covered in https://github.com/localstack/lstk/pull/337.

Merging once the checks pass. 🐌

PR #331, opened 2026-06-19
Add inspect command for snapshots

gtsiolis:
## What

Adds `lstk snapshot inspect <file>` — an offline, per-service size breakdown of a local `.snapshot` archive.

No running emulator, no platform call, no auth.

```
$ lstk snapshot inspect ./checkpoint.snapshot

  ~ Snapshot analysis for checkpoint.snapshot
    4.5 MB
  s3                 3.9 MB   87%
  dynamodb         172.4 KB    4%
  iam              107.4 KB    2%
  cloudfront        78.4 KB    2%
  ...
  ───────────────────────────────
  TOTAL              4.5 MB  100%
```

## Why

`snapshot show` is cloud-only and reports resource *counts* plus one aggregate size — it can't answer "what's eating the space in my local snapshot?". `inspect` reads the archive directly and attributes bytes per service, combining each service's control-plane state (`api_states/`) and data payloads (`assets/`), aggregated across accounts and regions.

## Details

- Opens the `.snapshot` ZIP with stdlib `archive/zip` and tallies sizes per entry — no emulator, platform, or auth.
- One row per service, sorted largest-first, with % of total and a grand total.
- `--json` emits the same data (raw bytes) for scripts and agents.
- Cloud (`pod:`) refs are rejected and pointed at `snapshot show`.
- New typed `SnapshotInspectedEvent` with plain + TUI rendering parity. Unit + integration tests; no Docker required.

Fix PRO-322.

| CURRENT inspect | NEW inspect |
|--------|--------|
| Screenshot 2026-06-19 at 14 35 50 | Screenshot 2026-06-19 at 14 59 17 |

### DIFF
--- a/cmd/snapshot.go
+++ b/cmd/snapshot.go
@@ -52,6 +52,13 @@ Only Cloud Pods (pod: prefix) have versions; local files and s3:// remotes do no
   lstk snapshot load pod:my-baseline:3
   lstk snapshot show pod:my-baseline:3`
 
+const snapshotInspectLong = `Inspect a local snapshot file, showing its structure and where the bytes go.
+
+This reads the on-disk .snapshot file directly — no running emulator, no platform call, and no authentication. It is the local counterpart to 'snapshot show', which reports metadata for cloud snapshots. Sizes are broken down per service (combining control-plane state and data assets), largest first.
+
+  lstk snapshot inspect ./checkpoint.snapshot   # size breakdown of a local snapshot
+  lstk snapshot inspect ./checkpoint --json      # machine-readable output for scripts and agents`
+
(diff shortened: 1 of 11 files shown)


carole-lavillonniere:
We are adding json output format for the first time here. What about removing it from here, but do an overall implementation later over all commands that return structured data? My biggest worry is that this becomes the contract we have to stick to as soon as we release it and we should be deliberate about how we format this json (case, structure etc).

gtsiolis:
Good observation! 😁

Absolutely, already tracked in [PRO-307](https://linear.app/localstack/issue/PRO-307/add-machine-readable-json-output) ([draft branch](<https://github.com/localstack/lstk/compare/george/add-json-output>)).

I will remove from this PR.

gtsiolis:

PR #332, opened 2026-06-19
Support mounting volumes and init hooks via config

anisaoshafi:
Closes DEVX-925.

Adds a per-container `volumes` list to `config.toml` for arbitrary bind mounts. The main use case being [Snowflake init hooks](https://docs.localstack.cloud/snowflake/capabilities/init-hooks/) (scripts mounted into `/etc/localstack/init/ready.d/` that run on startup).

```toml
[[containers]]
type = "snowflake"
port = "4566"
volumes = ["./init.sf.sql:/etc/localstack/init/ready.d/init.sf.sql"]
```

- Docker-style `"host:container[:ro]"` specs.
- Relative sources resolve against the config file's directory; `~/` is expanded (the Docker SDK treats a non-absolute source as a named volume, so this is required). Windows drive-letter sources (`C:\...`) are handled.
- Extra mounts must already exist (init-hook entries are files); only the persistence dir is auto-created.

## `volume` vs `volumes`

The singular `volume` and the plural `volumes` are **not** general synonyms; they overlap only for the persistence directory:

- `volume` *only* sets the persistent-state dir, always mounted to `/var/lib/localstack` (the dir managed by `lstk volume path` / `lstk volume clear`).
- `volumes` is a superset: it can mount arbitrary paths **and** set the persistence dir, via the entry whose container target is `/var/lib/localstack`.

So these are equivalent for persistence:

```toml
volume  = "/data"
# same persistence mount as
volumes = ["/data:/var/lib/localstack"]
```

Differences to keep in mind:

- `volume` cannot express init hooks or any non-persistence mount — use `volumes`.
- A `volumes` source is path-resolved (relative → config dir, `~/` expanded); the legacy `volume` value is used **verbatim**, so prefer an absolute path there.
- Declaring the persistence dir in **both** `volume` and a `volumes` entry with different sources is a configuration error; the same source in both is allowed.

`volume` stays supported for backward compatibility so existing `config.toml` don't break.


anisaoshafi:
@mmaureenliu would you like to give it a try and see how this works for snowflake? In my test it works, but wanted to double check I'm doing the right thing. Any feedback is welcome!

mmaureenliu:
> @mmaureenliu would you like to give it a try and see how this works for snowflake? In my test it works, but wanted to double check I'm doing the right thing. Any feedback is welcome!

Do I just build from this branch and run it with snowflake?

anisaoshafi:
> > @mmaureenliu would you like to give it a try and see how this works for snowflake? In my test it works, but wanted to double check I'm doing the right thing. Any feedback is welcome!
> 
> Do I just build from this branch and run it with snowflake?

Yes, please. 
Use `make build`; add the snowflake init hook in the volumes variable in `config.toml`; then start lstk via `./bin/lstk`

mmaureenliu:
> ```toml
> volumes = ["./init.sf.sql:/etc/localstack/init/ready.d/init.sf.sql"]
> ```

Yes it works beautifully! Can you trigger a new release after this so I can start updating snowflake docs?

anisaoshafi:
> > ```toml
> > volumes = ["./init.sf.sql:/etc/localstack/init/ready.d/init.sf.sql"]
> > ```
> 
> Yes it works beautifully! Can you trigger a new release after this so I can start updating snowflake docs?

Thanks for testing 🙌🏼  I will likely do it on Monday 29th - will ping you when done.

carole-lavillonniere (approved):
This looks great 👏 
One small suggestion: we could add an integration test to make sure mounting multiple volumes works (list supported)

anisaoshafi:
> This looks great 👏 One small suggestion: we could add an integration test to make sure mounting multiple volumes works (list supported)

Thanks @carole-lavillonniere, addressed in https://github.com/localstack/lstk/pull/332/commits/82abd0a9217d14626facf606e9e641996461e512 🙌🏼

PR #336, opened 2026-06-23
Add lstk az start-interception and stop-interception commands

carole-lavillonniere:
## Motivation

`lstk az` previously only ran the Azure CLI against an isolated `AZURE_CONFIG_DIR`, leaving global `~/.azure` untouched. The Azure team and TGTM discussion agreed `lstk az` should *also* offer azlocal's global pattern so existing `az` scripts run unmodified against LocalStack — and let the user choose.

## Changes

- Add `lstk az start-interception` — registers and activates the `LocalStack` cloud in the user's **global** `~/.azure` so plain `az` (any terminal/script) targets the LocalStack Azure emulator. Independent of `lstk setup azure`; does not touch global telemetry/survey prefs.
az-start
az-to-ls

- Add `lstk az stop-interception` — switches the active cloud back to `AzureCloud` (override with `--cloud <name>`, validated against the live `az cloud list`) and re-enables instance discovery. Guarded: only changes the cloud when `LocalStack` is still active, otherwise it reports and no-ops so it never clobbers an unrelated selection.
az-stop
az-top-fails

Closes COSY-761


anisaoshafi (approved):
Tested and LGTM 🚀

PR #337, opened 2026-06-23
feat: degrade gracefully in offline / enterprise environments

gtsiolis:
> **Stacked on #325.** This PR's base is the image-override branch, so its diff shows offline changes only; it auto-retargets to `main` when #325 merges. The local-image fallback integration test relies on the `image=` field introduced in #325.

## What

Makes `lstk start` degrade gracefully in offline / enterprise environments that cannot reach Docker Hub or the license server (offline/air-gapped, corporate proxy, or TLS interception) — without an explicit flag.

- **Image pull**: if `PullImage` fails but the image is already present locally (new `runtime.ImageExists`), lstk warns `"using the local image"` and starts the local image instead of failing.
- **License pre-flight**: `validateLicense` now distinguishes a definitive server rejection (`*api.LicenseError`, e.g. HTTP 403/400 — still fatal) from a transport-level failure (any other error — offline/proxy/cert). On a transport failure it skips the pre-flight check and lets the container validate its own bundled license at startup.
- `runtime.PullImage` always closes its `progress` channel, even when `ImagePull` fails early, so the local-image fallback path does not leak the progress goroutine.
- Context cancellation is propagated during the license pre-flight so Ctrl+C aborts cleanly.

## Scope

Second of the two PRs split out of DEVX-703. The **custom `image` override** half ships separately in #325. This PR is **offline graceful-degradation only**.

> Note: this addresses the *truly offline* case (network requests **fail**). The separate slow-network / demo case raised in #team-cli (a working-but-slow link where the pull *succeeds* but takes 20 min) is a different problem — graceful degradation never triggers there because nothing fails. That is being tracked as its own piece of work (`--offline` to skip the pull when the image is local + a cancellable pull).

## Tests

- Unit (`internal/container`): `TestPullImages_FallsBackToLocalImageWhenPullFails`, `TestPullImages_FailsWhenPullFailsAndImageMissing`, `TestValidateLicense_ContinuesWhenServerUnreachable`, `TestValidateLicense_FailsOnServerRejection`.
- Integration: local-image fallback and license-server-unreachable E2E coverage (added on this branch).

Refs DEVX-703

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -73,7 +73,8 @@ Each `[[containers]]` block may set an optional `image` to override the default 
 There is no `--offline` flag. Instead `container.Start` degrades gracefully when internet requests fail (the common enterprise blockers: Docker Hub unreachable, proxy/TLS interception, license server unreachable):
 
 - **Image pull**: if `rt.PullImage` fails but `rt.ImageExists` reports the image is already present locally, lstk warns and uses the local image instead of failing.
-- **License pre-flight**: `validateLicense` distinguishes a definitive server rejection (`*api.LicenseError`, e.g. HTTP 403/400 — still fatal) from a transport-level failure (any other error — offline/proxy/cert). On a transport failure it skips the pre-flight check and lets the container validate its own bundled license at startup.
+- **License pre-flight (image already local)**: when a pinned image is already present locally — so `pullImages` won't pull it — `tryPrePullLicenseValidation` skips the pre-flight check entirely (gated on `rt.ImageExists`), since the redundant network round-trip would otherwise block a fully-offline start; the container validates its own bundled license at startup. This is symmetric with the skip-pull behaviour above.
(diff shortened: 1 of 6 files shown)


gtsiolis:
Note to any reviewers, this is a stacked PR against https://github.com/localstack/lstk/pull/325. Cc @anisaoshafi

anisaoshafi (approved):
LGTM, in my test, it works well 🙌🏼 

the web app link is irrelevant for offline users, but super small nit:
Image

gtsiolis:
Thanks for taking a closer look, @anisaoshafi! 🙏

carole-lavillonniere (approved):
Pushed a commit f51aeaf114867330dac4457fe530e48ba6d77946 to:
- fix one comment
- document the feature in readme
- fix 2 integration tests that were using the actual home directory (would break parallelism and determinism of tests)

@gtsiolis please make sure this looks fine to you

gtsiolis:
Thanks, @carole-lavillonniere! 
- Merged #325, rebasing this one and waiting for checks. 🙏 
- Checked the docs commit and looks good. ✔️

PR #339, opened 2026-06-24
Reuse local image when already present for pinned tags

gtsiolis:
## Summary

`lstk start` ran `docker pull` on **every** start, even when the image was already present locally — `rt.PullImage` was unconditional. On slow networks this meant repeated multi-minute re-downloads for no benefit.

This adds `ImageExists` to the `Runtime` interface (backed by `ImageInspect` + `errdefs.IsNotFound` on Docker) and, in `pullImages`, skips the pull when a **pinned-tag** image is already local — emitting a `Using local image` message instead and recording `pulled[name]=false` so start telemetry (`LifecycleStartSuccess.Pulled`) doesn't over-count. Stopped-container cleanup (`rt.Remove`) still runs on both branches, and post-pull license validation is unaffected since the image is present.

A net-new probe is needed because `GetImageVersion` can't double as one: it errors both when the image is absent *and* when `LOCALSTACK_BUILD_VERSION` is missing, conflating the two.

## Scope

Skip-if-present applies to **pinned tags** only. Floating `latest`/empty still pulls until the `pull_policy` work lands (PRO-325).

## Tests

- Integration: pre-pull a pinned image, run `lstk start`, assert `Using local image` present and `Pulling` absent.
- Unit (MockRuntime): `ImageExists` true → 0 `PullImage` calls, `pulled[name]==false`; false → 1 call, `pulled[name]==true`.

| CLI output using local image |
|--------|
| Screenshot 2026-06-24 at 20 59 29 |

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -314,6 +314,20 @@ func pullImages(ctx context.Context, rt runtime.Runtime, sink output.Sink, tel *
 			return nil, fmt.Errorf("failed to remove existing container %s: %w", c.Name, err)
 		}
 
+		// Reuse a locally present image for pinned tags instead of re-pulling.
+		// Floating "latest"/empty tags always pull until pull_policy support lands.
+		if c.Tag != "" && c.Tag != "latest" {
+			exists, err := rt.ImageExists(ctx, c.Image)
+			if err != nil {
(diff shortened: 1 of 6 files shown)


anisaoshafi (approved):
Good improvement, nice thought. No need to pull images that are already downloaded 💯  Thanks!

PR #341, opened 2026-06-25
Add ESC to abort image pull with fall-back to local

gtsiolis:
Makes a floating-tag (`latest`/empty) image pull interruptible and resilient, so a slow pull no longer blocks the whole start with no way out. Implements [PRO-324](https://linear.app/localstack/issue/PRO-324/esc-to-abort-image-pull-with-fall-back-to-local); builds on [PRO-323](https://linear.app/localstack/issue/PRO-323) (`ImageExists`). Idea: @carole-lavillonniere .

## Behavior

- **ESC to bail.** When a newer version is downloading, a local copy exists, and we're interactive, the spinner shows `Pulling new version… (press esc to keep current version)`. ESC cancels only the pull and starts on the local image; Ctrl+C / `q` still quit.
- **Auto offline fall-back.** A pull that *fails* with a local copy present falls back to it (any mode): `Could not pull <image> (<err>); using the local image`.
- **Still fatal otherwise.** Pull fails + no local copy → `ErrorEvent` + silent error + telemetry, as before.

The hint only arms once Docker reports real layer download, never on a cache hit.

## How it works

`pullImages` checks `ImageExists` and delegates each pull to a new `pullImage`, which runs under a per-pull cancel context and `select`s on the pull result or a buffered skip channel. The domain stays UI-free: it emits `output.PullSkippableEvent` with a cancel channel it owns; the TUI binds ESC to signal it during the pulling phase. Gated on `localExists && interactive`.

## Tests

`start_test.go` (skip/use-local, fall-back, fatal, non-interactive, cancel-propagates) · `docker_test.go` (progress closes on immediate pull error) · `app_test.go` (ESC signals skip, inert otherwise) · `plain_sink_test.go` (event suppressed). PTY coverage skipped — both triggers are timing/registry-dependent and are covered by the deterministic unit/UI tests.

| Pulling a new version... | Using existing version... `ESC` |
|--------|--------|
| Screenshot 2026-06-29 at 18 23 25 | Screenshot 2026-06-29 at 18 23 28 |

### DIFF
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -20,7 +20,7 @@ const maxLines = 200
 
 // pullSkipHint replaces the "Pulling <image>" spinner text once a skippable pull
 // starts downloading, telling the user they can bail out to the local image.
-const pullSkipHint = "Pulling new version… (press esc to keep current version)"
+const pullSkipHint = "Pulling new version… (press ESC to keep current version)"
 
 type runDoneMsg struct{}
 
--- a/internal/ui/app.go
+++ b/internal/ui/app.go
@@ -18,6 +18,10 @@ import (
 
 const maxLines = 200
 
+// pullSkipHint replaces the "Pulling <image>" spinner text once a skippable pull
+// starts downloading, telling the user they can bail out to the local image.
+const pullSkipHint = "Pulling new version… (press esc to keep current version)"
+
 type runDoneMsg struct{}
 
 type runErrMsg struct {
@@ -55,6 +59,7 @@ type App struct {
 	width            int
 	cancel           func()
 	pendingInput     *output.UserInputRequestEvent
+	pullSkip         *output.PullSkippableEvent
 	err              error
 	quitting         bool
 	hideHeader       bool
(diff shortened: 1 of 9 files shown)


gtsiolis:
@carole-lavillonniere I've added you as reviewer since you originally proposed this[[1](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1782310808007789?thread_ts=1782146915.927329&cid=C0A1857LGTU)]! 😁

carole-lavillonniere (approved):
Nice! 🌟

internal/ui/app.go
```diff
@@ -18,6 +18,10 @@ import (
 
 const maxLines = 200
 
+// pullSkipHint replaces the "Pulling <image>" spinner text once a skippable pull
+// starts downloading, telling the user they can bail out to the local image.
+const pullSkipHint = "Pulling new version… (press esc to keep current version)"
```
anisaoshafi: tiny nit: probably uppercase `ESC` is more user friendly and stands out? Screenshot 2026-06-29 at 18 23 25

gtsiolis: We don't have a strong consistency around this yet, but sounds good.

@Linear can you make the change?

gtsiolis:
Thanks, both! Merging... 🌮 🌮 🌮

PR #342, opened 2026-06-25
fix: propagate os.UserHomeDir() errors in snapshot save and load

gtsiolis:
## Summary

- `runSnapshotSave` and `runSnapshotLoad` were silently ignoring the error from `os.UserHomeDir()` (`home, _ := os.UserHomeDir()`), passing an empty home path to `ParseSource`/`ParseDestination` when the home directory cannot be resolved
- `runSnapshotRemove` in the same file already handled this correctly — this brings the other two functions in line
- Fixes DEVX-938: https://linear.app/localstack/issue/DEVX-938/snapshot-saveload-silently-drop-osuserhomedir-errors

## Test plan

- [ ] `lstk snapshot save` and `lstk snapshot load` return a clear error when the home directory cannot be resolved, rather than silently computing a wrong path

### DIFF
--- a/cmd/snapshot.go
+++ b/cmd/snapshot.go
@@ -209,7 +209,10 @@ func runSnapshotLoad(cfg *env.Env, tel *telemetry.Client, logger log.Logger) fun
 			return err
 		}
 
-		home, _ := os.UserHomeDir()
+		home, err := os.UserHomeDir()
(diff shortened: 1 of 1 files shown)


anisaoshafi (approved):
Thanks George 🎣

PR #346, opened 2026-06-29
fix: exit non-zero on unknown subcommand of grouping parents

gtsiolis:
Parent commands that only group subcommands (`config`, `setup`, `volume`, `snapshot`) had no `RunE`, so Cobra printed help and **exited 0** for an unknown or missing subcommand — e.g. `lstk config bogus` and `lstk setup bogus` reported success despite being invalid usage. Unknown top-level commands and unknown flags already exited non-zero; the bug was specific to subcommands of these non-runnable grouping parents.

The fix adds a `requireSubcommand` helper (`cmd/root.go`) that sets `cobra.NoArgs` plus a help-printing `RunE` on these parents. Cobra skips arg validation on a non-runnable parent, so the `RunE` is the load-bearing part — it makes the command runnable so `NoArgs` actually rejects the stray subcommand. A bare invocation (`lstk config`) still prints help and exits 0, while an unknown subcommand now exits 1 with `Error: unknown command "bogus" for "lstk config"` on stderr.

Cobra's autogenerated `completion` command is the same shape — a non-runnable parent grouping `bash`/`zsh`/`fish`/`powershell` — so `lstk completion bogus` had the identical exit-0 bug. Since that command is created lazily, lstk now materializes it (`InitDefaultCompletionCmd`) and applies `requireSubcommand` to it as well.

The issue's literal example (`lstk start --bogus-flag-xyz`, an unknown flag) already exited non-zero; the regression test covers it alongside the parent-command cases to lock the behavior down.

Verified manually and with integration tests in `test/integration/exit_code_test.go`: unknown flag, unknown command, and an unknown subcommand of every grouping parent (`config`/`setup`/`volume`/`snapshot`/`completion`) → exit 1; bare parent → exit 0.

> [!NOTE]
> A sibling exit-0 bug found while investigating this — `lstk setup aws` exiting 0 when it can't write the profile — is split out to #349 (DEVX-944) to keep this PR focused on invalid-usage exit codes. The npm-launcher half of the report (the wrapper swallowing the binary's exit code) is handled separately in #347 (DEVX-942).


anisaoshafi (approved):
Thanks for fixing this 👏🏼 makes sense!

PR #307, opened 2026-06-15
CI: Enforce `docs` label in `lstk` Repository

aidehn:
## Changes + Motivation
- Adds and enforces the `docs` label in the `lstk` repository.
- You can also migrate the existing `check-release-label.yml` workflow to this new `enforce-labels.yml` workflow as well as update the `sync-labels.yml` to sync up the `semver` labels from the `localstack/meta` repository.

```
# enforce-labels.yml

...
  semver:
    runs-on: ubuntu-latest
    steps:
      - uses: mheap/github-action-required-labels@0ac283b4e65c1fb28ce6079dea5546ceca98ccbe # v5
        with:
          mode: exactly
          count: 1
          labels: "semver: patch, semver: minor, semver: major"

          # not required, just matches old messaging
          message: "PR must have a release label: 'semver: patch', 'semver: minor', or 'semver: major'"
          add_comment: true
```

- Opted not to for now since I saw there is some repository level protection rules which requires this exact check (didn't want to break anything) 

**_Feel free to merge this if it's ready!_**

PR #347, opened 2026-06-29
fix: forward signals to child process in npm launcher

gtsiolis:
The `@localstack/lstk` npm wrapper was auto-generated by `goreleaser-npm-publisher` and spawned the Go binary with no signal handlers. A programmatic `kill <lstk-pid>` killed the Node wrapper but left the Go binary running, orphaning mid-flight container starts. Interactive Ctrl-C only worked because the TTY signals the whole process group.

This ships our own launcher (`npm/launcher.js`) that resolves the platform binary from the installed optional dependency, forwards `SIGINT`/`SIGTERM`/`SIGHUP` to the child, and propagates the child's exit code / terminating signal.

Since that wrapper isn't customizable through the publishing action, the release job now runs `goreleaser-npm-publisher build` explicitly, overwrites the generated `dist/npm/lstk/index.js` with our launcher, then `npm publish`es each package — replacing the single `evg4b/goreleaser-npm-publisher-action` step. Platform packages are byte-identical to before; only the main package's wrapper changes.

A new `test-launcher` CI job runs `node --test npm/*.test.js`; the regression test spawns the launcher against a fake binary, sends `SIGTERM`, and asserts the child receives it and its exit code propagates out.

> [!NOTE]
> The publish pipeline change only runs on tagged releases, so it can't be exercised by PR CI — worth a close look from a reviewer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,6 +194,21 @@ jobs:
             exit 1
           fi
 
+  test-launcher:
+    name: Launcher Tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v6
+        with:
+          node-version: "20"
+
+      - name: Run launcher tests
+        run: node --test npm/*.test.js
+
   release:
     name: Build and Publish Release
     if: startsWith(github.ref, 'refs/tags/')
@@ -202,6 +217,7 @@ jobs:
       - goreleaser-check
       - test-unit
       - test-integration
+      - test-launcher
     runs-on: ubuntu-latest
     permissions:
       contents: write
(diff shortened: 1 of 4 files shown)


skyrpex (approved):
The npm tests failed to me when ran locally in macos, but pushed a commit to fix it.

It looks good to me. The only thing I don't like too much is relying a lot on external GitHub actions such as `goreleaser/goreleaser-action@v7`, since it makes it harder to run things locally because the logic is hidden elsewhere.

I also wanted to share that we should pin all GitHub actions to their version hashes in order to reinforce security, but we can do it in a different PR.

gtsiolis:
Thanks for pushing 3867288, @skyrpex!

+1 for pinning actions, I know @aidehn added some pinned actions in localstack/lstk#307.

@Linear add an issue in [CLI: Make `lstk` the default CLI](https://linear.app/localstack/project/cli-make-lstk-the-default-cli-592bbe079b81/overview) to pin GH actions for lstk.

PR #348, opened 2026-06-29
fix: run containers with --rm to avoid lingering exited containers

gtsiolis:
Containers were created without `--rm` (`AutoRemove=false`), so stopped containers stayed in `Exited` state and could linger indefinitely on mixed hosts. This sets `AutoRemove` on the `HostConfig` in `DockerRuntime.Start`, the Docker SDK equivalent of `--rm`, matching the Python CLI lifecycle.

The existing `Stop` already tolerates the container being gone (it ignores conflict/not-found from `ContainerRemove`), so it stays correct now that Docker may auto-remove the container on stop.

Adds an `auto remove` assertion to `TestStartCommandSetsUpContainerCorrectly` to guard against regression.

### DIFF
--- a/internal/container/logtail.go
+++ b/internal/container/logtail.go
@@ -1,0 +1,37 @@
+package container
+
+import "sync"
+
+// logTail is a thread-safe, byte-bounded buffer that retains the most recent
+// maxBytes written to it. It captures a container's startup logs while the
+// container runs so they survive the container being auto-removed (--rm) the
+// instant it exits — otherwise a crash during startup would leave no diagnostics.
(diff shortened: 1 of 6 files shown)


skyrpex:
@gtsiolis I added a commit that avoids the `logtail`'s buffer initial merging when writing to it by trimming first. Next time I'll raise my concern first so we can promote a collaborating environment!

skyrpex (approved):

gtsiolis:
Thanks for adding a37e5a4, @skyrpex! 🏀 Merging ...

PR #349, opened 2026-06-30
fix: exit non-zero when `setup aws` fails to write the profile

gtsiolis:
Closes DEVX-944.

`lstk setup aws` (and the deprecated `lstk config profile`) reported success and **exited 0** even when it failed to write the LocalStack AWS profile. Writing the profile is the command's whole purpose, but `awsconfig.Setup` emitted a warning and returned `nil` when the write itself failed (read-only `~/.aws`, permission denied, full disk, or a corrupt existing ini). The profile was never written, yet the process exited 0 — masking the failure from users, CI, and agents, so downstream `aws --profile localstack` calls then failed confusingly.

`Setup` now takes an `explicit` flag:

- **User-invoked commands** (`lstk setup aws`, `lstk config profile`) emit a structured `ErrorEvent` and return a `SilentError` on a write failure → exit 1.
- **Best-effort post-start path** (`EnsureProfile` during `lstk start`) keeps warning and continuing, so a profile-write hiccup never aborts an already-running emulator.

### Tests

`test/integration/awsconfig_test.go` adds a case that makes `~/.aws` read-only, confirms the prompt, and asserts a non-zero exit — it fails before this change and passes after. (Skipped under root, where directory permissions are bypassed.)

### Context

Split out of #346 (DEVX-941), which stays focused on invalid-usage exit codes. This is a sibling instance of the same "prints an error yet exits 0" class, in a different command.

### DIFF
--- a/internal/awsconfig/awsconfig.go
+++ b/internal/awsconfig/awsconfig.go
@@ -236,7 +236,7 @@ func EnsureProfile(ctx context.Context, sink output.Sink, interactive bool, reso
 		return nil
 	}
 	if interactive && !configOK && !credsOK {
-		return Setup(ctx, sink, resolvedHost, status)
+		return Setup(ctx, sink, resolvedHost, status, false)
 	}
 
 	emitMissingProfileNote(sink)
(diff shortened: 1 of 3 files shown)


anisaoshafi (approved):
Good catch 🐛

PR #350, opened 2026-06-30
Add non-interactive mode to `lstk setup aws`

gtsiolis:
## Motivation

`lstk setup aws` was interactive-only and rejected non-TTY / `--non-interactive` invocations with `setup aws requires an interactive terminal`, so the LocalStack AWS profile could not be written unattended in CI or agent workflows. Unlike `login`, the command only writes deterministic defaults, so the gate was a false rejection.

## Changes

- `cmd/setup.go`: branch on `isInteractiveMode` (mirroring `setup azure`) instead of rejecting non-interactive mode, and add a `--force` flag.
- Non-interactive `setup aws` now writes the profile with defaults and exits 0. An already-correct profile is a no-op; completing a partial profile needs no flag.
- Overwriting an existing `localstack` profile whose values differ requires `--force` (which also skips the Y/n prompt in interactive mode).
- Non-interactive write/check failures are returned as errors (non-zero exit); the interactive path keeps its existing warn-and-continue behavior (`DEVX-944` owns changing that).
- `awsconfig.SetupNonInteractive` (no `UserInputRequestEvent`; overwrite gate via the existing `checkProfileSetup`), `awsconfig.Setup` gains `skipConfirm`, the writes are extracted into a shared `applyProfile`, and host resolution is shared via `awsconfig.ResolveProfileHost`.

## Tests

- New integration tests: writes the profile non-interactively, idempotent re-run, overwrite-requires-`--force` (errors with profile untouched), and `--force` overwrites.
- Inverted the former `TestSetupAWSNonInteractiveReturnsError` (which asserted the old rejection).
- Verified the interactive PTY path is unchanged (confirm/decline), the full unit suite (902 tests), and `make lint` (0 issues).

Closes PRO-353


anisaoshafi (approved):
It worked well for me, thanks for improving UX 🧈

gtsiolis:
Thanks for testing, @anisaoshafi! Rebased to resolve conflicts and waiting for checks...

PR #353, opened 2026-07-01
fix: preserve persistence setting across `lstk restart`

gtsiolis:
`lstk restart` recreated the container using only the per-run `--persist` flag. Restarting an instance that was started with `--persist` (without re-passing the flag) silently dropped persistence, so `lstk status` showed "No resources deployed" and looked like data loss.

This treats persistence as a property of the running instance: when the caller doesn't explicitly pass `--persist`, `Restart` now inspects the currently running container and carries its persistence setting forward before stopping it. Explicitly passing `--persist` still works as before. Both the interactive and non-interactive restart paths go through `container.Restart`, so both are covered.

Added an integration test (`TestRestartCommandPreservesPersistenceWithoutFlag`) that starts with `--persist`, restarts without the flag, and asserts `LOCALSTACK_PERSISTENCE=1` is still set — it fails before this change and passes after.

### DIFF
--- a/internal/container/restart.go
+++ b/internal/container/restart.go
@@ -3,14 +3,37 @@ package container
 import (
 	"context"
 
+	"github.com/localstack/lstk/internal/config"
 	"github.com/localstack/lstk/internal/output"
 	"github.com/localstack/lstk/internal/runtime"
 )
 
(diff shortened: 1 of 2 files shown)


anisaoshafi (approved):
👏🏼

PR #355, opened 2026-07-02
Fix bash completion failing on stock macOS without bash-completion

gtsiolis:
## Motivation

On stock macOS (bash 3.2.57, no `bash-completion` package) the generated bash completion script failed on every Tab with `_get_comp_words_by_ref: command not found` and produced zero completions, even when loaded correctly. Cobra's script depends on that function from the bash-completion package on both of its init paths, and a stock Mac has neither. Zsh was unaffected because Cobra's zsh script is self-contained.

The `completion bash --help` text also recommended `source <(lstk completion bash)`, which is a silent no-op on bash 3.2.

## Changes

- Prepend a guarded pure-bash `_get_comp_words_by_ref` fallback to the generated script (`cmd/completion.go`), defined only when the bash-completion package is absent. This is the pattern git ships in `git-completion.bash`. It uses `COMP_LINE` to recover word adjacency so `--flag=value` and `pod:` shapes complete correctly, and stays bash 3.2 compatible.
- Replace the bash help text to recommend `eval "$(lstk completion bash)"`, which works on every bash, and document per-shell setup.
- Generate the fallback and Cobra's body against the writer resolved at execution time, so a `SetOut` after `NewRootCmd` does not split them across writers.

When the bash-completion package is installed, the guard skips the fallback and behavior is unchanged. Zsh, fish, and PowerShell output is byte-identical to before.

## Scope

This makes the generated script work when loaded. It does not enable completion by default, which still depends on the install method and shell. Auto-enabling completion during install is tracked separately.

## Tests

- 6 integration tests drive the real generated script in a bare bash (`--noprofile --norc`, no bash-completion): completion works without the package, wordbreak reassembly parity across 4 fixtures, and the fallback yielding to an installed package. All failed before the fix with the reported error.
- 2 unit tests pin the single-writer wiring and `--no-descriptions` passthrough.
- Verified on stock `/bin/bash` 3.2.57 and Homebrew bash 5.2 in a clean environment.

Closes DEVX-950


gtsiolis:

skyrpex:
Currently reviewing and testing 👍🏻 Will merge soon if all is good.

skyrpex:
@gtsiolis as far as I understand, this PR isn't enough to enable the autocompletion for `lstk` users in all scenarios, right? Depending on the installation method (brew or npm) and their shell of choice (fish, zsh or bash) they'll have to source or eval the completion anyways?

gtsiolis:
You're right, @skyrpex! This is only fixing the bug mentioned in the linked discussion in the issue.

More specifically, on stock macOS bash the generated script was crashing. This makes the script work when loaded, it doesn't turn completion on by default. 

Small nuance: `brew` ships the completion files, so `zsh`/`fish` auto-load there. `npm` (any shell) and `bash` are the manual cases. Updated the PR title and description to be more accurate.

The enablement piece you're describing is already tracked in [PRO-245](https://linear.app/localstack/issue/PRO-245/prompt-for-command-completions-during-install), which I opened some months ago to add a prompt for completions during install. That relies on [DEVX-877 ](https://linear.app/localstack/issue/DEVX-877/create-curl-install-script)(curl install script) as the natural place to wire it in. I'd keep those separate from this fix.

What do you think?

skyrpex:
Thanks, it makes perfect sense with that piece of context!

skyrpex (approved):

gtsiolis:
Pushed `f7548df` — a small 3-line fix for one more issue turning CI red here: under Go 1.26, an interrupted `lstk start` (Ctrl+C during the readiness wait) was recorded as a status-check failure instead of a clean cancellation, which made `TestStartEmitsTelemetryWhenInterruptedDuringStartup` flaky.

PR #360, opened 2026-07-02
Remove deprecated `lstk config profile` command

gtsiolis:
## What

Removes `lstk config profile`, a deprecated backward-compat alias for `lstk setup aws`.

## Why

`config profile` was introduced **already-deprecated** in #146, but the "Deprecated: use 'lstk setup aws' instead" notice lived only in `--help` — running it gave no runtime nudge (flagged by DevRel). It is **interactive-only** (hard-errors in non-interactive mode, so it can't appear in scripts or CI) and a **strict subset** of `setup aws`; both delegate to the same handler.

We considered DevRel's suggestion of Cobra's built-in `Deprecated` field but rejected it: it prints raw to `os.Stderr` (bypassing our `output.Sink` layer) and silently strips the command from `--help` **and** generated `lstk docs`. Removing the alias outright is cleaner.

## Changes

- Remove `newConfigProfileCmd`; simplify `newConfigCmd` wiring (drop the now-unused `env`/`ui` deps).
- Rename `ui.RunConfigProfile` → `ui.RunSetupAWS` (only `setup aws` calls it now; mirrors `ui.RunSetupAzure`).
- Drop the `config profile` mentions in `internal/awsconfig` and `CLAUDE.md`.
- Tests: migrate the decline-path test to `setup aws` (no equivalent existed); drop the confirm-path duplicate (identical to `TestSetupAWSCreatesAWSProfileWhenConfirmed`); add an exit-code assertion that `config profile` now exits non-zero with `unknown command "profile"`.

## Testing

- `make build`, `go vet -C test/integration ./...`, and unit tests for `cmd/`, `internal/ui/`, `internal/awsconfig/` — all pass.
- Integration subset (no Docker): `TestSetupAWSDoesNotCreateAWSProfileWhenDeclined`, `TestSetupAWSCreatesAWSProfileWhenConfirmed`, `TestInvalidUsageExitsNonZero` (incl. `removed_config_profile_subcommand`), `TestBareParentCommandExitsZero` — 18/18 pass.

Closes PRO-360


test/integration/env/env.go
```diff
@@ -23,6 +23,11 @@ const (
 	OtelEndpoint       Key = "OTEL_EXPORTER_OTLP_ENDPOINT"
 	AWSAccessKeyID     Key = "AWS_ACCESS_KEY_ID"
 	AWSSecretAccessKey Key = "AWS_SECRET_ACCESS_KEY"
+	// AzureCollectTelemetry is the env form of the Azure CLI's `core.collect_telemetry`
+	// config key. Tests that run `az` against a temp HOME set it to "false" so the CLI
+	// does not spawn its detached background telemetry-upload process, which on Windows
+	// outlives `az` and keeps a handle inside the temp dir — breaking t.TempDir() cleanup.
+	AzureCollectTelemetry Key = "AZURE_CORE_COLLECT_TELEMETRY"
```
anisaoshafi, line 29: what was the reason for the az related code changes in this PR, like this. seems like outside the scope? 

updated: actually entire commit [28774fa](https://github.com/localstack/lstk/pull/360/commits/28774fa32ef7923ad85f4d70225c298d3a96464d) seems outside the scope of this PR. Isn't this already handled in this newer [PR](https://github.com/localstack/lstk/pull/363/changes)? in that case can we remove this from here to keep things clean? 🙌🏼

CLAUDE.md
```diff
@@ -106,7 +106,6 @@ Use `lstk setup <emulator>` to set up CLI integration for an emulator type:
 - `lstk az start-interception` / `lstk az stop-interception` — Opt-in second mode: instead of the isolated dir, these mutate the user's **global** `~/.azure` so plain `az` (any terminal/script) targets LocalStack, then switch back. `start-interception` runs the same register → activate → `instance_discovery=false` → dummy-login flow against the global config (but does not touch global telemetry/survey prefs) and is independent of `lstk setup azure`. `stop-interception` switches the active cloud back to `AzureCloud` (override with `--cloud <name>`, validated against the live `az cloud list`) and re-enables instance discovery — but only if `LocalStack` is still the active cloud, to avoid clobbering an unrelated selection.
 
 This naming avoids AWS-specific "profile" terminology and uses a clear verb for mutation operations.
-The deprecated `lstk config profile` command still works but points users to `lstk setup aws`.
```
anisaoshafi, line 111: Do you think it's a good idea to add in CLAUDE.md the decision around not using Cobra's deprecated field, as explained in the PR description already?

anisaoshafi (approved):
Two small comments, otherwise LGTM!

CLAUDE.md
gtsiolis: Added one line in https://github.com/localstack/lstk/pull/366/commits/e4836b0469bdd3d7024cd14bf3478b6e0622feae.  Cc @anisaoshafi

PR #361, opened 2026-07-02
Wrap logs correctly at terminal width

jw2:
## Why

In `lstk logs --follow`, log lines wrapped narrower than the terminal. `HardWrap` counts characters, but styled prefixes (e.g. `styles.Secondary.Render("container | ")`) embed invisible ANSI escape sequences. Those bytes were counted as visible width, so wrapping fired before the visible content reached the terminal edge.

## What changes

- **ANSI-aware log wrapping.** `renderLogLine` now takes an `availableWidth` and a `continuationIndent`. It wraps the message on visible width, indents continuation lines to align under the message (past the prefix), and preserves the meta/message styling split (`" : "` boundary) across wrap points.
- **`app.go` LogLineEvent handler** measures the visible prefix width via `lipgloss.Width` and passes the remaining width to `renderLogLine`. Pre-wrapped lines are flagged (`styledLine.preWrapped`) so `View()` emits them verbatim instead of re-wrapping.
- **Tests**: unit coverage in `logrender_test.go` for plain wrap, visible-prefix width, meta/message styling across wraps, continuation indent, and zero-width passthrough; a golden-file test (`TestAppLogLineEventWrapsAtTerminalWidth`) for the end-to-end event path.

## Note on the test changes

`withTrueColorProfile` mutates lipgloss's **package-global** color profile. The tests using it were marked `t.Parallel()`, which is a race: a sibling's cleanup restoring the default profile mid-run strips the colors another test is asserting on, producing intermittent failures. Those tests are now serial, matching the convention already used by the inline profile-mutating tests in `app_test.go`.

## Testing

- `make test` — all green (`internal/ui` verified stable across 20 consecutive runs).


gtsiolis:
Swapped `semver: minor` → `semver: patch` here. Per [docs/RELEASING.md](../blob/main/docs/RELEASING.md), `minor` is for a *new* user-facing feature or command (new subcommand, new emulator), while bug fixes are `patch` even when the output is user-visible. This corrects existing `logs --follow` wrapping rather than adding a capability, so it lands in patch. Thanks for the fix!

openspec/changes/logs-follow-full-terminal-width/proposal.md
```diff
@@ -0,0 +1,22 @@
...
+
+- `logs-full-width-rendering`: Log lines in `--follow` mode wrap at the correct terminal width by accounting for invisible ANSI sequences in styled prefixes.
+
+### Modified Capabilities
+
+## Impact
+
+- `internal/ui/logrender.go` — `renderLogLine` signature or callers updated to accept a `width` parameter
+- `internal/ui/app.go` — `output.LogLineEvent` handler uses ANSI-aware wrapping
+- `internal/ui/wrap/` — may need a new ANSI-aware wrap helper
```
carole-lavillonniere: issue: can you remove this spec? We don't usually add any of those. Only Peter commited a couple when he worked on bigger features, but this is just a small bug fix.

carole-lavillonniere:
@jw2 I took the freedom of pushing a small refactor 0866a30.
It swaps the `preWrapped bool` for a `logLine *output.LogLineEvent` which is what we do for messages. This looked a bit weird, `preWrapped` looked like a workaround for wrapping too early in `Update()`.
Moving it to `View()` keeps layouting away from `Update() `.

carole-lavillonniere (approved):
Thanks for the fix @jw2 and your first contribution 👏

carole-lavillonniere:
@jw2 I also pushed 3ccee90. Your fix adding `withTrueColorProfile` was correct, but actually most of these unit test don't really need a color profile (only `TestRenderLogLineWrapsMetaAndMessage`), so removed it and ran them concurrently.

PR #357, opened 2026-07-02
Add `az` alias for `lstk setup azure`

gtsiolis:
## Motivation

`lstk setup az` failed with `unknown command "az" for "lstk setup"` even though the Azure tool-proxy command is literally `lstk az` (whose own not-set-up error points users at `lstk setup azure`). Azure was the only pairing where the setup name and the tool name diverge — `lstk aws` ↔ `lstk setup aws` match exactly, and `lstk terraform`/`lstk tf` set the precedent for tool-name aliases.

## Changes

- Add `az` as a Cobra alias on the `setup azure` command, mirroring the `terraform`/`tf` pattern; the help template renders `Aliases: azure, az` automatically
- Canonical name stays `azure` everywhere (help listing, docs, error strings); telemetry records the canonical `CommandPath`, so command analytics don't fragment
- Mention the alias in CLAUDE.md's setup commands section

## Tests

- New integration test `TestSetupAzureAliasAz`, red-green verified (failed with `unknown command "az"` before the fix); proves routing into the azure setup logic via its domain error, so it needs no Docker or az CLI and runs in parallel
- `make test` (924 unit tests) and `make lint` (both modules) clean

Closes PRO-356

PR #363, opened 2026-07-03
test: disable Azure CLI telemetry in integration test env

gtsiolis:
## Problem

`main` is red on the **Windows** integration-test job. `TestAzStopInterceptionNoOpWhenNotIntercepting` passes its assertions but then fails during `t.TempDir()` auto-cleanup:

```
testing.go:1464: TempDir RemoveAll cleanup: unlinkat
C:\Users\RUNNER~1\AppData\Local\Temp\TestAzStopInterceptionNoOpWhenNotIntercepting3716188435\001:
The process cannot access the file because it is being used by another process.
```

(Introduced by #357 — failing run: https://github.com/localstack/lstk/actions/runs/28601912948)

## Root cause

The test runs real `az` commands (via `lstk az stop-interception` and a raw `az cloud show`). When Azure CLI telemetry is enabled, `az` spawns a **detached background telemetry-upload process** that outlives the `az` invocation and keeps an open handle on the test's temp working directory (`t.TempDir()`, dir `\001`). When Go's `t.TempDir()` cleanup runs `RemoveAll` while that uploader is still alive, Windows refuses to delete a directory another process holds open. Unix removes directories with open handles, which is why only the Windows job flakes.

The codebase already disables this telemetry (`core.collect_telemetry=false`) for the *isolated* Azure config dir in `setLstkPrefs`, but `stop-interception` deliberately targets the *global* `~/.azure` and doesn't touch telemetry prefs — so under test it runs with telemetry on.

## Fix

Default `AZURE_CORE_COLLECT_TELEMETRY=false` in every test environment (`env.With` / `env.Without`), mirroring the existing `UnreachableAnalyticsEndpoint` guard that already keeps tests off our own analytics backend. It propagates to every `az` invocation (through `runLstk` → `az` and `runAzRaw`), so no uploader spawns → no lingering handle → cleanup succeeds. Explicit per-test overrides still win (exec dedups to the last value).

## Tests

Added two unit tests on the env helper, matching the existing analytics-endpoint pair:
- `TestBaseEnvDisablesAzureTelemetry` — the default is set by both constructors.
- `TestExplicitAzureTelemetryOverridesDefault` — an explicit value wins.

## Verification note

The failure is Windows-only (`RemoveAll` with an open handle is a no-op on Unix), so it can't be reproduced on macOS/Linux dev machines or the Linux CI shards. The new unit tests guard the mechanism (telemetry disabled in the base env); the Windows CI job on this PR is the end-to-end confirmation.

Fixes PRO-362.

### DIFF
--- a/test/integration/env/env.go
+++ b/test/integration/env/env.go
@@ -23,7 +23,18 @@ const (
 	OtelEndpoint       Key = "OTEL_EXPORTER_OTLP_ENDPOINT"
 	AWSAccessKeyID     Key = "AWS_ACCESS_KEY_ID"
 	AWSSecretAccessKey Key = "AWS_SECRET_ACCESS_KEY"
+	// AzureCollectTelemetry controls the Azure CLI's usage telemetry.
+	AzureCollectTelemetry Key = "AZURE_CORE_COLLECT_TELEMETRY"
 )
+
+// DisableAzureTelemetry is the value used to turn the Azure CLI's telemetry off in
+// every test environment. When telemetry is on, `az` spawns a detached background
+// uploader process that outlives the `az` invocation and keeps a handle on its
+// working directory / HOME — which for tests is a t.TempDir(). On Windows that open
+// handle makes t.TempDir()'s RemoveAll cleanup fail ("being used by another process").
+// Disabling it here prevents that flake and stops tests emitting to Azure's telemetry
+// backend, mirroring UnreachableAnalyticsEndpoint for our own analytics.
+const DisableAzureTelemetry = "false"
 
 // UnreachableAnalyticsEndpoint is a closed local port used as the default
 // analytics endpoint for every test environment, so the binary under test never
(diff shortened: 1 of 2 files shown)


anisaoshafi (approved):
Great finding on `AZURE_CORE_COLLECT_TELEMETRY`, thanks for fixing the flaky windows test 🥇

test/integration/env/env.go
```diff
@@ -23,8 +23,19 @@ const (
...
 )
 
+// DisableAzureTelemetry is the value used to turn the Azure CLI's telemetry off in
+// every test environment. When telemetry is on, `az` spawns a detached background
+// uploader process that outlives the `az` invocation and keeps a handle on its
+// working directory / HOME — which for tests is a t.TempDir(). On Windows that open
+// handle makes t.TempDir()'s RemoveAll cleanup fail ("being used by another process").
+// Disabling it here prevents that flake and stops tests emitting to Azure's telemetry
+// backend, mirroring UnreachableAnalyticsEndpoint for our own analytics.
+const DisableAzureTelemetry = "false"
```
anisaoshafi: Could you simplify this explanation? 
And potentially remove this variable - I don't see why it's needed when we can use `false` directly.

anisaoshafi:
@gtsiolis anything against merging this?

gtsiolis:
Thanks, @anisaoshafi! Rebased, addressed review comments, and merging.

PR #364, opened 2026-07-03
fix: guidance for port-conflict errors on 443 / service ports

gtsiolis:
## What

`lstk start` runs a pre-flight port check before starting the emulator. The **primary** edge port (4566) conflict path gave the user a next step (reconfigure the port); the **extra-ports** path (443 and the 4510–4559 service range) emitted a bare title + summary with **no actions at all**. Bart hit this on 443 (held by an unrelated `sshd`) and had no way to tell what was holding the port.

This routes both branches through the same `emitPortInUseError` helper so they render consistent guidance, and adds an OS-aware "identify the process" hint.

### Before (extra-port conflict)
```
✗ Port 443 is already in use
  LocalStack requires this port. Free it before starting.
```

### After (both 4566 and 443 / service ports)
```
✗ Port 443 already in use
  Another process is already using this port.
  ==> Identify the process using it: sudo lsof -i tcp:443
  ==> Or use another port in the configuration: …/config.toml
```

## Details

- `internal/container/start.go` — extra-ports branch now calls `emitPortInUseError` instead of emitting a bare `ErrorEvent`; the helper gained an "Identify the process using it" action.
- `internal/ports/ports.go` — new `InspectCommand(port)`: `lsof -i tcp:<port>` on macOS/Linux, `netstat -ano | findstr :<port>` on Windows. Privileged ports (<1024, e.g. 443) suggest **`sudo lsof`**, because `lsof` only lists other users' sockets under sudo — without it the hint returns nothing for a root-owned holder like `sshd` on 443, i.e. the reported case.
- Neutral summary ("Another process is already using this port.") — deliberately does **not** claim it's a stray LocalStack instance (that case is already handled upstream by `FindRunningByImage`, and Bart's was `sshd`).
- Tests: unit test for `inspectCommand` across OS/privileged-port branches; integration test `TestStartCommandFailsWhenExtraPortInUse` (binds 4510, since 443 is privileged and unbindable in a test).

Design choice, consistent with peers: this **suggests a command** rather than naming the process inline. Docker keeps its hot-path error generic and points to `netstat` in docs; ddev put process-naming in a separate opt-in `port-diagnose` command. Inline process-naming was intentionally kept out.

## Scope / follow-ups

This closes the concrete defect (a guidance-less extra-ports error). It does **not** yet cover the case where the pre-flight `localhost` dial passes but Docker fails to bind at container start — that raw `Bind for …: port is already allocated` still surfaces with no guidance and is mis-bucketed as `start_failed`. That path (which a privileged-port / non-loopback-bindHost / TOCTOU conflict routes into) is the harder half and is tracked as a follow-up; **this PR's green check does not mean that half is done.**

Refs PRO-363.

### DIFF
--- a/internal/container/port_conflict_test.go
+++ b/internal/container/port_conflict_test.go
@@ -326,3 +326,14 @@ func TestPortConflictActions(t *testing.T) {
 	assert.Empty(t, portConflictActions(runtime.FlavorDockerDesktop, runtime.FlavorUnknown, "443"))
 	assert.Empty(t, portConflictActions(runtime.FlavorRancherDesktop, runtime.FlavorUnknown, "8443"))
 }
+
+func TestEmitPortInUseErrorOmitsConfigForFixedPort(t *testing.T) {
+	sink := &recordingSink{}
+
+	emitPortInUseError(sink, "4510", false)
+
+	errs := sink.errorEvents()
+	require.Len(t, errs, 1)
+	require.Len(t, errs[0].Actions, 1)
+	assert.Equal(t, "Identify the process using it:", errs[0].Actions[0].Label)
+}
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -908,11 +908,7 @@ func selectContainersToStart(ctx context.Context, rt runtime.Runtime, sink outpu
 			}
 		}
 		if conflictPort, err := ports.CheckAvailable(requiredSpecs...); err != nil {
-			sink.Emit(output.ErrorEvent{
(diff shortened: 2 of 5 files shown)


gtsiolis:
Closing as superseded by the newer startup behavior that treats the default HTTPS port as optional and continues with guidance when it is unavailable. For required port conflicts, we’ll keep the startup error concise rather than introduce OS-specific process-inspection commands. We can also revisit an opt-in lstk doctor diagnostic if this becomes a recurring support issue.

PR #365, opened 2026-07-03
fix: show invoked command form for aliased cmds 'lstk save/load --help'

anisaoshafi:
Examples for aliased cmds "lstk save/load --help" were not showing the actual invoked cmd, but the full form of the corresponding cmd (`lstk snapshot save/load` instead of `lstk save/load`).
| Before | After |
|---|--|
 | image | image |
| image |  image |


gtsiolis (approved):
Great use of BEFORE/AFTER table! 💯 

LGTM. ✔️

PR #366, opened 2026-07-04
Restructure agent instructions and skills

gtsiolis:
## Motivation

The agent-facing instructions had drifted from the code (nonexistent APIs in skills, an incomplete architecture map) and the root CLAUDE.md had grown past 4k words of always-loaded context, much of it deep per-feature reference.

## Changes

- **Root CLAUDE.md goes from 4,235 to 3,158 words (-25%)**: ~1,650 words of deep reference moved into nested files (loaded only when working in those packages) while ~570 words of new conventions were added
- Restructure CLAUDE.md: complete the `internal/` package map and command groups, correct the `test-integration` and mock-generation claims (mockgen via `go:generate`), and add sections for commit/PR/Linear conventions, the release process, spec-driven (openspec) changes, and human-facing writing style, plus notes on the single-test/lint/JUnit workflows
- Move deep reference sections (snapshots, Azure setup/interception, volume mounts, GATEWAY_LISTEN + offline fallbacks) into nested CLAUDE.md files under `internal/{snapshot,azureconfig,config,container}/`, each with an `AGENTS.md` symlink mirroring the root convention
- Refresh skills: fix `add-command` (the `sink.Emit(output.XxxEvent{...})` pattern instead of nonexistent `output.EmitXxx` helpers, `initConfig(nil)`, Cobra group registration, `requireSubcommand`, `isInteractiveMode`, `testEnvWithHome`), fix the `review-pr` events checklist (`sealedEvent()` marker, no emit helper), update `create-pr` (current Linear MCP tool name, Linear-first ticket step), and refresh `add-component` (`SuccessColor` palette constant)
- Make `test-integration` depend on `build` (one-line Makefile change) so the binary is always rebuilt before integration tests — this is what makes the updated build docs accurate; previously a stale `bin/lstk` could silently run after source edits

Closes PRO-365

### DIFF
--- a/.claude/skills/add-command/SKILL.md
+++ b/.claude/skills/add-command/SKILL.md
@@ -16,7 +16,7 @@ Before writing any code, understand what the command should do and whether the a
 **Core questions:**
 1. **What does this command do?** (one sentence — e.g., "shows the status of running emulators")
 2. **Does it need to talk to Docker/the runtime?** (determines whether `runtime.Runtime` is a dependency)
-3. **Does it need configuration?** (determines whether `PreRunE: initConfig` is needed)
+3. **Does it need configuration?** (determines whether `PreRunE: initConfig(nil)` is needed)
 4. **Does it need authentication?** (determines whether auth flow is involved)
 5. **Does it need any new event types?** (e.g., a new kind of progress, a new status phase — if yes, use `/add-event` for each)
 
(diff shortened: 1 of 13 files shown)


skyrpex (approved):

gtsiolis:
Thanks for taking a look, @skyrpex! Expecting the separate agent instructions in sub-folders to make better use of context, but let's revert if this does not work as expected. Cc @anisaoshafi

gtsiolis:
Thanks for merging, @skyrpex! 🏀

---

skyrpex:
@gtsiolis I've seen this technique being mentioned a few times as well, also heard it works well hence my approval. Let's keep an eye on it! Thank you.

PR #367, opened 2026-07-05
Add global --json flag to `lstk`, so commands can support JSON output

peter-smith-phd:
## Summary

- Adds a global `--json` flag for `lstk` that's made available to all built-in commands, as well as `lstk` extensions.
- Built-in commands without JSON support will reject `--json` with an explicit error message. This ensures that nobody expects JSON output from a command that doesn't support it.
- Note the special case for proxy commands (`aws`, `terraform`, etc) where `--json` after the sub-command name is forwarded directly to the proxied command, since those tools often support `--json` in their own way. Only `--json` that appears _before_ the subcommand will be handled by `lstk`.
- At this time, there are no commands that actually support `--json`, but the ability is now available.

Fixes PRO-361.


cmd/root.go
```diff
@@ -364,6 +371,45 @@ func instrumentCommands(cmd *cobra.Command, tel *telemetry.Client) {
 	}
 }
 
+// requireJSONSupport walks the Cobra command tree and wraps every RunE so that,
+// when cfg.JSON is set, a command lacking the jsonSupportedAnnotation is
+// rejected instead of silently rendering plain-text output. It reuses the same
+// root/extension-dispatch carve-out instrumentCommands has (a positional arg on
+// the root command means extension dispatch, which owns its own output format
+// and is never gated).
+func requireJSONSupport(cmd *cobra.Command, cfg *env.Env) {
```
anisaoshafi, line 400: `requireJSONSupport` has some overlap with existing func `instrumentCommands`. Perhaps consider moving the common logic to another function if possible to avoid maintenance issues in the future?

anisaoshafi (approved):
LGTM, thanks for keeping the scope small on purpose 💯

PR #373, opened 2026-07-08
fix: reject directory paths early in snapshot load

anisaoshafi:
lstk snapshot load accepted a directory path and let it fall through to the upload code path, where it failed with a confusing HTTP-layer error. Now it returns a clear "<path>" is a directory, expected a snapshot file error when no file candidate resolves instead of silently treating a directory (e.g. . or ~/) as a valid local snapshot source.

| Before | After |
|---|---|
| imagen | imagen |


skyrpex (approved):
Looks good to me! Thanks.

PR #374, opened 2026-07-08
Proposal for `lstk` JSON output schema

peter-smith-phd:
## Summary

This PR is primarily a **review of the `lstk` JSON output schema design**. The main thing to review is `docs/structured-output.md`, containing the detailed specification of how the JSON output from `lstk` should be structured. This is a proposal, but as soon as we solidify the design it'll be locked in (e.g. we can't iterate without changing the schema version number).

This includes:
* The overall payload structure, with lots of examples.
* How errors are reported (the "error" object)
* The various error codes that give us programmatic access (rather than textual) for error situations.

To prove this all works, three commands (`stop`, `reset`, `update`) are wired up end-to-end with `--json` as a pilot. These were chosen as simple commands that could exercise the new schema, and give a feel for how it looks.

Additionally, the docs/structured-output.md` **proposes** but doesn't solidify the details for all the other commands. We can revisit these schemas when we actually add JSON support for each command (in later PRs).

If you do nothing else, please read `docs/structured-output.md`


skyrpex (approved):
Looks good to me, specially the logs streaming using NDJSON :) Nice job.

internal/output/envelope_sink.go
```diff
@@ -0,0 +1,151 @@
...
+	case DeferredEvent:
+		s.Emit(e.Inner)
+	case ErrorEvent:
+		s.setError(e)
+	case MessageEvent:
+		if e.Severity == SeverityWarning {
+			s.warnings = append(s.warnings, Warning{Code: warningCodeNotice, Message: e.Text})
+		}
+	case EmulatorStoppedEvent:
+		s.appendEmulator(map[string]any{"type": e.Type, "name": e.Name, "wasRunning": e.WasRunning})
```
carole-lavillonniere: suggestion: here we use string literals that are repeated for each command. This looks quite fragile and this is what I was worried about when I said:
> don't rely on discipline for the data to be consistent. For example, an attribute should always have the same name across all commands. Can we make sure the data model structs are shared across all commands to prevent such accident? Should we have some kind of a convention doc in the repo? Or is there a spec format that we could use to document this?

Can we instead have structs that are shared across commands? So an object emulator would always look the same, across all commands.


It could look something like this:
```go

// EmulatorRef identifies an emulator in JSON `data` payloads. Shared by every
// command that names an emulator (stop, start, restart, status, reset, ...) so
// "type"/"name" can't drift apart or be typo'd per command.
type EmulatorRef struct {
    Type string `json:"type"`
    Name string `json:"name"`
}

// StoppedEmulator is the per-emulator entry in `stop`'s data.emulators.
type StoppedEmulator struct {
    EmulatorRef
    WasRunning bool `json:"wasRunning"`
}

and line 39 becomes:

case EmulatorStoppedEvent:
    s.appendEmulator(StoppedEmulator{
        EmulatorRef: EmulatorRef{Type: e.Type, Name: e.Name},
        WasRunning:  e.WasRunning,
    })

The payoff shows up at the next command: start's entry reuses the same EmulatorRef rather than re-typing "type"/"name":

type StartedEmulator struct {
    EmulatorRef
    Host           string `json:"host"`
    Version        string `json:"version"`
    AlreadyRunning bool   `json:"alreadyRunning"`
    Persist        bool   `json:"persist"`
}
```

docs/structured-output.md
```diff
@@ -0,0 +1,573 @@
...
+  "status": "ok",       // "ok" | "error"
+  "data": { /* ... */ }, // non-null iff status is "ok"
+  "warnings": [],        // always an array, possibly empty
+  "error": null          // non-null iff status is "error"
+}
+```
+
+| Field | Type | Notes |
+|---|---|---|
+| `schemaVersion` | integer | The envelope's wire-format version, currently always `1`. Not versioned per-command — bumped only on a breaking change to the envelope itself or to a specific command's `data` shape (additive fields never require a bump). A script should check it once: `if (resp.schemaVersion !== 1) { /* bail or adapt */ }`. |
```
gtsiolis, line 31: +1 for a centrally defined envelope. From [relevant comment](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1783940467078179?thread_ts=1782995878.637949&cid=C0A1857LGTU):

> @gtsiolis: One thing I’d still consider is whether the envelope and each command’s data should evolve independently, so a change to one command doesn’t invalidate consumers of every other command

openspec/changes/json-output-schema/design.md
```diff
@@ -0,0 +1,365 @@
...
+- **Considered and declined: sparse fieldset selection (`gh`-style `--json field1,field2`).** Letting a caller pick which fields of `data` to receive keeps payloads small and makes adding fields always backward-compatible. Declined for v1: lstk's payloads are small relative to `gh`'s (nothing like a hundred-row PR list), and `schemaVersion` already provides an evolution path, so the added flag-parsing surface isn't worth it yet. Revisit if a command's `data` shape grows large enough that most callers only want a slice of it.
+- **`-v`/`--version` gaining JSON output.** A permanent gap, not a deferred one — see the "Decision: `-v`/`--version` stays JSON-incapable" below for why.
+- **How a future second output format is selected on the command line** (e.g. whether `--json` itself changes shape). Explicitly deferred — not addressed by this change at all, beyond noting that the new type/capability names here (`EnvelopeSink`, `EnvelopeError`, `EnvelopeAction`, `output-envelope`, `error-codes`) don't depend on the answer.
```
gtsiolis, line 42: **thought**: Agreed on `--json` for now. From [relevant discussion](https://localstack-cloud.slack.com/archives/C0A1857LGTU/p1783940467544839?thread_ts=1782995878.637949&cid=C0A1857LGTU):

> @gtsiolis: [...] NDJSON could be useful for long-running commands such as logs --follow.

anisaoshafi (commented):
**issue:** this doesn't work as expected if sth goes wrong while loading the config. Because config validation is done before the JSON code is reached, JSON output is skipped.

It doesn't respect the JSON schema, even thought this error is theoretically mapped.
Image

anisaoshafi:
another **issue**: when lstk returns an error, it normally has a short headline ("Docker is not available") and the actual diagnostic/details linked with it. In plain text mode we get both. 
In `--json` mode, only the headline survives and the real reason gets thrown away before it reaches the JSON.  This is the case for start, stop, reset, but scope is likely broader.
imagen

Could you please make sure the detailed error it is also included in the `--json` schema? 



_Updated_
On a similar note, when `lstk update --json` returns an error it show only "update failed: exit status 1" but the real reason (e.g. brew/npm's actual error output) is captured for plain text but silently dropped for `--json` flag.
This is a simulation of a brew fail:
imagen

internal/output/envelope_sink.go
peter-smith-phd: Thanks, that's a great idea. Unfortunately, my lack of Golang expertise is what caused that to slip in. I'll make sure that type-safe approach is added.

docs/structured-output.md
peter-smith-phd: Yes, but within reason. It wouldn't be a problem that have some commands on v2 of the schema, while the rest are on v1. However, if we arbitrarily allowed commands to evolve in their own direction, that would completely invalidate the purpose of having a standard envelope.

No code changes required today (just doc changes), but it's something to be considered in future. These types of changes must be done very carefully, since even a single command changing schema can break a lot of scripts.

peter-smith-phd:
> **issue:** this doesn't work as expected if sth goes wrong while loading the config. Because config validation is done before the JSON code is reached, JSON output is skipped.

Thanks, fixed.

peter-smith-phd:
> **issue:** this doesn't work as expected if sth goes wrong while loading the config. Because config validation is done before the JSON code is reached, JSON output is skipped.

Good catch, I've now fixed it.

Note: I ran into a similar problem where `lstk --version --json` doesn't output JSON, but that's because Cobra automatically handles the `--version` flag, and we aren't able to reformat the output. At least, I couldn't find a way.

peter-smith-phd:
> another **issue**: when lstk returns an error, it normally has a short headline ("Docker is not available") and the actual diagnostic/details linked with it. In plain text mode we get both. In `--json` mode, only the headline survives and the real reason gets thrown away before it reaches the JSON. 

Thanks this is now fixed.

> _Updated_ On a similar note, when `lstk update --json` returns an error it show only "update failed: exit status 1" but the real reason (e.g. brew/npm's actual error output) is captured for plain text but silently dropped for `--json` flag. 

This will be a different fix from the first issue, because it involves capturing the raw output from the underlying npm/brew command, and inserting it into the JSON payload.

Given the sensitivity of the `update` mechanism, I plan to fix this in a separate PR. I don't want to risk leaving the customers with an `lstk` binary that has a broken `update` mechanism. I've created a [new ticket](https://linear.app/localstack/issue/DPX-604/capture-output-from-lstk-update-process-making-it-available-in-json), and plan to test it carefully.

PR #375, opened 2026-07-09
test: stop login tests from opening real browser tabs

gtsiolis:
## Problem

The device-flow login opens the auth URL in a real browser (`browser.OpenURL` in `internal/auth/login.go`). The UI login tests and the integration login tests exercise the *real* login flow against an `httptest.Server`, so running `make test` / `make test-integration` on a machine with a GUI spawned real browser tabs (three from the unit tests, two from the integration tests). It's invisible in CI because headless runners have no `open`/`xdg-open` target, so the open just fails silently.

Worse than a nuisance: the auth URL is built from the web app URL, not the API endpoint, and the integration tests only overrode `LSTK_API_ENDPOINT` — so the two integration-test tabs pointed at the **production** `https://app.localstack.cloud/auth/request/…` with synthetic auth-request IDs.

This is a test-side issue only — auto-opening the browser on a real `lstk login` is intended (DEVX-530).

## Changes

No change to login behavior; the only production-code change is a test seam.

- **`internal/auth`**: `loginProvider` gains an injectable `openBrowser func(string) error` field (defaulting to the existing `browser.OpenURL` call, moved into `defaultOpenBrowser`). `auth.New` accepts variadic `Option`s, with `WithBrowserOpener` to override the opener.
- **TUI login tests** (`internal/ui/run_login_test.go`): all three device-flow tests inject a `browserRecorder` via `auth.WithBrowserOpener` and now assert the exact URL that would have been opened.
- **Integration tests** (`test/integration/login_test.go`): the real binary can't take a Go func, so a new `fakeBrowserOpener` helper prepends a temp dir to `PATH` with fake `open`/`xdg-open`/`x-www-browser`/`www-browser` scripts that record the URL to a file instead of spawning a tab. Both device-flow tests assert the recorded URL and set `LSTK_WEB_APP_URL` to the mock server (alongside the existing `LSTK_API_ENDPOINT` override), making them fully hermetic — nothing references production anymore.
- **`test/integration/env`**: new `Path` and `WebAppURL` env keys.

Both assertions fail before the fix (no URL recorded / real tab) and pass after.

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -21,10 +21,25 @@ type Auth struct {
 	licenseFilePath string
 }
 
-func New(sink output.Sink, platform api.PlatformAPI, storage AuthTokenStorage, authToken, webAppURL string, allowLogin bool, licenseFilePath string) *Auth {
+// Option configures optional behavior of the login flow.
+type Option func(*loginProvider)
+
+// WithBrowserOpener overrides how the login flow opens the auth URL in a
+// browser. Tests inject a recorder so they don't spawn real browser tabs.
+func WithBrowserOpener(open func(string) error) Option {
+	return func(lp *loginProvider) {
+		lp.openBrowser = open
+	}
+}
+
+func New(sink output.Sink, platform api.PlatformAPI, storage AuthTokenStorage, authToken, webAppURL string, allowLogin bool, licenseFilePath string, opts ...Option) *Auth {
+	lp := newLoginProvider(sink, platform, webAppURL)
+	for _, opt := range opts {
+		opt(lp)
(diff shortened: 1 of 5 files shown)


anisaoshafi (approved):
LGTM! 

<s> @gtsiolis have you run all the integration tests in local to verify the fix is comprehensive (no browser window opens)? </s>
_update:_ All good 🚀

gtsiolis:

---

gtsiolis:
Thanks for checking[[1](https://github.com/localstack/lstk/pull/375#pullrequestreview-4694009879)], @anisaoshafi!

PR #161, opened 2026-03-23
Forward host environment variables to the emulator container

gtsiolis:
Forwards two sets of host environment variables into the LocalStack container at startup, matching the behaviour of the legacy CLI:
- `CI`: ensures LocalStack behaves correctly in CI environments.
- `LOCALSTACK_*`: forwards all LocalStack configuration variables set on the host. LOCALSTACK_AUTH_TOKEN is excluded since it is already explicitly injected.

PR #378, opened 2026-07-09
fix: don't let LOCALSTACK_PATH clobber emulator PATH

gtsiolis:
## Motivation

Reported in DEVX-984 (Slack, patrick.lafond): `lstk start` died with `/usr/local/bin/docker-entrypoint.sh: line 22: mkdir: command not found` on any config, tag, and lstk version, but only on one machine. **Root cause since confirmed by the reporter**: a `LOCALSTACK_PATH` var exported in his zsh config (pointing agents at a localstack source checkout); `env -u LOCALSTACK_PATH lstk` starts cleanly.

Why it breaks: since #161 lstk forwards every host `LOCALSTACK_*` env var into the emulator. The emulator image's `docker-entrypoint.sh` strips the `LOCALSTACK_` prefix and re-exports the remainder (excluded: `LOCALSTACK_HOST*` and names whose stripped remainder starts with a digit), so `LOCALSTACK_PATH=/home/user/repos/localstack` becomes `export PATH=/home/user/repos/localstack` inside the emulator. The entrypoint itself is started by absolute path, so the first command that needs `PATH` is `mkdir` — exactly the reported failure — and `set -eo pipefail` kills startup.

## Changes

- `filterHostEnv` drops `LOCALSTACK_*` vars whose prefix-stripped name collides with a critical runtime variable: `PATH`, `HOME`, `IFS`, `BASH_ENV`, `LD_PRELOAD`, `LD_LIBRARY_PATH`, `PYTHONPATH`, `PYTHONHOME` (`criticalContainerVar`). Every entry has a failure mode verified against the actual image: `HOME` silently redirects every `~`-anchored path (plugin cache, `~/.localstack`, `~/.aws`), `BASH_ENV` names a file every non-interactive bash executes (init hooks inherit it), `LD_*` hijack the dynamic loader for every process, `PYTHONHOME` is fatal to the interpreter, `PYTHONPATH` shadows emulator imports, and `IFS` corrupts word splitting in the shell sourcing the re-exports. Matching is on the exact stripped name, so near misses like `LOCALSTACK_PATHFINDER`, `LOCALSTACK_SHELL`, or `LOCALSTACK_ENV` still forward, and `LOCALSTACK_HOSTNAME` still forwards since the entrypoint excludes it from stripping.
- `filterHostEnv` also drops forwarded values containing a newline/CR: the entrypoint re-exports vars through a line-oriented `env | sed` pipeline, so an embedded line like `LOCALSTACK_PATH=` injects a rogue `export PATH=` that blanks `PATH` — a second, independent way to hit the same crash (adopted from #379).
- Dropped vars are returned as `droppedHostEnv{name, overrides}` and `Start` emits a warning for each, with the reason (`…it would override PATH inside the emulator` / `…its value contains line breaks`), so neither drop is silent. The long-standing `LOCALSTACK_AUTH_TOKEN` drop stays silent as before.

## Testing

- Extended `TestFilterHostEnv` unit test covering both drop reasons, the reported drop metadata, and the names that must keep forwarding (fails before the fix).
- Extended the `TestStartCommandPassesCIAndLocalStackEnvVars` integration test to start with `LOCALSTACK_PATH` set: before the fix the emulator dies exactly as reported; after it, start succeeds, the warning is printed, and the var is absent from the container env. Needs Docker + `LOCALSTACK_AUTH_TOKEN`, so it runs in CI.

Supersedes #379 (its multiline guard and wider reserved set are folded in here, with warnings added on top).

A follow-up worth filing separately: excluding `PATH` and friends from prefix-stripping in the image entrypoint itself, which would also protect compose users who export `LOCALSTACK_PATH` without lstk involved.

Fixes DEVX-984

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -82,11 +82,12 @@ func Start(ctx context.Context, rt runtime.Runtime, sink output.Sink, opts Start
 	tel := opts.Telemetry
 
 	hostEnv, droppedEnv := filterHostEnv(os.Environ())
-	for _, name := range droppedEnv {
-		sink.Emit(output.MessageEvent{
-			Severity: output.SeverityWarning,
-			Text:     fmt.Sprintf("Not forwarding %s to the emulator: it would override %s inside the emulator", name, strings.TrimPrefix(name, "LOCALSTACK_")),
-		})
+	for _, d := range droppedEnv {
+		text := fmt.Sprintf("Not forwarding %s to the emulator: it would override %s inside the emulator", d.name, d.overrides)
+		if d.overrides == "" {
+			text = fmt.Sprintf("Not forwarding %s to the emulator: its value contains line breaks", d.name)
+		}
+		sink.Emit(output.MessageEvent{Severity: output.SeverityWarning, Text: text})
 	}
 	agentEnvVars := agentEnv(caller.New().Classify())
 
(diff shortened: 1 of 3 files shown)


gtsiolis:
I'd appreciate a check from @anisaoshafi before merging. 🏓

gtsiolis:

internal/container/start.go
```diff
@@ -874,19 +881,64 @@ func awaitStartup(ctx context.Context, rt runtime.Runtime, sink output.Sink, con
...
+// would break the emulator: the image's docker-entrypoint.sh strips the
+// LOCALSTACK_ prefix and re-exports the remainder (only LOCALSTACK_HOST and
+// LOCALSTACK_HOSTNAME are excluded), so e.g. a host LOCALSTACK_PATH becomes
+// PATH inside the emulator and startup dies with "mkdir: command not found"
+// (DEVX-984). IFS is included because the entrypoint sources the stripped
+// exports mid-script, corrupting its own word splitting.
+func criticalContainerVar(name string) bool {
+	switch name {
+	case "PATH", "HOME", "SHELL", "IFS", "ENV", "BASH_ENV",
+		"LD_PRELOAD", "LD_LIBRARY_PATH", "PYTHONPATH", "PYTHONHOME":
```
anisaoshafi: Do you have evidence all of these variables need to be skipped? Some of them make sense, but not sure what is the effect of having `HOME, SHELL, ENV, BASH_ENV`? Just wondering what these variables do

anisaoshafi (approved):
Thanks for catching this weird behaviour 🥇

internal/container/start.go
gtsiolis: Thanks for taking a closer look, @anisaoshafi! Did some more investigation...

- Removed `SHELL` as nothing in the image reads `SHELL`.
- Removed `ENV` as it's only read by interactive shells, which never inherit the re-exported env.
- Kept `HOME`, which is what `LOCALSTACK_HOME` becomes for the whole emulator process tree.
- Kept `BASH_ENV`, which names a file every non-interactive bash executes, and init hooks inherit that env.

anisaoshafi: Thanks for checking them!

gtsiolis:

---

PR #379, opened 2026-07-09
fix: don't forward host LOCALSTACK_* env that clobbers container PATH

gtsiolis:
> **Note:** This overlaps heavily with the pre-existing draft #378 (same root cause, same `filterHostEnv` blocklist approach). Opened independently before I saw #378. Reviewers should treat these as alternatives — either merge one, or fold the two additions this PR has (below) into #378. Happy to close this in favor of #378.

## Problem

`lstk start` crashed immediately with:

```
LocalStack exited unexpectedly:
/usr/local/bin/docker-entrypoint.sh: line 22: mkdir: command not found
```

`mkdir: command not found` means the container's `PATH` was empty when the entrypoint ran.

## Root cause

LocalStack's `docker-entrypoint.sh` re-exports every `LOCALSTACK_<NAME>` variable as `<NAME>` (stripping the prefix) via a line-oriented `source <(env | … | sed …)`, and does **not** set `PATH` itself — it relies on the image's own `ENV PATH`.

Since #161, lstk forwards **all** host `LOCALSTACK_*` variables into the container verbatim (`filterHostEnv`). So a host `LOCALSTACK_PATH` became `export PATH=…` inside the container, blanking `PATH`. The entrypoint's very next command (`mkdir -p /var/lib/localstack/logs`) then failed with `mkdir: command not found` and the emulator exited on startup — independent of config file and image tag, matching the report.

## Fix

`filterHostEnv` now drops:
- `LOCALSTACK_*` variables whose entrypoint-stripped name is a reserved shell variable (`PATH`, `HOME`, `SHELL`, `IFS`, `ENV`, `BASH_ENV`, `LD_PRELOAD`, `LD_LIBRARY_PATH`, `PYTHONPATH`, `PYTHONHOME`); and
- any value containing a newline/carriage return, which could inject spurious `export` lines through the entrypoint's line-oriented pipeline (a second, distinct way to blank `PATH`).

`LOCALSTACK_AUTH_TOKEN` filtering and normal `CI` / `LOCALSTACK_*` forwarding are unchanged.

**Compared to #378, this PR additionally:** (1) sanitizes newline/CR values, and (2) covers a slightly wider reserved set. #378 additionally emits a user-facing warning for each dropped var, which is a nice touch this PR lacks and worth keeping.

## Tests

- Unit: `TestFilterHostEnv` now asserts `LOCALSTACK_PATH`, other reserved-name vars, and newline-containing values are not forwarded.
- Integration: `TestStartCommandIgnoresPathClobberingHostEnv` starts the emulator with `LOCALSTACK_PATH` set and asserts it comes up healthy (fails with the `mkdir: command not found` crash before the fix). Needs Docker + `LOCALSTACK_AUTH_TOKEN`, so it runs in CI.

Separately worth filing upstream: excluding `PATH` and friends from prefix-stripping in the image entrypoint itself, which would also protect compose users who export `LOCALSTACK_PATH` without lstk involved.

### DIFF
--- a/internal/container/start.go
+++ b/internal/container/start.go
@@ -863,15 +863,56 @@ func awaitStartup(ctx context.Context, rt runtime.Runtime, sink output.Sink, con
 	}
 }
 
+// entrypointReservedEnvKeys are variable names that must not be overwritten
+// inside the emulator container. LocalStack's docker-entrypoint.sh re-exports
+// every LOCALSTACK_<NAME> variable as <NAME> (stripping the prefix) via a
+// line-oriented `source <(env | ... | sed ...)`, and does not set PATH itself —
+// it relies on the image's own PATH. So forwarding e.g. LOCALSTACK_PATH turns
+// into `export PATH=...` inside the container, blanking/overwriting PATH; the
+// next entrypoint command (`mkdir -p ...`) then fails with
+// "mkdir: command not found" and the emulator exits on startup (DEVX-984).
+var entrypointReservedEnvKeys = map[string]bool{
+	"PATH":            true,
+	"HOME":            true,
+	"SHELL":           true,
+	"IFS":             true,
+	"ENV":             true,
+	"BASH_ENV":        true,
+	"LD_PRELOAD":      true,
+	"LD_LIBRARY_PATH": true,
+	"PYTHONPATH":      true,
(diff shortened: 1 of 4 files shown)


gtsiolis:
Closing in favor of #378.

PR #380, opened 2026-07-09
fix: emulator selection not prompting on fresh install

anisaoshafi:
Quite some commands (login, status, stop, setup, snapshot, az, aws, etc.) wired `PreRunE: initConfig(nil)` which eagerly writes the default config.toml (type = "aws") to disk on their own first. 
That's the reason why the "Which emulator would you like to use?" prompt didn't trigger, because this one relies on the config.toml file being present or not.

I had a new device set up, where I freshly install lstk, that's how I noticed.

Renamed `nitConfig(nil)` func to `initConfigDeferCreate(nil)` which now does read-only config load, no eager write.
Added regression test, and also verified manually uninstalling then installing lstk, which now triggers the emulator selector to pop up.


gtsiolis:

gtsiolis:
Heads-up: Rebased and resolved conflicts.

anisaoshafi:
Thanks @gtsiolis. Feel free to review it and we can merge it if it looks good to you 🙌

CLAUDE.md
```diff
@@ -102,12 +102,14 @@ Uses Viper with TOML format. lstk uses the first `config.toml` found in this ord
...
 
 Created automatically on first run with defaults. Supports emulator types: `aws`, `snowflake`, and `azure`.
 
+`initConfigDeferCreate` (wrapping `config.Load`) only ever *reads* config — it never writes the default config.toml to disk. That's deliberate: the emulator-selection prompt (`container.SelectEmulator`) is shown only when `firstRun` is still true, and only bare `lstk` and `lstk start` wire it in (`NeedsEmulatorSelection: firstRun` in `startEmulator`). If some other command eagerly persisted a default (`type = "aws"`) config on its own first run, the selector would never get a chance to show on a genuinely fresh install — every command must use `initConfigDeferCreate`, never a hypothetical eager-create variant, so that only a real emulator start (interactive selection, or the non-interactive default-emulator path) ever writes the file. `EnsureCreated()` therefore has exactly two legitimate callers: the non-interactive first-run path in `cmd/root.go` (after a successful default start) and `container.SelectEmulator` (after the user picks one).
```
gtsiolis: **issue(non-blocking)**: This says two callers, but there are three: `cmd/root.go`, `container.SelectEmulator`, and `container.ApplyEmulatorType` (the --type path from #381, already on main), right?

cmd/aws.go
```diff
@@ -53,7 +53,7 @@ Examples:
 					return err
 				}
 			}
-			return initConfig(nil)(cmd, args)
+			return initConfigDeferCreate(nil)(cmd, args)
```
gtsiolis, line 56: **suggestion**: L51 above has a stale reference to the removed function, right?

gtsiolis (approved):
Confirmed the behavior. Left two comments.

cmd/aws.go
anisaoshafi: Thanks @gtsiolis 👏🏼  handled in here: https://github.com/localstack/lstk/pull/380/commits/c18d5c3f8d84b807f446945c5fe5cd93d6f290dc

CLAUDE.md
anisaoshafi: Handled in https://github.com/localstack/lstk/pull/380/commits/c18d5c3f8d84b807f446945c5fe5cd93d6f290dc

PR #381, opened 2026-07-10
feat: add --type flag to start for non-interactive emulator selection

gtsiolis:
Adds `--type`/`-t` (`aws|snowflake|azure`) to `lstk start` and the bare root command as the non-interactive answer to the first-run emulator picker. Because `stop`, `status`, `logs`, `volume`, and snapshot auto-load all resolve from the configured type, the flag **records** the selection in config rather than overriding it per-run — an ephemeral override would strand every follow-up command (none of them take `--type`), so persisting keeps config and reality in sync for CI scripts.

Behaviour:
- **No config yet:** creates config with the selected type, reusing the picker's `EnsureCreated`/`SetEmulatorType` path.
- **Config matches:** no-op.
- **Config differs:** switches the `type` line in place via a surgical rewrite. The search is anchored to the active `[[containers]]` block (from its header to the next table header), so quotes, spacing, trailing comments, other keys, and any `type` key in another table (e.g. an `[env.*]` profile) are all left untouched. A note names the file.

On a switch the other block fields are kept, with guards: a custom `image` is a **hard error** (it pins a specific product that can't be reinterpreted under a new type — use `--config` for a separate profile); a non-`latest` `tag` and any `volumes`/`volume` are kept with a **warning**; `port`/`env`/`snapshot` are kept silently. A config with more than one `[[containers]]` block is rejected **before** any write, since only one block can start.

The domain logic lives in `container.ApplyEmulatorType` (parallel to `container.SelectEmulator`) and is applied at the top of `startEmulator` before snapshot/start-options are resolved, so it runs before the TUI.

Also in this PR:
- A positional (`lstk start azure`) is rejected with a hint pointing at `--type` (previously it silently started AWS).
- README CI examples for selecting the emulator non-interactively.

Dropped along the way (with why):
- **Positional emulator selection** (`lstk start azure` as an accepted alias). `lstk aws`/`lstk az` are already proxy subcommands and `lstk azure`/`lstk snowflake` route to extension dispatch, so a positional could only work on `start`, never symmetrically on the bare root — it would teach a mental model the root contradicts. `start` now rejects a positional with a hint instead.
- **`-c` shorthand for `--config`.** It was stripped from proxy args, which swallowed the `-c` flags wrapped tools own (CDK `-c`/`--context`, SAM `-c`/`--cached`) — e.g. `lstk cdk synth -c env=prod` failed with `failed to read config file: open env=prod`. `--config` stays long-form only; proxy commands forward `-c` to the wrapped tool.

- **Ephemeral per-run override** (start Azure once without touching config). See the opening note: the follow-up commands are config-seeded, so a non-persistent flag would half-work. The flag persists instead.

| Switch type | Cannot switch type |
|--------|--------|
| Screenshot 2026-07-10 at 16 55 33 | Screenshot 2026-07-10 at 16 56 15 |


gtsiolis:

anisaoshafi (changes requested):
I run these two commands consecutively:
`./bin/lstk -t azure`
`./bin/lstk -t snowflake`. 
Because azure emulator was already running, `./bin/lstk -t snowflake` fails with an error `Your config specifies the LocalStack Snowflake Emulator. Only one emulator can run on a port at a time.`, however it DOES change the configuration profile.

Image

I think it shouldn't change the config file if it doesn't finish successfully, because some subsequent commands will fail as they won't detect the running emulator, e.g `./bin/lstk status`:

Image

gtsiolis:
@anisaoshafi Nice catch, fixed in de2b2e6.

The --type switch now checks whether a different emulator is already running on the port first, and refuses the switch (config untouched) if so. Added tests, including one reproducing your azure -> snowflake repro.

Could you take another look?

anisaoshafi:
> @anisaoshafi Nice catch, fixed in [de2b2e6](https://github.com/localstack/lstk/commit/de2b2e671a79ceec52662e1169b9708c2c80cc92).
> 
> The --type switch now checks whether a different emulator is already running on the port first, and refuses the switch (config untouched) if so. Added tests, including one reproducing your azure -> snowflake repro.
> 
> Could you take another look?

Thanks for tackling that 💯

anisaoshafi (approved):

gtsiolis:

---

PR #385, opened 2026-07-14
Fix dev and custom image tags failing license validation

gtsiolis:
## Motivation

Starting with `tag = "dev"` (or any tag the license server cannot parse as a version, e.g. custom enterprise-mirror tags) failed before pulling anything: the pre-flight license check treated the server's tag-format rejection as a fatal license verdict. This blocks dev-image use in CI and forces users testing locally built images back to the v1 CLI.

## Changes

- `validateLicense` treats an `IsUnsupportedTag` rejection as "the server cannot judge this tag", not a license verdict: it warns and skips the pre-flight, and the emulator validates the license at startup — the same degradation as the existing offline/transport-failure path. Genuine token/subscription rejections stay fatal.
- The warning keeps a tag suggestion (`try a tag like "2026.6" or "latest"`) so a typo'd tag stays diagnosable, and the benign skip logs at Info instead of Error.
- Updated the `IsUnsupportedTag` doc comment to its new load-bearing semantics; extracted `config.TagSuggestion()` and unexported `unsupportedTagMessage`.
- Parameterized `createLicenseMockServer` (status + body) so the new integration test reuses it instead of duplicating the handler.
- Documented the new degradation path in CLAUDE.md.

## Tests

- New integration test `TestUnsupportedTagSkipsPreflightValidation` reproduces the exact DEVX-912 failure (fails before the fix, passes after): dev tag + format-rejecting license server, asserting the start proceeds past the license gate with the warning. An unreachable-registry image override keeps it fast (no real image pull).
- New unit test `TestValidateLicense_SkipsPreflightOnUnsupportedTag`; existing tests confirm 403 rejections still fail fast pre-pull and the offline fallback is unchanged.

## Todo

- [ ] TUI plan label shows "No license" for dev-tag starts (`label.go` re-sends the raw tag to the license API; pre-existing, becomes more visible with this fix)
- [ ] Consider a distinct telemetry event for skipped pre-flights (the offline and unsupported-tag paths are both silent today)

Closes DEVX-912


internal/container/start.go
```diff
@@ -758,16 +758,27 @@ func validateLicense(ctx context.Context, sink output.Sink, opts StartOptions, c
...
-		// degrade. Gating this on licErr.Status is tracked as follow-up.
+		if licErr.IsUnsupportedTag {
+			// The server rejecting the tag *format* (e.g. a "dev" nightly or a custom
+			// enterprise tag) is not a verdict on the license itself. Degrade like the
+			// transport failure above: skip the pre-flight and let the container
+			// validate its own bundled license at startup. The suggestion keeps a
+			// typo'd tag diagnosable, since the later pull failure won't mention tags.
+			opts.Logger.Info("license server does not support tag %q, deferring license validation to the emulator: %s", version, licErr.Detail)
+			sink.Emit(output.MessageEvent{Severity: output.SeverityWarning, Text: fmt.Sprintf(
+				"The license server does not support tag %q; continuing with the image's bundled license — if this is unintended, %s",
```
tiurin: suggestion: update the wording to tell the user what is going to happen. ~The meaning of "bundled license" is not very clear.~ Checking bundled license is only the last fallback after checking other expected license file locations

```suggestion
				"The license server does not support tag %q; the emulator will validate the license once it starts. If this is unintended, %s",
```

tiurin (approved):

internal/container/start.go
```diff
@@ -758,16 +758,27 @@ func validateLicense(ctx context.Context, sink output.Sink, opts StartOptions, c
...
 		}
-		// Known limitation: any *api.LicenseError — i.e. any non-200 HTTP response,
-		// including a 5xx or a 407 from a corporate proxy — is treated as a definitive
-		// verdict and stays fatal here; only connection-level failures (handled above)
-		// degrade. Gating this on licErr.Status is tracked as follow-up.
+		if licErr.IsUnsupportedTag {
+			// The server rejecting the tag *format* (e.g. a "dev" nightly or a custom
+			// enterprise tag) is not a verdict on the license itself. Degrade like the
+			// transport failure above: skip the pre-flight and let the container
+			// validate its own bundled license at startup. The suggestion keeps a
```
tiurin: If I'm not mistaken, the emulator container will first validate license in a standard location or try to get on using auth token before defaulting to the bundled one.
```suggestion
			// validate the license at startup. The suggestion keeps a
```

internal/container/start_test.go
```diff
@@ -475,6 +475,40 @@ func TestValidateLicense_FailsOnServerRejection(t *testing.T) {
 	assert.Contains(t, err.Error(), "license validation failed")
 }
 
+func TestValidateLicense_SkipsPreflightOnUnsupportedTag(t *testing.T) {
+	// The server rejecting the tag *format* (e.g. a "dev" nightly or a custom
+	// enterprise tag) is not a verdict on the license itself — the pre-flight must
+	// be skipped so the container validates its own bundled license at startup,
```
tiurin:
```suggestion
	// be skipped so the container validates the license at startup,
```

internal/container/start.go
tiurin: @gtsiolis I've pushed this change, along with updating other "bundled license" references in https://github.com/localstack/lstk/pull/385/commits/9ef6e7a2161cc6634bd92a9ac207c6160d736de5

anisaoshafi (approved):
Tested and looks good, thanks 🚀

gtsiolis:
Thanks for merging, @tiurin and @anisaoshafi! 🏀

---

PR #387, opened 2026-07-16
Fix lstk logs to properly detect emulators running under custom container names

anisaoshafi:
`lstk logs` didn't detect externally named containers.

logs should use the same container resolution logic as the other commands. That would make the behavior consistent across lstk.

Integration test failed and worked after fix.

### DIFF
--- a/internal/container/logs.go
+++ b/internal/container/logs.go
@@ -39,10 +39,25 @@ func Logs(ctx context.Context, rt runtime.Runtime, sink output.Sink, containers 
 	// TODO: handle logs per container
 	c := containers[0]
 
+	name, err := ResolveRunningContainerName(ctx, rt, c)
+	if err != nil {
+		return fmt.Errorf("checking %s running: %w", c.Name(), err)
+	}
+	if name == "" {
+		sink.Emit(output.ErrorEvent{
+			Title: fmt.Sprintf("%s is not running", c.DisplayName()),
+			Actions: []output.ErrorAction{
+				{Label: "Start LocalStack:", Value: "lstk"},
+				{Label: "See help:", Value: "lstk -h"},
+			},
+		})
+		return output.NewSilentError(fmt.Errorf("%s is not running", c.Name()))
+	}
+
 	pr, pw := io.Pipe()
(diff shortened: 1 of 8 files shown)


internal/container/logs.go
```diff
@@ -39,10 +39,25 @@ func Logs(ctx context.Context, rt runtime.Runtime, sink output.Sink, containers
...
+	if name == "" {
+		sink.Emit(output.ErrorEvent{
+			Title: fmt.Sprintf("%s is not running", c.DisplayName()),
+			Actions: []output.ErrorAction{
+				{Label: "Start LocalStack:", Value: "lstk"},
+				{Label: "See help:", Value: "lstk -h"},
+			},
+		})
+		return output.NewSilentError(fmt.Errorf("%s is not running", c.Name()))
+	}
```
tiurin, line 48: question: this is the 4th copy of this code, the first three are in aws.go, az.go and status.go. Maybe we can extract it to a `handleNoRunningContainer` in, let's say, running.go?

internal/container/logs_test.go
```diff
@@ -97,3 +98,21 @@ func TestLogs_FilteredLinesDropped(t *testing.T) {
 	require.Len(t, sink.lines, 1)
 	assert.Contains(t, sink.lines[0].Line, "keep")
 }
+
+func TestLogs_NotRunning(t *testing.T) {
```
tiurin: question: It is not very clear from the test name and assertions what is the intention of the test. Could you please provide more details and maybe update test name and message/comment for assertion?

tiurin (approved):

internal/container/logs_test.go
anisaoshafi: Done in https://github.com/localstack/lstk/pull/387/commits/79a67682ccda9bf3d84e9e0fc0836e37705d4ae1

internal/container/logs.go
anisaoshafi: Addressed in https://github.com/localstack/lstk/pull/387/commits/b1eb7e4311bc9c189dd1e6657ba55afc42768a4e, good point

PR #391, opened 2026-07-16
feat: forward SIGINT/SIGTERM to wrapped tools for graceful shutdown

gtsiolis:
Pressing Ctrl-C during a wrapped-tool run (`lstk terraform apply`, `aws`, `cdk`, `sam`, `az`, extensions) killed the child immediately instead of letting it shut down gracefully — so terraform never got to release its state lock. The tools run via `exec.CommandContext` with lstk's root context, which is cancelled on `SIGINT`/`SIGTERM`; `exec.CommandContext`'s default `Cancel` then SIGKILLs the child right away.

All wrapped-tool passthroughs now run through a new `proc.Run` helper that:

- **disarms the SIGKILL-on-cancel** (its `Cancel` returns `os.ErrProcessDone`), so the tool terminates from the signal it received and lstk waits for its own shutdown — preserving the real exit code (Ctrl-C on terraform now exits 130 after full cleanup instead of `signal: killed`/255);
- **forwards `SIGTERM` unconditionally**: a terminal never generates it, so without forwarding `kill <lstk-pid>`, `timeout 30 lstk …`, or an IDE stop button would never reach the tool;
- **forwards `SIGINT` only when none of stdin/stdout/stderr is a terminal**: an attached terminal already delivers Ctrl-C to the child via the foreground process group, and a second near-simultaneous SIGINT makes tools like terraform abort without cleaning up. Checking all three streams matters — with only stdin redirected (`yes | lstk terraform apply`) the process group still delivers;
- **registers its signal handler before `Start`**, so a signal landing during startup is queued rather than lost.

(`npm/launcher.js` forwards unconditionally instead — safe there because its child is lstk itself, which tolerates duplicate signals; wrapped tools don't.)

Short internal captured-output execs (version checks, schema discovery, backend provisioning) are unchanged. No configurable grace period yet, per the issue scope.

**Tests**: `internal/proc` unit tests cover the disarm/exit-code semantics and the forwarding policy; `test/integration/signal_forwarding_test.go` covers the behavior end to end through a new `signal-wait` mode of the reference extension (exits 40 + number of signals received) — non-interactive SIGINT/SIGTERM forwarding, exactly-one SIGINT on Ctrl-C in a PTY, the piped-stdin double-signal regression, and SIGTERM reaching the tool from inside a terminal.

Fixes DEVX-1000


skyrpex (approved):
The changes look good to me! Thanks for this PR... This change very important for some of the CLIs.

PR #397, opened 2026-07-20
Add real-emulator e2e tests for snapshot load --merge strategies

peter-smith-phd:
## Summary

Adds a real end-to-end test for `lstk snapshot load --merge=overwrite|account-region-merge|service-merge` against a real LocalStack container — the existing snapshot load tests only exercise a mocked emulator and never touch the actual merge round-trip.

Builds two snapshots files containing slightly-overlapping content. The test loads the first snapshot, merges in the second snapshot (using the three different strategies), then validates the output. `lstk aws`, `lstk save`, `lstk load --merge=<strategy>` are all used, just as the user would use them.

Closes DEVX-946.


anisaoshafi (approved):
Quite useful 💯, thanks for the comprehensive test

PR #398, opened 2026-07-21
Warn when multiple lstk installs are detected on PATH

skyrpex:
## Motivation

A previous Homebrew installation can shadow a fresh npm install: `npm install` appears successful while `lstk` still resolves to the older binary. Detect this and tell the user.

## Changes

- Add `FindInstalls` (`internal/update/installs.go`): scans every absolute `PATH` directory for lstk executables, resolves symlinks, and deduplicates real duplicates via `os.SameFile` (symlink aliases like `/usr/local/bin` → Caskroom, hardlinks, repeated PATH dirs count once). Results follow PATH order, so the first entry is what a shell executes.
- Per-OS candidate probing: Unix checks for a regular `lstk` file with an execute bit (same test as `exec.LookPath`); Windows probes `PATHEXT` extensions (`lstk.exe` from binary/scoop installs, `lstk.cmd` from npm shims) and ignores bare extensionless files, matching cmd.exe resolution.
- Each install is labeled with the existing install-method classifier (homebrew/npm/binary) and flagged when it is the currently running executable — including the npm case where the PATH entry resolves to the Node launcher script while the process is the Go binary from the platform package (matched by shared `node_modules` root).
- `WarnMultipleInstalls` emits a new `MultipleInstallsEvent` when two or more distinct installs are found. Wired into `lstk update` (all modes) and the start-path update notification (`NotifyUpdate`), since a stale shadowing install is exactly what makes updates appear to not take effect.
- Event plumbing: plain-text formatter, TUI rendering via the existing `FormatEventLine` fallback, and a `MULTIPLE_INSTALLS` warning in the `--json` envelope.

Example output:

```
> Warning: Multiple lstk installations found on PATH:
  /opt/homebrew/bin/lstk (homebrew, currently running)
  /Users/me/.nvm/versions/node/v22/bin/lstk (npm)
  Your shell runs the first one; remove the others to avoid using a stale version.
```

## Tests

- Unit tests for the PATH scan: dedup of symlink aliases and repeated dirs, exec-bit and broken-symlink handling, install-method classification, running-executable marker, plus Windows-tagged `PATHEXT` tests (also cross-compiled with `GOOS=windows`).
- Format-parity tests for the new event (plain formatter and JSON envelope warning).
- Integration tests running the real binary: warns with two copies on PATH, stays silent with a single install or symlinked aliases, and reports the warning code in `--json`.

Closes DEVX-882

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,7 +67,7 @@ Notes:
   - `terminal/` - Plain-mode terminal helpers (spinner, TTY detection)
   - `tracing/` - OpenTelemetry setup (`LSTK_OTEL=1`)
   - `ui/` - Bubble Tea views for interactive output
-  - `update/` - Self-update logic: version check via GitHub API, binary/Homebrew/npm update paths, archive extraction
+  - `update/` - Self-update logic: version check via GitHub API, binary/Homebrew/npm update paths, archive extraction; also detects multiple lstk installs on PATH (`FindInstalls`/`WarnMultipleInstalls`, warned on `lstk update` and the start-path update notification)
   - `version/` - Version info
   - `volume/` - `lstk volume` domain logic
 
--- a/internal/output/envelope_sink.go
+++ b/internal/output/envelope_sink.go
@@ -25,6 +25,10 @@ func NewEnvelopeSink(format Format) *EnvelopeSink {
 // which carries no machine-readable code of its own today.
 const warningCodeNotice = "NOTICE"
 
+// warningCodeMultipleInstalls flags that more than one distinct lstk install
+// was found on PATH.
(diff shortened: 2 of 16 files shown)


anisaoshafi (approved):
Looks good, thanks for handling this 🏅 

There was an issue with the rendering in TUI (interactive mode) because it was missing some colors so it looked the same in non-interactive mode. I took the liberty to push a fix: https://github.com/localstack/lstk/pull/398/changes/24400a21c2223e0f42aefd6a9aebaea709bcf9d6

| Before | After |
|-------|--------|
| Image | Image |

skyrpex:
Thank you for the attention to the detail and the proactiveness :)

PR #399, opened 2026-07-21
Auto-detect Rancher Desktop and Podman container runtimes

skyrpex:
## What

`lstk` now works out of the box with Rancher Desktop and Podman, alongside Docker Desktop, Colima, OrbStack, and Lima:

- **Runtime auto-discovery**: added socket auto-detection for Rancher Desktop and Podman (machine on macOS, rootful/rootless on Linux), and `lstk` now also resolves the current Docker CLI context (`docker context use …` / `DOCKER_CONTEXT`). An explicit `DOCKER_HOST` still always wins.
- **Helpful "runtime not available" errors**: the error now suggests the start command for the runtime you actually have (`rdctl start`, `podman machine start`, `colima start`, …) instead of always assuming Docker Desktop.
- **Port 443 degrades gracefully**: runtimes like Rancher Desktop (Traefik ingress) and rootless Podman often make port 443 unavailable. Instead of failing the start, `lstk` now starts without 443 and prints a warning with a runtime-specific fix. HTTPS stays available on the edge port (4566 serves both HTTP and HTTPS). If you explicitly configured 443 via `GATEWAY_LISTEN`, a conflict remains a hard error.
- **Leftover containers self-heal**: every container lstk creates is now labeled (`cloud.localstack.lstk=true`). If a start fails after the container was created (e.g. a rejected port bind), the next `lstk start` removes the leftover automatically instead of failing with a name conflict; `lstk restart` racing the old container's self-removal is handled the same way, silently. Containers lstk did not create are never removed — a clear "name is already taken" error replaces the daemon's raw name-conflict, and if cleanup itself fails the error names the container and the exact `docker rm` command.
- **Docs**: new [docs/container-runtimes.md](https://github.com/localstack/lstk/blob/claude/container-runtimes-research-83ccd0/docs/container-runtimes.md) with per-runtime setup notes and caveats.

## Why

Users on Rancher Desktop or Podman had to manually export `DOCKER_HOST`, and then still hit a fatal, cryptic port-443 error on startup. Discovery and startup should just work across Docker-API-compatible runtimes (DEVX-639, DEVX-637).

## How to test

- **Rancher Desktop**: install it, set container engine to dockerd (moby), quit Docker Desktop, then `lstk start` — auto-detected via `~/.rd/docker.sock` or the `rancher-desktop` docker context. With Kubernetes enabled, expect the 443 warning with the Traefik hint (transcript below).
- **Podman machine (macOS)**: `podman machine init && podman machine start`, quit other runtimes, then `lstk start` — auto-detected via the podman machine socket; also works via a manually created context (`docker context create podman --docker "host=unix://$(podman machine inspect --format '{{.ConnectionInfo.PodmanSocket.Path}}')" && docker context use podman`). In rootless mode, expect the 443 warning with the `--rootful` hint (transcript below).
- **Colima / OrbStack**: start one, quit the others, `lstk start` — auto-detected as before.
- **Override**: `DOCKER_HOST=unix://… lstk start` must beat context and auto-detection.
- **Stopped runtime**: with everything stopped, `lstk start` should suggest the start command for whatever runtime is installed.

## Before / after

Before (any runtime that cannot bind 443 — fatal):

```
$ ./bin/lstk start

   ▟████▖   lstk (dev)
  ▟██▙█▙█▟  LocalStack Enterprise
    ▀▛▀▛▀   ~/.config/lstk/config.toml

✔︎ Pulled localstack/localstack-pro:latest
✔︎ Valid license (Enterprise)
✗ failed to start LocalStack: Error response from daemon: something went wrong with the request: "listen tcp 127.0.0.1:443: bind: permission denied\n"
```

After (rootless Podman machine — the daemon rejects the 443 bind at start time, `lstk` retries without it):

```
$ ./bin/lstk start

   ▟████▖   lstk (dev)
  ▟██▙█▙█▟  LocalStack Enterprise
    ▀▛▀▛▀   ~/.config/lstk/config.toml

✔︎ Pulled localstack/localstack-pro:latest
✔︎ Valid license (Enterprise)
> Warning: Port 443 cannot be published (bind: permission denied) — starting without it. Clients that hardwire HTTPS on port 443 must use
           https://localhost:4566 instead (the edge port serves both HTTP and HTTPS). Podman machine cannot publish ports below 1024 in rootless mode
           — switch with `podman machine set --rootful`, then restart lstk.
✔︎ LocalStack is running (containerId: 18c4f21195cf)
• Endpoint: localhost.localstack.cloud:4566
• Web app: https://app.localstack.cloud
> Tip: View emulator logs: lstk logs --follow
```

After (Rancher Desktop with Traefik holding 443 — caught by the preflight, before the image pull):

```
$ ./bin/lstk start

   ▟████▖   lstk (dev)
  ▟██▙█▙█▟  LocalStack Enterprise
    ▀▛▀▛▀   ~/.config/lstk/config.toml

> Warning: Port 443 is in use — starting without it. Clients that hardwire HTTPS on port 443 must use https://localhost:4566 instead (the edge port
           serves both HTTP and HTTPS). Rancher Desktop's Kubernetes ingress (Traefik) usually holds port 443 — free it with `rdctl set
           --kubernetes.options.traefik=false`, then restart lstk.
✔︎ Pulled localstack/localstack-pro:latest
✔︎ Valid license (Enterprise)
✔︎ LocalStack is running (containerId: d14e0903f36d)
• Endpoint: localhost.localstack.cloud:4566
• Web app: https://app.localstack.cloud
> Tip: View emulator logs: lstk logs --follow
```

Closes DEVX-637
Towards DEVX-639


skyrpex:
Replying to the review (https://github.com/localstack/lstk/pull/399#pullrequestreview-4772608622):

**1. Leftover container on failed cleanup** — went further than a hint. Every container lstk creates is now stamped with a `cloud.localstack.lstk=true` label, and the pre-start check self-heals: a labeled leftover from a failed start is removed automatically, so the follow-up name conflict never happens. The guard is conservative — containers without the label are never removed (clear "name already taken" error instead of the daemon's raw conflict), unlabeled-but-AutoRemove containers mid-self-removal are waited out (covers pre-label lstk versions and `lstk restart` racing AutoRemove — the created-state-only version of this guard actually broke CI's restart tests before being relaxed). The manual `docker rm -f <name>` hint survives only as the true last resort, when removal itself fails twice in a row.

**2. Context dial on the hot path** — conscious call. The probe only runs when a non-default context is set, and it's a local unix `connect()`: sub-millisecond when the socket exists, immediate `ENOENT` when it doesn't — 200ms is the worst-case cap, not the per-call cost. Honoring the context blindly would turn a stale context (deleted VM, leftover `DOCKER_CONTEXT`) into a hard connection failure later with a worse message — the same failure mode the socket probe list already guards against for stale sockets. Every lstk command needs the daemon anyway, so the dial is work the SDK would do immediately afterward regardless; if it ever shows up in a profile we can cache per invocation.

anisaoshafi:
Currently reviewing this 👀

anisaoshafi (approved):
Thanks for following up on claude review. 

Didn't check in detail, but tried to test according to the instructions in PR description, and didn't get any issue.
Test coverage is good with new tests added, and existing tests remaining (and adjusted), also nice that this is documented in CLAUDE.md, helps to understand the detection logic 💯

PR #408, opened 2026-07-24
Fix multiple-installs false positives for asdf and mise shims

gtsiolis:
## Motivation

lstk warns about "multiple installs" whenever a version-manager shim and the install it dispatches to are both reachable on PATH, even though they are the same install — a false positive that fires on every run.

- **asdf**: an asdf-managed npm install is also reachable through its asdf shim.
- **mise** (reported by Daniel): with shim mode and PATH mode both enabled (mise's recommended mix for interactive + non-interactive sessions), `~/.local/share/mise/shims/lstk` and `~/.local/share/mise/installs/<tool>/<version>/lstk` both land on PATH. A mise shim is a bare symlink to the `mise` binary itself (argv[0] dispatch, no metadata), so neither symlink dedup nor content parsing can relate it to its install.

## Changes

- Resolve asdf shim scripts back to their real install path (`internal/update/installs_unix.go`) by parsing the `# asdf-plugin: <name> <version>` comment asdf writes into each shim, before dedup runs in `FindInstalls` (`internal/update/installs.go`)
- Resolve dispatcher shims (mise): a candidate in a `shims/` dir that symlink-resolves to an executable not named `lstk` is aliased to the first PATH candidate under the sibling `installs/` tree — the same check mise itself uses to recognize its shims (`is_active_mise_shim`), no hardcoded tool names. A shim with no backing install on PATH (pure shim mode + e.g. a homebrew lstk) still counts, so genuine conflicts keep warning.
- Add a no-op `executableAlias` on Windows (`installs_windows.go`); mise's Windows `.exe`-copy shims are a known gap left out of scope
- Add unit tests `TestFindInstallsDeduplicatesAsdfShimAlias`, `TestFindInstallsDeduplicatesMiseShimAlias`, `TestFindInstallsKeepsMiseShimWithoutBackingInstallOnPath` and integration tests `TestUpdateCheckDoesNotWarnOnAsdfShimAlias`, `TestUpdateCheckDoesNotWarnOnMiseShimAlias`

Known limitation: mise resolves the shim's dispatch target per-invocation from its config, so a hand-pinned stale `installs/<tool>/<old-version>` PATH entry diverged from the configured version is deduped rather than warned. Inherent to any heuristic that doesn't execute mise; documented in the code.

## Review

Human review advised — user-facing behavior with a teammate-reported case (Daniel's mise setup), and the dispatcher heuristic deserves a second pair of eyes.

## Tests

New unit + integration tests reproduce both false positives and confirm they're resolved. Also verified against real installs: asdf + npm on this machine, and a real `mise use github:localstack/lstk` install (dual shim + PATH mode) that reproduced Daniel's warning byte-for-byte with released v0.18.0 and is silent with this branch, while a mise shim + separate binary install still warns.

Closes DEVX-1015

### DIFF
--- a/internal/update/installs.go
+++ b/internal/update/installs.go
@@ -26,34 +26,38 @@ type Install struct {
 func FindInstalls(getenv func(string) string) []Install {
 	runningInfo, runningResolved := runningExecutable()
 
-	var installs []Install
-	var seen []os.FileInfo
+	var candidates []string
 	for _, dir := range filepath.SplitList(getenv("PATH")) {
 		// Relative and empty entries (cwd-dependent lookup) are skipped: they
 		// resolve differently per invocation and are not real install locations.
 		if dir == "" || !filepath.IsAbs(dir) {
 			continue
 		}
-		for _, candidate := range executableCandidates(dir, getenv) {
-			resolved, err := filepath.EvalSymlinks(candidate)
-			if err != nil {
-				resolved = candidate
-			}
-			info, err := os.Stat(resolved)
-			if err != nil {
-				continue
-			}
-			if isDuplicate(seen, info) {
-				continue
-			}
-			seen = append(seen, info)
-			installs = append(installs, Install{
-				Path:         candidate,
-				ResolvedPath: resolved,
-				Method:       classifyPath(resolved),
-				Running:      isRunning(info, resolved, runningInfo, runningResolved),
-			})
+		candidates = append(candidates, executableCandidates(dir, getenv)...)
+	}
+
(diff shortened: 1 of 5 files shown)


gtsiolis:
Cc @skyrpex because https://github.com/localstack/lstk/pull/398.

skyrpex:
@gtsiolis thanks, didn't know about this. I thought we were fine deduping symlinks 😄

gtsiolis:
Closing in favor of #455. Cc @skyrpex

PR #415, opened 2026-07-28
Prioritize LOCALSTACK_AUTH_TOKEN over stored credentials

gtsiolis:
## Summary

Stored keyring credentials previously overrode `LOCALSTACK_AUTH_TOKEN`. Once a user had logged in, a token supplied for a CI job, a second account, or a single invocation was silently ignored until the stored credential was removed.

This change makes caller-provided authentication authoritative. Token resolution is now:

1. `LOCALSTACK_AUTH_TOKEN`
2. Stored keyring credentials
3. Browser login, when the command supports it

The non-interactive resolution used by telemetry, snapshot commands, extensions, and emulator startup shares the same precedence helper as `auth.GetToken`, preventing the two paths from drifting. Tokens are trimmed before resolution, so an empty or whitespace-only environment value does not mask a valid stored credential.

This also composes with DEVX-1022's authentication boundary. An environment token overrides a stored token and is sent as the caller credential for both local and external emulator-backed Cloud Pod operations. Reuse of a locally managed emulator's startup identity is only considered when neither an environment nor stored caller token is available; external targets still require caller authentication.

Logout behavior is unchanged: it removes stored credentials, while an environment-provided token remains controlled by the caller's environment.

## Verification

- `go test ./internal/auth ./cmd`
- `make test-integration RUN='Test(EnvAuthTokenOverridesStoredToken|StoredTokenUsedWithoutEnvAuthToken)'`
- `make test-integration RUN='TestLogoutCommand(RemovesToken|SucceedsWhenNoToken|WithEnvVarToken)$'`
- `go vet ./...` in the root and integration-test modules
- Pre-commit secret scan

Review: advisable — this changes credential precedence for every authenticated command and defines how explicit caller credentials interact with the local-emulator identity fallback.

Closes DEVX-1023


anisaoshafi:
Will look at this a bit later 👀

anisaoshafi (approved):
Great UX improvement especially for CI/non-interactive setups.
It is now more aligned with the convention that other CLI tools follow (token in env overriding stored token)  💯

gtsiolis:

---

PR #416, opened 2026-07-28
Restrict Cloud Pod identity reuse to local emulators

gtsiolis:
## Summary

`lstk start` passes `LOCALSTACK_AUTH_TOKEN` to the emulator without persisting an environment-provided secret. A later command against that same locally managed emulator can therefore have no caller token even though the emulator already has a valid startup identity.

This change lets emulator-backed Cloud Pod operations reuse that identity only for an lstk-managed local emulator. When the caller supplies a token, lstk continues to send it as the `Authorization` header, overriding the startup identity.

External endpoints do not get this fallback. `snapshot load`, `save`, `remove`, and `load --dry-run` require a token from the environment or keychain before lstk calls a protected pod endpoint. Network access to an external emulator is therefore not treated as permission to use its startup identity.

Platform-direct `snapshot list`, `show`, and `versions` continue to require caller authentication. S3 remotes are unchanged because they use the caller's AWS credentials instead of a LocalStack platform token.

## Error handling

When a locally managed emulator cannot use either the caller token or its startup identity, 401/403 responses are mapped to `snapshot.ErrAuthRequired` and rendered as an actionable authentication error. Pod/version not-found and unavailable-feature classifications keep precedence over the generic auth mapping.

This change limits lstk's client behavior; it does not add request authentication to the emulator itself. Per-request protection of privileged emulator endpoints remains a separate server-side hardening concern.

## Verification

- `go test ./cmd ./internal/emulator/aws ./internal/snapshot`
- `make test-integration RUN=TestExternalPodOperationsRequireCallerAuthentication`
- `make test-integration RUN=TestSnapshotLoadPodVersion`
- `go vet ./...` in the root and integration-test modules

The full unit run passed 1,516 of 1,517 tests. The remaining port-sensitive test could not bind because an existing `localstack-snowflake` container already owned port 4510.

Review: advisable — this changes the authentication boundary for emulator-backed Cloud Pod operations.

Closes DEVX-1022

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -178,12 +178,13 @@ When lstk's stdout and stderr are both terminals, `lstk aws` runs the child via 
 
 - `lstk snapshot save [destination]` (`-s`/`--services`: comma-separated list to limit a save to a subset of services; applies uniformly to local files, `pod:` cloud snapshots, and S3-remote saves) / `lstk snapshot load REF` (`--merge`: `account-region-merge` default, `overwrite`, `service-merge`) / `list` (cloud; `--all` for org-wide) / `remove REF` / `show REF` / `versions REF`.
 - `show`, `versions`, and bare `list` (no `s3://` arg) only ever call the LocalStack platform API — never the emulator. `remove`, by contrast, proxies its delete through the running emulator despite being a "cloud" (`pod:`) concept, so it still requires one to be reachable; don't conflate "operates on cloud-hosted storage" with "never touches the emulator" — they're different axes.
-- A REF is a local `.snapshot` file, a `pod:` cloud snapshot on the LocalStack platform (requires auth), or an `s3://bucket/prefix` remote in the user's own bucket (the emulator performs the transfer; S3 supports save/load/list only).
+- A REF is a local `.snapshot` file, a `pod:` cloud snapshot on the LocalStack platform (requires an identity), or an `s3://bucket/prefix` remote in the user's own bucket (the emulator performs the transfer; S3 supports save/load/list only).
 - Every save to an existing `pod:` snapshot creates a new **version**. `versions REF` lists them; `load` and `show` accept a `pod:<name>:<version>` REF (latest when omitted). `save`, `remove`, `versions`, and S3 remotes reject a version suffix rather than ignore it.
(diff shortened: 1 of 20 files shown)


peter-smith-phd:
I'm a little bit concerned about the security aspects of this PR. Are we saying that an `lstk start` command requires a valid `LOCALSTACK_AUTH_TOKEN`, but any other `lstk` commands after that are just accepted without any authentication?<br><br>We now get into a situation with a valid user starts the emulator, but then any other user can assume that user's credentials, without supplying their own. This makes sense when the emulator is locked down to only run on a single (network-confined) host, but as soon as we start running emulators in cloud-based environments, this becomes an issue.<br><br>I know that are a lot of other security gaps (especially with our `test` / `test` ) credentials, but this PR feels like it's creating a new problem, rather than making it better.<br><br>I'd be in favour of either enforcing `LOCALSTACK_AUTH_TOKEN` is provided, or enforcing a login session. In either case, `lstk` should always be sending an auth token to the emulator, and the emulator should always be validating it.

gtsiolis:
Good callout.

The intent is to remove repeated authentication friction for a locally managed, single-user emulator, not to make network access equivalent to authorization. ✔️

I’ve narrowed the PR so startup-identity reuse now applies only to an lstk-managed local emulator. External targets require caller authentication from the environment or keychain before any Cloud Pod load, save, remove, or dry-run operation. S3 remotes remain separate because they use the caller’s AWS credentials.

I think that’s the right boundary for this PR. Separately, we could harden the emulator so privileged endpoints authenticate every request, ideally with a scoped instance credential rather than the container’s platform token.

Could you take another look at this new approach?

peter-smith-phd:
> I’ve narrowed the PR so startup-identity reuse now applies only to an lstk-managed local emulator. 

Sorry, still not convinced. Now we're making the behaviour of a "local" emulator different from that of a "remote" emulator. I honestly don't see the problem with telling customers, to 1) Login properly, 2) Use `LOCALSTACK_AUTH_TOKEN` consistently.

If we're picking-and-choosing when authentication is required, and when it's OK to not have it, that just makes a more complicated security story.

gtsiolis:
Fair point. I agree that a single, explicit authentication model is preferable.

Now that #415 is merged, users can log in locally or set `LOCALSTACK_AUTH_TOKEN` at the job level in CI, and the environment token will take precedence consistently. That covers the original workflow without introducing different authentication behavior for local and remote emulators.

Closing this. Thanks for the thoughtful pushback on the security model.

PR #418, opened 2026-07-29
Stream lstk logs lines to real terminal scrollback

anisaoshafi:
- Interactive `lstk logs` capped history at 200 lines and repainted it in place, so terminal scrollback stopped growing past that point (unlike `docker logs` / `--non-interactive`).
- `LogLineEvent` now prints permanently above the TUI via a new `output.NewStreamingLogSink` + `Program.Println`, bypassing the capped buffer.
- Printed synchronously (not via `tea.Cmd`) to keep lines in order — Bubble Tea runs returned `Cmd`s concurrently, which reordered lines in an earlier pass.

## Test plan
- [x] Unit tests for the new sink
- [x] New integration test: 550 lines on a real container, `lstk logs` under a real PTY, all survive in order
- [x] `make test` + `make lint` clean
- [x] Manually verified against a real emulator (600 lines, real PTY)
- [x] `--non-interactive` unchanged

Closes DEVX-1027

### DIFF
--- a/internal/output/tui_sink.go
+++ b/internal/output/tui_sink.go
@@ -5,16 +5,39 @@ type Sender interface {
 	Send(msg any)
 }
 
+// LogLinePrinter abstracts Bubble Tea's Program.Println — permanent output
+// that persists across renders, unlike a Sender's capped/repainted lines.
+type LogLinePrinter interface {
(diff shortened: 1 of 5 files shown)


skyrpex (approved):
Thanks, now the `lstk logs` command is really useful 👍🏻

PR #420, opened 2026-07-30
Support a custom container name in config

skyrpex:
Adds one optional `[[containers]]` field, `container_name`, so the emulator can be given a fixed name that something outside lstk can address. This closes the gap that kept the Jenkins blog tutorial (LocalStack behind a sidecar proxy) on the legacy CLI.

## Example

```toml
[[containers]]
type = "aws"
tag  = "latest"
port = "4566"

container_name = "ls-jenkins"
```

```bash
lstk start
docker ps --format '{{.Names}}'   # ls-jenkins
```

## How it works

`container_name` sets `docker run --name`. The default is unchanged: `localstack-<type>`, plus `-<tag>` when the tag isn't `latest`.

It is not cosmetic. The same value is exported into the container as `MAIN_CONTAINER_NAME`, which the emulator uses to introspect itself over the Docker socket and to name the containers it spawns (`<name>-lambda-<fn>-<id>`). lstk already exported that variable from the derived name, so both sides move together for free.

Three details worth a reviewer's attention:

- **The persistence directory does not follow the name.** `VolumeDir()`'s default deliberately uses the *derived* name, so setting or changing `container_name` never silently orphans existing state, and the path stays byte-identical for existing users. `lstk volume path` is unaffected.
- **Discovery needed no changes.** `ResolveRunningContainerName` is already name-first, falling back to an `(image repo, internal port)` scan, so `stop`, `status`, `logs`, `restart`, snapshots, and the proxy commands all pick up a custom name automatically.
- **`MAIN_CONTAINER_NAME` in an `[env.*]` profile was a trap** — it never reached `docker run --name`, so the container's real and self-reported names disagreed. `startOnce` now warns and points at this field.

Naming: the key is `container_name` rather than a bare `name` because a `[[containers]]` block already carries other names (the image, the emulator type), against which an unqualified `name` reads ambiguously.

Invalid values are rejected at config load by `validate.ContainerName`, mirroring Docker's own `^[a-zA-Z0-9][a-zA-Z0-9_.-]*$` rule, so a bad name fails immediately instead of at container creation.

## Scope

Docker network attachment and DNS aliases were implemented on this branch and have been removed — they are out of scope for DEVX-996. The diff no longer touches `internal/runtime/` at all.

## Testing

- Unit: `validate.ContainerName` rule-code table, `Name()` precedence, config validation, and a regression guard that `VolumeDir()`'s default ignores a custom name.
- Integration: `TestStartCommandUsesCustomContainerName` starts a real container and asserts the Docker-assigned name, a matching `MAIN_CONTAINER_NAME`, an unchanged `lstk volume path`, and that `lstk status` / `lstk stop` both resolve it. `TestConfigWithInvalidContainerNameFails` covers the rejection path with no daemon needed.
- Green in CI, including the container-starting test. `make test` and `make lint` (0 issues, both modules) pass locally.

One fix came out of running these rather than trusting a local run: the custom-name test asserted removal immediately after `lstk stop`, but containers are created with `AutoRemove`, so removal is asynchronous and the assertion raced. It now polls with `require.Eventually`, the pattern already used elsewhere in the suite.

## Review

Advising a human review: this adds user-facing config surface, and the container name feeds container creation, discovery, and the emulator's own self-identification.

Closes DEVX-996


peter-smith-phd:
Thanks for doing this. Off the top of my head, I have some thoughts:

First, this is similar to [DEVX-949: Allow configuring the emulator container's Docker network mode (`--network`) in `lstk`](https://linear.app/localstack/issue/DEVX-949/allow-configuring-the-emulator-containers-docker-network-mode-network), which I recently suggested that we cancel. The reason was that anybody trying to manage a non-trivial set of containers should instead be using `docker-compose` or the Kubernetes equivalent. If we add these new fields to `lstk`, it opens the door for every docker-compose feature to be added into `config.toml`, making `lstk` a lot more complicated, and will ultimate lead us to expose `DOCKER_FLAGS` (etc) that make this non-portable across different container runtimes.

So, I'm wondering if the tutorials/blogs should just use `docker-compose.yml` instead? @remotesynth, what are your thoughts on this? This is what I've been planning for the docs changes.

But, if we agree to proceed with this PR, then a couple of other things:
- I generally dislike fields such as `name` because it's too generic. Name of the container? DNS Name? Both? I'd prefer to qualify it with the type of object it's naming (e.g. `dns_name`)
- We also have [DEVX-994: Allow the `--host-dns` (or similar) option to export network ports](https://linear.app/localstack/issue/DEVX-994/allow-the-host-dns-or-similar-option-to-export-network-ports), which needs a similar (`config.toml`) solution. Contrary to my thoughts about name/network/aliases, I do feel that exposing port 53 is something we should bake into `lstk`.

skyrpex:
Agreed. IMO, `lstk` generally serves two distinct purposes:

* Interact with the emulator APIs with helpers (AWS, Azure, etc)
* Orchestrate the emulators

There are many ways and technologies to launch the emulators (containers, docker-compose, Kubernetes, container-less), but we chose to stick to Docker for simplicity. The way I see it, `lstk start` is just a fast and simple way to get started. For more complex scenarios, relying on docker-compose and writing proper documentation is the way to go.

That's my opinion though, and I might be missing crucial context. Let's see what Brian says 🙂.

skyrpex:
Thanks — the scope has moved since this review was written, so two of the three points are already resolved.

**Network and aliases are gone.** `network` and `network_aliases` were removed from this PR; the diff no longer touches `internal/runtime/` at all. What's left is a single field, `container_name`. That's consistent with cancelling DEVX-949 rather than in tension with it.

**Naming.** Renamed `name` → `container_name`, which follows your principle of qualifying by the object being named. I didn't use `dns_name` because it would describe the field inaccurately: this sets `docker run --name`, i.e. the container's identity. It doubles as a DNS name on a user-defined network, but that's a side effect of Docker's embedded DNS, not what the field is for. `container_name` also matches docker-compose's key for the same concept.

**On whether this belongs in `lstk` at all** — your call and @remotesynth's, and I'll follow whatever you decide. One data point for the decision: `container_name` isn't purely a compose-style knob. lstk already exports the derived name as `MAIN_CONTAINER_NAME`, which the emulator uses to introspect itself over the Docker socket and to name the containers it spawns (`<name>-lambda-<fn>-<id>`). A compose user has to set `container_name:` *and* `MAIN_CONTAINER_NAME` and keep them in sync by hand — the legacy CLI has a lint warning for exactly that mismatch. This field keeps the two coupled by construction. So the slippery-slope argument lands squarely on network/aliases (now removed) but is weaker for the container's own name.

Happy to close this if the answer is "use docker-compose in the tutorials" — no attachment to the change.

DEVX-994 / port 53 is out of scope here; no objection to it being handled separately.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #422, opened 2026-07-30
Fix wrong daemon picked when another container runtime is installed

DrisDary:
Since 0.20.0, `lstk` can report a running emulator as absent:

```
Error: LocalStack Azure Emulator is not running
  ==> Start LocalStack: lstk
```

...while the container is up and serving its health endpoint. This broke every sample run in `localstack-azure-samples` CI (~80% of GitHub runners, unaffected by retries, not reproducible on a dev machine).

**Cause.** #399 added a native Podman socket to the probe list (`/run/podman/podman.sock`, `$XDG_RUNTIME_DIR/podman/podman.sock`). Those are probed before the Docker SDK's default, and Podman ships preinstalled on GitHub's Ubuntu runner images. Where its socket is live, `lstk` connects to Podman, finds no emulator container there, and reports it as not running. Probing only dials the socket — it can't tell that this is the wrong daemon.

A second bug sits next to it: an explicit `DOCKER_HOST` was silently overridden. `NewDockerRuntime` starts from `client.FromEnv`, then appends `client.WithHost(...)` for whatever detection found, and the later option wins. Both `CLAUDE.md` and `docs/container-runtimes.md` already promise the opposite — "This always wins — `lstk` never overrides an explicit environment configuration".

## Changes

- `DOCKER_HOST` short-circuits detection, matching the documented contract.
- On Linux, a live `/var/run/docker.sock` is preferred over the alternative-runtime probes. A co-installed runtime shouldn't outrank Docker for a tool that speaks the Docker API, and an operator who wants the other one still has `DOCKER_HOST` and the Docker CLI context, both honoured first.
- Linux only, deliberately: on macOS/Windows that path belongs to whichever VM runtime is active, and the VM probes must stay first there or `isVM`/`Flavor` stop classifying it as VM-backed — which drives both the bind-mount rewrite and the tailored "start Docker Desktop" hints.
- `findDockerSocket` gained an injectable form (`findDockerSocketFor(limaInstance, home, nativeSocket, goos)`), following the same style as `resolveDockerContextHost`. Two existing tests called `findDockerSocket()` directly and inherited the host's real Docker socket, so they would have started failing on any machine with a Docker daemon; they now pass an absent native socket and test the probe order deterministically.

## Tests

Reproduced end to end against a real emulator with a fake rootless-Podman socket live alongside Docker:

| | result |
|---|---|
| 0.19.0 | works |
| 0.20.0 | fails |
| 0.20.0 + this fix | works |

Unit tests cover both halves and each is falsifiable — reverting either one alone fails its own test. New cases: Docker preferred over a co-installed runtime, fallback when Docker is absent, a stale socket file not shadowing a live alternative, macOS keeping VM probes first, Lima short-circuiting, and `DOCKER_HOST` beating the probe.

`go vet`, `gofmt`, `go build` clean. `TestStart_SecondLicenseRejectionAfterReloginRendersErrorEvent` fails both with and without this change — pre-existing on `main`, unrelated. Integration suite not run (needs `LOCALSTACK_AUTH_TOKEN`).

**Review:** worth a human review — this changes daemon-selection behaviour for all users, and the Linux-only scoping is the main judgement call. Unit tests rather than integration: which socket got selected isn't observable from the binary, so `DaemonHost()` is the direct assertion.


anisaoshafi:
Thanks for the prompt PR and thorough testing, really appreciated @DrisDary 🙌🏼 
Just pushed a commit to fix some tests on macos https://github.com/localstack/lstk/pull/422/commits/71b8502dd2be2652e5f1c523dba2accad6b8b7fe

anisaoshafi (approved):

PR #428, opened 2026-08-03
Convey the telemetry session ID to extensions so their events join lstk's

carillan81:
Why
lstk records every extension invocation as an lstk_command telemetry event named ext:<name>, carrying the duration and exit code. That is all lstk can know — an extension is a separate process, and the detail that matters often exists only inside it. The doctor extension is the first case: it wants to report its per-check diagnostic results, which lstk's event cannot contain.

So two events describe one invocation, with no way to join them. The telemetry session ID is generated per telemetry.Client, i.e. per process, so an extension that builds its own client invents an unrelated ID. The only join available today is machine_id plus client-time proximity, which is fuzzy in general and simply wrong when several runs land in the same second on one host — exactly what CI does.

What
Adds sessionId to LSTK_EXT_CONTEXT, set to the same session ID stamped on the ext:<name> event lstk emits for that invocation. An extension that stamps it on its own events gets an exact join, and it works for every bundled extension rather than needing per-extension plumbing.

The field is omitted when lstk's telemetry is disabled — there is no session to correlate. That means an extension cannot distinguish "telemetry disabled" from "an lstk older than this field"; both are absence, both mean no correlation is available, and both call for the same fallback. The ambiguity is documented rather than engineered away.

LSTK_EXT_API_VERSION stays at 1: the addition is additive, detected by presence like every post-v1 field. Nothing about what lstk itself emits changes, and an extension built before this runs unmodified.

The consuming side lives in lstk-bundled-extensions and treats the field as optional, falling back to a locally generated ID, so the two repos don't block each other.

Towards DPX-612


anisaoshafi (approved):
LGTM, thanks for the clear context and congrats on your first contribution 👏🏼

PR #440, opened 2026-08-09
Add --account flag for account selection in lstk aws

peter-smith-phd:
## Motivation

Prior to this PR, `lstk terraform` and `lstk sam` had accepted `--account` to target a specific LocalStack account, but `lstk aws` did not support `--account`. It did support the `AWS_ACCESS_KEY_ID` environment variable, but that only worked on a fresh machine and then silently stopped once `lstk setup aws` was run (to create `config.toml`). This happened because lstk was passing `--profile localstack`, which caused Botocore to ignore all the environment variables.

## Solution

This PR adds `lstk aws --account <12 digits>`, following the terraform/sam convention.

Also included are some smaller fixes found along the way, thanks for exhaustive manual testing:

- terraform's S3 backend provisioning had been ignoring the resolved account, creating the state bucket in a different account than the generated backend block addressed.
- `lstk sam --region` was silently overridden by a `region` in `samconfig.toml`; it is now also passed on sam's command line, and only when explicitly given.
- lstk's leading flags now parse in any order relative to the wrapped tool's own — `lstk aws --region X --account Y …` used to leak `--account` to the AWS CLI, while the reverse order worked.

<details>
<summary>Docs</summary>

- New `lstk aws --account <id>` — 12 digits, before the AWS service name, falling back to `AWS_ACCESS_KEY_ID` then the default `000000000000`.
- `lstk sam --region` now takes effect over `samconfig.toml`; without the flag, `samconfig.toml` still decides.

Help text is updated in place; `lstk docs` regenerates from it.

</details>

Closes DEVX-1039


cmd/aws.go
```diff
@@ -38,14 +38,22 @@ Equivalent to running:
   aws --endpoint-url http://localhost:4566 <args>
 with AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and AWS_DEFAULT_REGION set automatically.
 
+lstk-specific flags (must appear before the aws subcommand):
```
anisaoshafi, line 41: @peter-smith-phd I'm a bit late, but just noticed this wording. Shouldn't "before" here, be "after"? aws" is the subcommand, maybe we can say this needs to be before the aws service name instead?

peter-smith-phd: I see what you're saying, but from the `aws` point of view, the service name is the subcommand. So it's saying "must appear before aws's subcommand". I agree it's confusing, and you're welcome to change it if you want to.

anisaoshafi: Understood, it's actually correct both ways, only point of view changes (lstk vs aws). No need to change it, thank you for clarifying!

PR #442, opened 2026-08-10
Skip redundant image pulls in integration test helpers

skyrpex:
## Motivation

The integration-test stand-in helpers (`startTestContainer`, `startNamedTestContainer`, `commitNeverHealthyImage`, plus five direct sites in `start_test.go`) called `ImagePull` unconditionally on every invocation — ~100 times per suite run. Even with the image already local, each pull is a registry round trip (manifest/digest resolution, seconds each), adding minutes of pure network overhead per run.

## Solution

- New `ensureImage(t, ctx, image)` helper: `ImageInspect` first, `ImagePull` only when the image is missing. All alpine/nginx stand-in pull sites now go through it. Stand-in images only need to exist — a stale local copy is fine.
- `commitNeverHealthyImage` is now inspect-first and leaves `lstk-never-healthy:test` in place across runs (it adds no layer data on top of the alpine test image), so repeat runs skip the build entirely; the source container is removed right after the commit instead of at test cleanup, with a leftover-name pre-remove guarding interrupted runs.
- `TestStartFallsBackToLocalImageWhenPullFails` deliberately keeps pulling `localstack/localstack-pro:latest` — it starts a real emulator, so image freshness matters there.

Measured on a 42-test batch (status/stop/logs): 100.0s → 73.4s (~27%), identical results otherwise. The full suite has ~100 helper call sites versus ~35 exercised in that batch, so total savings are larger.

## Docs

<details>
<summary>Docs impact</summary>

Nothing to document — test-internal change only, no user-facing behavior.

</details>

## Review

Small, test-only, mechanical (pull sites swapped for a shared inspect-first helper); verified with vet, golangci-lint, and targeted + batch integration runs, with a stash-baseline confirming the only local failures are pre-existing environment issues. Self-merge candidate.

### DIFF
--- a/test/integration/main_test.go
+++ b/test/integration/main_test.go
@@ -171,16 +171,30 @@ const (
 	testImage     = "alpine:latest"
 )
 
+// ensureImage makes sure image is present locally, pulling it only when missing.
+// The inspect-first check matters: the stand-in helpers below run ~100 times per
+// suite, and an unconditional ImagePull contacts the registry (manifest/digest
+// resolution, seconds each) even when the image is already local. Stand-in
+// images only need to exist — a stale local copy is fine.
+func ensureImage(t *testing.T, ctx context.Context, image string) {
+	t.Helper()
+
+	if _, err := dockerClient.ImageInspect(ctx, image); err == nil {
+		return
+	}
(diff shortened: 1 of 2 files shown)


skyrpex:
This was an experiment to reduce the CI clock time but was a failure :)

PR #444, opened 2026-08-10
Run PTY and fake-tool integration tests on Windows

skyrpex:
## Motivation

Windows CI skipped ~48 integration tests in two whole categories: every PTY-based test (`creack/pty` has no Windows support) and every test faking an external CLI (`aws`, `az`, `terraform`, `cdk`, `sam`, `aws_completer`) with a `#!/bin/sh` script. That left the interactive paths — login flow, setup prompts, emulator selector, spinner behavior, `--json` TTY gating — with zero Windows coverage, exactly where OS-specific bugs (keyring, `%AppData%` paths, ConPTY rendering) would hide.

## Solution

- **Cross-platform PTY helpers** (`test/integration/pty_helpers_test.go`): `runLstkInPTY` / `startLstkInPTY` / `startCmdInPTY` return a `ptyProc` with `waitForOutput`/`write`/`wait`/`kill`, backed by `charmbracelet/x/xpty` (Unix PTY on macOS/Linux, ConPTY on Windows). Output is ANSI-stripped (ConPTY injects its own repaint sequences), `wait` yields an `*exec.ExitError` on every platform, and the ConPTY buffer is wide (300 cols) so its line re-wrapping can't break substring assertions. All 31 direct `pty.Start` sites migrated; `creack/pty` remains only in the `!windows`-tagged signal-forwarding tests.
- **Compiled fake tools** (`test-samples/faketool` + `faketool_test.go`): one Go stand-in binary, built once per run and copied onto PATH under the impersonated name, driven by a JSON sidecar (placeholder echo lines, `--version`-style arg cases, sleep, exit codes, record-to-file). Replaces every shell-script fake, including `aws_completer` and the fake browser openers (`rundll32` on Windows).
- **Windows-correct env isolation**: new `env.WithHome(dir)` sets `HOME` + `USERPROFILE` + `APPDATA` (Windows binaries read the latter two), swept across all 119 `With(env.Home, ...)` sites; `unreachableDockerHost` is now `tcp://localhost:1` (valid on every OS, unlike a unix socket path).
- **Skips**: down from ~48 to 14, each with an accurate reason. What remains is inherent: Docker (`requireDocker` — no nested virtualization on Windows CI), unix-socket tests, POSIX signal/`os.Interrupt` tests, bash completion, symlink-permission tests, the two inner-PTY streaming tests (`proc.RunInPTY` is unix-only), and one read-only-directory permission test. One stale skip (az_interception, which uses no PTY) removed outright.

After this, Windows runs the same test set macOS does, modulo the inherent list above. Verified: `go vet`/build clean for darwin and `GOOS=windows`, `make lint` clean on both modules, unit tests pass, and the migrated integration tests (including interactive PTY flows) pass locally on macOS with Docker.

## Docs

Nothing user-facing to document — test infrastructure only. CLAUDE.md's Testing section is updated in this PR with the new conventions (use the PTY helpers, use faketool instead of shell scripts, no convenience `GOOS == "windows"` skips, `tcp://localhost:1` for unreachable Docker).

## Review

Human review advisable: wide test-infrastructure change, and the ConPTY drain/close ordering in `ptyProc.wait` is behavior-sensitive. Real ConPTY behavior is only provable on Windows CI — watch that job for flakes in TUI-layout assertions (`TestStatusEndpointURLInteractiveRendersTUI` is the most at-risk).


anisaoshafi (approved):
Great job un-skipping so many tests, Cristian! 
We gain more confidence around windows behaviour and avoid breaking things in the future 🪨 :shipit: 
Thanks for challenging the status quo and finding this solution.

Skimmed through the code and looks good to me!

PR #445, opened 2026-08-11
Mandate TDD with observable-behavior e2e tests in agent guidance

skyrpex:
## Motivation

Agent sessions were writing implementations first and backfilling tests. We want the opposite: every feature or fix starts from a failing end-to-end test.

## Solution

Extends the Testing section in CLAUDE.md (AGENTS.md symlinks to it) with three rules: TDD is mandatory (red → green → refactor — write the e2e test, confirm it fails for the expected reason, then implement); e2e tests assert only observable behavior (exit codes, output, files, requests received), asking the user when the observable behavior is unclear; and observable mocks of external systems (mock auth endpoint, fake `aws`/`az` on PATH, fake browser opener) are encouraged as the product boundary — never mocks of lstk internals. Also adds a matching checklist line to the `/review-pr` skill so the rule is enforced at review time.

<details>
<summary>Docs</summary>

Nothing to document — agent/contributor guidance only, no user-facing behavior change.

</details>

## Review

Self-merge candidate — guidance-only change, no code touched, content was specified directly in the session.


gtsiolis:
What do you think, @anisaoshafi? Looks good to merge, or close?

anisaoshafi (approved):
💯 Good idea to make this explicit - I did have to prompt several times during on my reviews to actually do this red -> green testing to make sure the implementation took the right effect.

PR #446, opened 2026-08-11
Convey endpointUrl and machineId to extensions via LSTK_EXT_CONTEXT

carillan81:
Extensions gain two optional runtime-context fields: endpointUrl carries the resolved --endpoint-url / LSTK_ENDPOINT_URL / AWS_ENDPOINT_URL value verbatim (unvalidated and unprobed, omitted when no source is set), and machineId carries the anonymized machine id lstk stamps on its own telemetry, so an extension reports the same machine without re-deriving it (omitted together with sessionId when telemetry is disabled).

The machine-id derivation now lives in MachineID(ctx) behind the enabled guard, so a disabled client never dials Docker or persists a machine_id file, and the Docker lookup is bounded by a 3s timeout so an unreachable DOCKER_HOST cannot stall dispatch for the OS TCP connect timeout before the extension execs.


skyrpex (approved):
LGTM, let's wait for Shelton's decision.

sheltongraves:
@peter-smith-phd This came up today and I can see it is connected to Doctor extension. Do you have any feedback before it merges?

peter-smith-phd (approved):
Thanks, looks good to me. A lot of file changes, but this essentially just adds two necessary fields to the extension context.

PR #450, opened 2026-08-11
Remove the four nested internal/*/CLAUDE.md files

tiurin:
## Motivation

The four nested `CLAUDE.md` files under `internal/` read as prose restatements of the code they describe, and agents kept appending more literal detail to them. An audit section by section against the code found **~75-80% of their content already stated** in doc comments, error strings, test names, or Cobra help text — frequently more precisely there — and **9 claims that had drifted into contradicting the code**, including one file disagreeing with itself two sections apart.

#366 introduced these files to cut always-loaded context by 25%, and was merged explicitly provisionally: *"Expecting the separate agent instructions in sub-folders to make better use of context, but let's revert if this does not work as expected."* Measured 22 days later:

| Surface | Before #366 | At merge | Today |
|---|---:|---:|---:|
| Root `CLAUDE.md` | 4,649 w | 3,541 w | **6,402 w** |
| 4 nested files | 0 | 1,808 w | **4,387 w** |
| **Total** | **4,649 w** | **5,349 w** | **10,789 w (2.3×)** |

The root file is now larger than before the split that shrank it. Every commit to a nested file *appended a section*; none revisited an existing one. Files a feature PR touched tripled in size (`container` 486→1,535, `snapshot` 491→1,464); `azureconfig`, which no PR touched, went stale and never learned `lstk az` grew an `--endpoint-url` flag. This PR exercises that revert clause.

## Solution

Two commits, deliberately split so the migration can be reviewed without deletions in the diff.

**5d15e93 — move the residue into doc comments.** The ~12 facts that genuinely were not in code are anchored on the symbols that own them, so the next edit to each mechanism has the rationale in its diff. Notably `isFeatureUnavailableResponse` now carries the emulator-side `plux` plugin behaviour, why the 404 discriminator must stay narrow, and why a local license pre-check would risk blocking paying customers — it is the single chokepoint all 11 gated methods funnel through. Comment-only: 56 added lines, zero Go statements.

**101ad46 — delete the 8 files and retarget the root.** Removes 4 `CLAUDE.md` + 4 `AGENTS.md` symlinks. Root `CLAUDE.md` repoints its 9 references at the owning code symbols and now prohibits nested agent files, with an explicit destination for each kind of detail: per-symbol → doc comment, user config → `default_config.toml`, user commands → Cobra `Long`, design rationale → `openspec/changes/<id>/design.md`, provenance → commits and tickets. Each of those has a reader who notices when it is wrong, which is the property the nested files lacked.

Before deleting anything, verified that **all 164 backticked identifiers** named in the four files still resolve in code, and that every migrated fact has a home. `make build` / `make lint` clean (0 issues, both modules); `make test` 1626/1627, the one failure being a local port-4510 conflict from a running emulator, unrelated to the diff.

Deleted content stays available via `git show <sha>^:<path>`, recorded in the commit message.

<details>
<summary>Docs</summary>

**Nothing user-facing in this PR.** It changes agent-instruction files and Go doc comments only — no command, flag, env var, output, or documented behavior is affected, hence `docs: skip`.

One adjacent gap it surfaces but does not fix, worth a writer's awareness: **`expose_ports` has no user-facing documentation anywhere.** It appears only in Go code and in the two files deleted here — not in `default_config.toml` (which ships as the user's own commented config and documents every other container field), not in `docs/`, and not on docs.localstack.cloud. Being handled in a separate follow-up.

</details>

## Review

Human review advised — this reverses a recently merged team decision (#366) whose approvers explicitly asked to keep an eye on it, and it was not discussed beforehand. Small and behavior-free, but the call belongs to the people who wrote the revert condition. Worth a look from the #366 reviewers in particular.

## Todo

- [ ] Add a Linear ticket reference before undrafting — none exists for this work yet
- [ ] Follow-up: document `expose_ports` in `default_config.toml`
- [ ] Follow-up: fix two drifted code comments the audit found in passing (`snapshot/destination.go` lists `show` among version-rejecting contexts while `ParseShowable` accepts one; `ports/ports.go` says `CheckAvailable` reports "free to bind" when it dials)


gtsiolis (approved):
Thanks, @tiurin! 
 
Indeed, the majority of the checkable claims in those nested files were already in doc comments, error strings, test names, help text, etc. Good observation! 👏 

Nested `CLAUDE.md` loading works as intended and is a documented setup for **[large codebases](https://code.claude.com/docs/en/large-codebases)**. ✔️  

That said, nested `CLAUDE.md` files and instruction sets like these will most probably become obsolete in the near future. The **[newer generation models](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models)** are tuned to need fewer instruction files and more judgment. 

One thing worth adding is the smaller main `CLAUDE.md` we aimed for with #366 didn't stick either, the root file is back above its pre-split size, something we have to keep smaller. ⚠️ 

I'm fine with merging since we're mostly using newer models and the code now carries the useful part of these instructions. Easy **[two-way door decision](https://localstack-cloud.slack.com/archives/C060ZRAULEL/p1750156338696689?thread_ts=1750118116.031059&cid=C060ZRAULEL)**. 🚪

anisaoshafi (approved):
Agree, it was starting to look like it commented what almost each line of code does line by line 😅 
We should store in claude the decisions that are important for future reference, but no need to comment the code in such detail. Thanks for the clean up 🧹

PR #452, opened 2026-08-12
expand openspec for extension delivery architecture

carillan81:
This PR includes only openspec docs for the extension bundling mechanism. It expands previous work (Thanks Peter!) and defines how the bunlding would work and the tasks to make it real.
This is intended to be the first draft to move forward the project. Any feedback, change, correction, pun or non pejorative expression will be very welcome!


openspec/changes/add-bundled-extension-distribution/proposal.md
```diff
@@ -2,31 +2,39 @@
...
+- **Package bundled extensions into every install channel** so they land in the directory lstk resolves, with no `PATH` change required by the user:
+  - binary archive: `lstk-*` binaries and `lstk-extensions.toml` as siblings of `lstk` at the archive root;
+  - Homebrew: automatic via the cask's Caskroom staging of the whole archive (lstk ships as a **cask**, not a formula — no libexec involved); the cask's post-install quarantine hook is widened from the single `lstk` binary to the whole staged directory;
+  - npm: bundled files are copied into each **platform package** (`@localstack/lstk-<os>-<arch>`), where the real binary lives — not the wrapper package — via a post-processing step in the release job.
+- **Pull the prebuilt closed-source bundled binaries from the private extensions repository's releases** into the release build context, **version-pinned via a pin file in this repo** (`bundled/extensions.version`), checksum-verified against the private release's manifest, authenticated with a dedicated read-only token, without exposing source in the public repository.
```
peter-smith-phd: I'm not clear on the the `bundled/extensions.version` concept. I had been thinking that each `lstk` package would take the _latest_ version of the bundled extensions. There would be no need for any "pinning". Was this a conscious change? Does it introduce a maintenance burden we don't currently have?

openspec/changes/add-bundled-extension-distribution/proposal.md
```diff
@@ -2,31 +2,39 @@
...
+- **Ship the hand-authored descriptions file** (`lstk-extensions.toml`), owned by LocalStack's private extensions repository, and **validate it at release time** (`scripts/check-descriptions.sh`) so a described-but-missing extension is a release-blocking error.
+- **Update the `lstk`/`lstk-*` set as one unit** in `internal/update` for the self-managed binary channel (stage `.lstk-new` siblings, then rename, lstk last); Homebrew and npm replace the whole package — and therefore the whole set — via their package managers.
+- **Guarantee update continuity**: `lstk update` keeps working for every existing install across the transition — a pre-bundling lstk updates cleanly into the first bundling release on all three channels (Homebrew and npm especially, where the updater shells out to the package manager), and a bundling lstk updates cleanly from an archive that carries no extensions (rollback). Bundled extensions are payload, never a precondition: their absence must not fail an update.
```
peter-smith-phd: _their absence must not fail an update._

Kind of... at least this makes sense for a rollback of the initial non-bundling -> bundling transition (that is, going back to a non-bundled-extensions release). But upgrading from a bundling `lstk` to another bundling `lstk` should require the extensions to be downloaded/installed correctly. They're not optional.

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,76 @@
...
+`add-extension-mechanism` (PR #340) delivered the extension *mechanism* and everything lstk needs at runtime: bundled-directory resolution (`extension.BundledDir` — the directory of the symlink-resolved lstk executable, searched ahead of PATH), descriptions loading (`extension.LoadDescriptions` reads `lstk-extensions.toml` from that directory, degrading to an empty map on any failure), and help wiring (`cmd/extension.go`). It intentionally deferred *distribution* so the first release could validate bundled extensions by manual placement. This change automates getting LocalStack's bundled extensions into the install artifacts and keeping them version-matched with lstk.
+
+The distribution code was prototyped during `add-extension-mechanism` and removed before the squash-merge; it is **not recoverable from git history** (the squashed commit contains only the runtime code and these openspec docs). Everything below is designed against the current tree.
```
peter-smith-phd, line 5: I love it how Claude remembers everything that I did 😄 (or should that make me nervous?)

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,76 @@
...
+
+### Decision 2: Bundled binaries come from the private repo's releases, pinned by a file in this repo
+
+The private extensions repository — the same source of truth that builds the closed-source binaries and hand-authors the descriptions file — publishes **tagged releases** whose assets are the per-platform binaries (`lstk-<name>_<os>_<arch>[.exe]`), `lstk-extensions.toml`, and a `checksums.txt` manifest covering them.
+
+This repo carries a **pin file**, `bundled/extensions.version` — a single line naming the private release tag. Each lstk release therefore maps deterministically and reproducibly to one extensions bundle; bumping the pin is an ordinary reviewable PR (automatable later from the private repo's release workflow). `scripts/fetch-bundled-extensions.sh` reads the pin, downloads the assets (`gh release download`), **verifies each against the bundle's `checksums.txt`**, and stages them under `bundled/` with canonical names. It hard-fails when any lstk target platform has no matching asset (subject to an explicit not-supported allowlist), so platform gaps surface at pull time, not as an empty-glob failure inside GoReleaser.
```
peter-smith-phd: The "reviewable PR" seems like an extra burden for the `lstk` maintainers. Is the expectation that somebody will review the extensions release and make a go/no-go decision on whether to start using them? I was thinking that we always release the latest extensions alongside each `lstk` release.

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,76 @@
...
 
-`internal/update` treats `lstk` and its bundled `lstk-*` set (binaries + the descriptions file) as one unit. For the self-managed binary channel, the extractor stages every new `lstk`/`lstk-*` member next to its destination (`.lstk-new` siblings) and renames each into place, so an interrupted update never leaves `lstk` and a bundled extension at mismatched versions. For Homebrew and npm, the package manager replaces the whole package — and therefore the whole bundled set — atomically.
+The credential is a **dedicated fine-grained read-only PAT** (contents: read on the private repo only), stored as a repository/organization secret — not a reuse of the broader `PRO_ACCESS_TOKEN`. Least privilege, independent rotation.
```
peter-smith-phd, line 36: Just to clarify, this "credential" is GitHub's access to allow the `lstk` repository actions to access the `lstk-bundled-extensions` releases?

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,76 @@
...
+
+`internal/update/extract.go`'s `extractAndReplace` generalizes from "find `lstk` in the temp dir, rename it over the executable" to a stage-then-commit **set** replacement:
+
+1. **Discover** the set at the extracted archive root: the lstk binary, every executable `lstk-*`, and `lstk-extensions.toml`. An archive with no extensions yields a set of size one — today's behavior, byte for byte.
+2. **Clean orphans**: remove any `*.lstk-new` siblings left by a previously crashed update.
+3. **Stage**: copy each member into the destination dir (the running executable's directory) as a `<name>.lstk-new` sibling — same directory ⇒ same filesystem ⇒ each upcoming rename is atomic — and set 0755 on binaries. Any failure here removes the staged files and leaves the installation untouched.
+4. **Commit**: rename each `.lstk-new` over its final name — extensions and the descriptions file first, `lstk` itself **last**, so the load-bearing swap is the final act and "update reported success" implies the whole set committed. Windows keeps the existing rename-running-exe-to-`.old` dance for `lstk.exe` only; extensions are not running during `lstk update` and rename directly.
```
peter-smith-phd: Migrating the `lstk` binary last seems to only matter if `lstk` refers to specific version of the extension. Given that it solely relies on the name, such as `lstk-doctor`, it doesn't really matter if `lstk` is first or last. The update of multiple files is simply going to be non-atomic, but since renames rarely fail (I've never seen it happen, when within the same directory), I'm not worried about this.

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,76 @@
...
-### Decision 3: Cross-channel packaging places bundled files where lstk resolves them
+### Decision 4: Set-wise update = stage-then-commit; additive-only on the binary channel
+
+Only the self-managed binary channel needs Go work — `brew upgrade` and `npm install -g` replace the whole package directory, which replaces the whole set (including removals and renames) by construction.
+
+`internal/update/extract.go`'s `extractAndReplace` generalizes from "find `lstk` in the temp dir, rename it over the executable" to a stage-then-commit **set** replacement:
+
+1. **Discover** the set at the extracted archive root: the lstk binary, every executable `lstk-*`, and `lstk-extensions.toml`. An archive with no extensions yields a set of size one — today's behavior, byte for byte.
+2. **Clean orphans**: remove any `*.lstk-new` siblings left by a previously crashed update.
+3. **Stage**: copy each member into the destination dir (the running executable's directory) as a `<name>.lstk-new` sibling — same directory ⇒ same filesystem ⇒ each upcoming rename is atomic — and set 0755 on binaries. Any failure here removes the staged files and leaves the installation untouched.
```
peter-smith-phd, line 58: What's missing from this discussion is that there's only one extensions binary (called `bundled-extensions`), and everything else is a symlink to that binary. At least, that's how it works in Unix-based systems. I think for Windows there are restrictions around symlinking, so a different solution might be required (needs more investigation?)

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,76 @@
...
+
+**Additive-only**: the binary-channel update replaces and adds members but never deletes an `lstk-*` sibling absent from the new archive. Deleting safely requires knowing lstk *owns* the file — users may place their own extensions next to the binary, and the descriptions file is not an ownership manifest (the spec deliberately permits undescribed bundled binaries). A renamed/dropped extension therefore leaves its old binary behind on this channel only: it keeps working at its old version and shows name-only help (the replaced descriptions file no longer describes it, so help never disagrees with the shipped set). Deletion is deferred to the managed-extensions-directory work. Corollary: if a user manually placed an `lstk-<name>` in the install dir and a release later bundles that same name, the update overwrites it — that directory is lstk's install dir; PATH is the supported home for user-installed extensions.
```
peter-smith-phd, line 63: I think we can solve this problem later, if we ever decide to remove an extension. For example, we could add custom `rm` code to explicitly remove the old extension. No need to worry about this now.

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,76 @@
...
+
+**Alternatives considered**: making release validation two-sided so the descriptions file becomes a complete ownership manifest enabling safe deletion (rejected here: contradicts the `extension-bundling` allowance for undescribed binaries; revisit with the managed dir); a separate shipped manifest file (rejected: a third shipped file to solve a rename edge case).
+
+### Decision 5: Update continuity is a hard requirement — no existing install may be cut off from `lstk update`
+
+The transition release (the first that ships bundled extensions) must be reachable by every in-the-field updater, and a later extension-free release must be reachable from a bundling one (rollback). Per channel:
```
peter-smith-phd, line 69: This is the thing that makes me the most nervous. We need careful planning (i.e. solid and reviewed test plan) to convince ourselves this will work. It would be a really bad thing if 1000s of customers were stranded with an `lstk` that couldn't be upgraded anymore.

For example, if we introduce bundling in `lstk` version N, but a user goes on holiday for a month, then tries to upgrade from `N-1` to `N+3` in a single upgrade, are we sure the upgrade will work successfully?

peter-smith-phd (commented):
Thanks for doing this! As I've mentioned several times, the update process is quite sensitive and we need to test it very carefully.  I'd suggest your next step is to generate a detailed test plan for the DevX team to review.

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: exactly, the extensions repo is private

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: IMHO it doesn't seem to be a complex thing and it's a safety net. Happy to remove this requirement if everyone agree is not needed.

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: Agree, problem for future Carlos and Peter 😆

openspec/changes/add-bundled-extension-distribution/proposal.md
carillan81: This was one of my points of discussion with Claude. The reasoning is to be able to track versions bundled together and have control when we are doing (hope not soon) braking changes. Blindly taking latest release could potentially produce unexpected results. In addition, it's not obvious to identify the version from the binaries. Pinning version is a fairly easy way to control versions in the bundle. I consider it easy enough to not be a real burden. Extensions developers should update this file when they want to release a new version.
If we want to remove this friction and just bundle latest version is fine, but we must make sure everyone is aware of this (merging anything to main will generate a new version in less than a week).

openspec/changes/add-bundled-extension-distribution/proposal.md
carillan81: I'm changing the spec to aknowledge this

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: Issues (I can think of) that could come of no pinning:
An extension can ship before lstk is ready for it.
A broken extensions build breaks the lstk release.
The same lstk version can contain different extensions (if version is repackaged for any reason)
Undoing a bad extension needs a new extensions release.
No record of what shipped. If a customer reports a bug in lstk-deploy, you can't tell from their lstk version which extension build they have. You'd be guessing from dates.
Merging to the extensions repo means shipping to everyone.

If we're comfortable with "if extensions break, we cut a new release," always-latest is a reasonable call and it is genuinely simpler. If everyone is aligned in this, I'm fine with always-latest approach.

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: I added a section in the design to decide actual delivery mechanism as symlinks might be a problem here.

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: Actually, this might be a requirement if we make extensions mandatory for the update and must fail if extensions fail. 🤔

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: test plan added to tasks

carillan81:
I documented a pending decision about symlink distribution and how to fix it with some proposals
Added the requirement of update failure in extension failure
Pending: pinned version removal if there is agreement

openspec/changes/add-bundled-extension-distribution/proposal.md
peter-smith-phd: >  If we want to remove this friction and just bundle latest version is fine, but we must make sure everyone is aware of this (merging anything to main will generate a new version in less than a week).

Yes, this is normal behaviour in a CI/CD environment. The rule is that whenever you hit merge, it could be in the customer's hands within a few hours. So basically we should feel confident enough that our testing is solid, and that we're not pushing out bad code. If we don't have that confidence, then we need to fix our processes/culture 😄 

Having said that, if we could set the default version to `latest`, then that would alleviate the need update it manually, but give us the ability to set it to `<explicit version>` if we did want to lock it down to an older version.  I think that's a reasonable trade-off.

openspec/changes/add-bundled-extension-distribution/design.md
peter-smith-phd: > An extension can ship before lstk is ready for it.

This is a general problem when coordinating software, and it just requires planning. In this case, we update `lstk` to have the new code, _before_ we merge code in the extensions. It's also mandatory that our pipeline tests against the latest `lstk`, rather than something in a private branch.

> A broken extensions build breaks the lstk release.

Then we have a problem with our testing, and we need to improve. It should _not_ be able to break _all_ of `lstk` though, but just the extension's sub-command itself.

> The same lstk version can contain different extensions (if version is repackaged for any reason)

I'm not sure I understand this one. Could you give an example?

> Undoing a bad extension needs a new extensions release.

You mean, a new `lstk` release? Yes, that's true, but I don't think we want to start deploying extensions separately, since that's more of a headache for customers. 

> No record of what shipped. 

We can fix this by having the `lstk` build process clearly state (in the logs) which Git hash from `lstk-bundled-extensions` was included.

> If a customer reports a bug in lstk-deploy, you can't tell from their lstk version which extension build they have. You'd be guessing from dates.

Yes you can, because each version of `lstk` is shipped with exactly one version of the bundled extensions, and we should log the Git hash in the `lstk` build logs. In most cases it'll be the latest `lstk` (because of auto-update), so it would be rare to go back and fix bugs in older versions.

> Merging to the extensions repo means shipping to everyone.

Yes, that's how CI/CD works 😄

gtsiolis (approved):
Solid expansion of the bundled-extension distribution spec — per-channel packaging (archive root / Homebrew cask staging / npm platform packages), pinned-and-checksum-verified provenance, and cross-transition update continuity are all now spelled out. This is spec/design only, no code.

1. **thought(non-blocking):** on `specs/extension-bundling-distribution/spec.md` — the "incomplete bundled set is repaired when lstk is already current" scenario requires `internal/update` to detect an incomplete set with no version change, i.e. to know the expected set and re-fetch the archive even when versions match. That's a real departure from today's version-only gating; worth making explicit in the test plan @peter-smith-phd asked for, since it's the subtlest part of the update path.
2. **praise:** the provenance requirement (a committed pin file plus hard-fail checksum verification before staging) mirrors the existing self-update SHA-256 archive verification, and the public-visibility note — bundled binaries are downloadable, so any gating must be runtime via `LSTK_EXT_CONTEXT`, not distribution — is exactly the right thing to pin down now.

---

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,97 @@
...
+
+**Validation targets a single, host-native staging dir** — descriptions are os/arch-independent, so the check runs once against the release runner's own platform staging dir (`linux_amd64`), where binaries are bare `lstk-<name>` with no `.exe`/PATHEXT ambiguity.
 
 **Rationale**: validating (not generating) keeps one source of truth in the private repo while preserving the version-lock guarantee; key-only parsing keeps the shell check trivially correct.
 
 **Alternatives considered**: a Go validator reusing the runtime `scanDir` + go-toml (rejected: extra build entrypoint and against the repo's "domain logic in Go, helpers in shell" grain for a small set-difference); generating the file from an in-repo manifest (rejected: duplicates a list the private repo already owns).
 
-### Decision 3: Cross-channel packaging places bundled files where lstk resolves them
+### Decision 4: Set-wise update = stage-then-commit; additive-only on the binary channel
```
anisaoshafi, line 50: I like the "stage-then-commit" approach 💯

openspec/changes/add-bundled-extension-distribution/design.md
```diff
@@ -1,27 +1,97 @@
...
+
+The `archives.files` entries land commented until the private pull is wired, then pull + payload are enabled **in one PR**: the PR-level `goreleaser check` job only validates config syntax (globs are not evaluated), but `goreleaser release/build` fails on a glob with zero matches, so the entries must never be live without the staging step that populates `bundled/`. After enabling, local snapshot builds require `scripts/fetch-bundled-extensions.sh` first; its `--stub` mode stages placeholder files for local, never-published builds so contributors without the private-repo credential can still run `goreleaser` locally. The staging tree is gitignored (only the pin file is tracked); it deliberately does not live under `dist/`, which `goreleaser --clean` wipes at startup.
+
+The first bundling release ships with the smallest viable bundle (a single extension) and is verified against the release-candidate checklist in `docs/extensions-bundling.md` — fresh install and upgrade-from-previous on all three channels — before further extensions are added to the bundle.
+
+### Decision 7: Bundled binary layout — OPEN, needs a call before implementation
```
anisaoshafi, line 86: Since decision 7 on bundled binary layout is not decided yet (2 alternatives proposed), it affects some other decisions where it's linked (1, 4, 5 for example), while the test plan is written considering alternative a).
Could we remove this and the plan around it for now? We can open another PR later when it's clearer, and once we've experimented with the doctor extension for example.
Or do you think we have to make that decision now?

openspec/changes/add-bundled-extension-distribution/tasks.md
```diff
@@ -1,22 +1,120 @@
-## 1. On-disk layout and packaging
+> **How to read this plan.** Sections 1–3 have no dependency on the private extensions repo — they can be built and merged right away (section 1 changes nothing user-visible until archives actually contain extensions). Section 4 needs the private repo's owners and is the long-lead item. Section 5 turns packaging on and MUST merge in the same PR as the CI step from 4.3 — never separately (see design Decision 6 for why a half-enabled state breaks releases). Sections 6–7 are verification and docs.
```
anisaoshafi: Task 4.3 & section 5 must ship together (a CI step that downloads the extension files into a bundled/ folder; and a config change indicating the release tool to grab files from that bundled/ folder and package them).

Only a comment mentions that shipping one without the other, will break the releases. 
Can we add an actual check that catches this automatically, instead?

anisaoshafi (approved):
Thanks for taking care of the extensions architecture and focusing on the update mechanism, Carlos 🙏🏼 

I like the `stage-than-commit` approach, to not break the update.

I tried to understand the delivery mechanism, though I have a bit of a hard time reading tech specs in text only. I might miss many details, and you have better context than me already.
Things will become clearer once doctor implementation starts and when we actually that extension. We can still tweak parts of the design later if needed.

I think the extension you made to the design makes sense. Added some comments for your consideration.

openspec/changes/add-bundled-extension-distribution/proposal.md
carillan81: `latest` as the default with the option to set an explicit version works for me — that gets us out of the routine maintenance while keeping a way to lock down when something's wrong. I'll rewrite the decision that way.
One thing it doesn't cover, which is also the example you asked for in the other thread: `latest` re-resolves on every build, including a re-run of an already-tagged release. Cheap fix I'd like to fold in — have the release job resolve `latest` to a concrete tag once and record it in the release notes, so the mapping from an lstk version to an extensions build is permanent and a re-run can be pointed back at it. That also covers your "log the git hash" point, but somewhere that doesn't expire the way build logs do.

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: I agree we can just accept this. If we need to regenerate a version we will point the commit as commented in the other file.

openspec/changes/add-bundled-extension-distribution/tasks.md
carillan81: Agreed, a comment isn't enough. I'll add a task for a CI check that fails when the two halves are out of step — packaging entries pointing at \`bundled/\` with no fetch step wiring it up, or the reverse.

openspec/changes/add-bundled-extension-distribution/design.md
carillan81: I'd rather keep it and scope it properly than drop it. The symlink problem is easy to rediscover the hard way, and the analysis is the useful part.

Decision 7 blocks section 5 (packaging) only. Sections 1–3 don't depend on the layout and can be built and merged now, so it holds nothing up. We can settle it once doctor is real and we know whether it genuinely is one binary and how big it is; which is exactly the input the decision is missing today.

I'll also note in the test plan that the cases naming individual `lstk-<name>` files are written for option (a) and need rephrasing, not rethinking, under (b).

PR #456, opened 2026-08-13
Ignore stale Rancher Desktop during detection

gtsiolis:
## Motivation

When Docker is unavailable, lstk treats any existing `~/.rd` directory as evidence that Rancher Desktop is installed. Rancher Desktop leaves this directory and broken CLI symlinks behind after uninstalling, which causes lstk to recommend an unusable `rdctl start` action.

Here's what I was seeing on start.

Screenshot 2026-08-13 at 12 59 26 copy 2

## Solution

Use a resolvable `rdctl` executable on `PATH` as fallback installation evidence instead of the Rancher state directory alone. Configured Rancher Desktop socket detection remains unchanged.

Add regression coverage for both an installed-but-stopped Rancher Desktop and stale `~/.rd` state after uninstalling.

<details>
<summary>Docs</summary>

No documentation changes are needed. This corrects a runtime detection false-positive without changing the documented container-runtime workflow.

</details>

## Review

Self-merge candidate: this is a small, focused detection fix with regression coverage and no new user-facing behavior. A human review is optional unless the team wants to revisit the existing cross-platform assumption that Rancher Desktop installs `rdctl` on `PATH`.

Closes PRO-374

### DIFF
--- a/internal/runtime/docker.go
+++ b/internal/runtime/docker.go
@@ -319,7 +319,11 @@ func detectRuntimeFlavor(
 	if configuredFlavor != flavorUnknown && configuredFlavor != flavorDockerNative {
 		return configuredFlavor
 	}
-	if _, err := statFn(filepath.Join(home, ".rd")); err == nil {
+	// Rancher Desktop leaves ~/.rd behind when the application is removed. Its
+	// rdctl executable is installed on PATH on every supported platform, and is
+	// stronger evidence that Rancher Desktop is still installed than the state
+	// directory alone.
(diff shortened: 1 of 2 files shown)


gtsiolis:
@skyrpex could you take a look since you added this in https://github.com/localstack/lstk/pull/399. Cc @anisaoshafi

skyrpex (approved):

skyrpex:
LGTM, honestly a much better heuristic 👍

gtsiolis:
Thanks for taking a look, @skyrpex! 🙏

PR #458, opened 2026-08-13
Render the license re-login prompt choices vertically

tiurin:
## Motivation

#449 gave the license re-login prompt a second, advertised option (`ESC to exit`), but left it on the horizontal rendering path. Both labels are flattened into one dimmed hint and appended to a question that already wraps, so the choices read as prose glued to the end of a sentence — and the sentence's own tail (`Log`) dangles at the wrap point. #457 fixed exactly this for the slow-start recovery prompt; this applies the same treatment to the re-login prompt.

## Before / After

Before:

image

After:

image

---

| Aspect | Before | After |
| --- | --- | --- |
| Choices | Flattened into one hint: `[ENTER to log in again/ESC to exit]` | One line each: `● [ENTER] Log in again` / `○ [ESC] Exit` |
| Prompt text | Statement + question, wrapping mid-sentence with `Log` left dangling | Statement only; the actions live in the labels |
| Selection | Key press only | ↑/↓ + Enter, or the advertised key directly |

<details>
<summary>Docs</summary>

Nothing to document. This changes the layout and wording of one interactive prompt; no command, flag, env var, or documented behavior changes. The decline key advertised in #449 still works the same way, and the manual recovery steps a decline prints are unchanged.

</details>

Closes DEVX-1045

PR #457, opened 2026-08-13
Improve slow-start prompt behavior

gtsiolis:
## Motivation

When LocalStack takes longer than the interactive startup timeout, the recovery prompt wraps awkwardly and hides its shortcut keys. More importantly, startup health checks pause while the prompt is open, so it can remain stuck after the emulator is ready and a stale Stop selection can stop a healthy emulator.

## Solution

- Render the slow-start actions as vertical, shortcut-first choices.
- Continue monitoring readiness and container exit while the prompt is visible.
- Dismiss only the matching prompt when startup resolves automatically.
- Recheck readiness before stopping to protect an emulator that became healthy during the prompt.
- Add regression tests for layout, automatic dismissal, and readiness/Stop races.

## BEFORE / AFTER

| Aspect | Before | After |
| --- | --- | --- |
| Prompt layout | Question and actions form one long line and wrap awkwardly | Question and actions render on separate lines |
| Shortcuts | `Keep waiting [W]` / `Stop LocalStack and exit [S]` | `[W] Keep waiting` / `[S] Stop and exit` |
| Successful startup | Prompt remains stuck until a key is pressed | Prompt closes automatically and the success sequence appears |
| Stop action | A stale prompt can stop an emulator that has become ready | Readiness is checked again before stopping |
| Monitoring | Health checks pause while awaiting input | Health and exit monitoring continue behind the prompt |

| BEFORE | AFTER |
|--------|--------|
| Screenshot 2026-08-13 at 13 36 25 | Screenshot 2026-08-13 at 13 58 06 | 

## Docs

<details>
<summary>Docs assessment</summary>

No documentation changes are needed. This refines existing interactive startup behavior and keeps the recovery command visible in the prompt.

</details>

## Review

Self-merge candidate: this is a small, already-discussed UX fix with focused unit and race regression coverage.

## Validation

- `go test -race ./internal/container -run 'TestStartupMonitorAwait_(InteractivePromptKeepWaitingThenStop|DismissesPromptWhenEmulatorBecomesReady|DoesNotStopEmulatorThatBecameReadyBeforeSelection)$' -count=1`
- `go test ./internal/ui/... ./internal/output/... -count=1`
- `make build`
- `git diff --check`

Closes DEVX-1051

Co-Authored-By: Claude <noreply@anthropic.com>

PR #459, opened 2026-08-13
refactor: Build user input prompts with intent constructors

tiurin:
## Motivation

#457 and #458 each fixed one prompt that had shipped with the wrong layout, and the [review comment on #457](https://github.com/localstack/lstk/pull/457#issuecomment-5284351655) wrote down the rule they were both applying:

> - Inline Y/n for binary confirmations: "Remove snapshot? [y/N]"
> - Vertical choices for distinct actions: "Keep waiting", "Stop and exit", etc.

That rule is not written anywhere. `UserInputRequestEvent` asked its author for a `Vertical bool`, which is a rendering decision. It is easy to leave the field out which defaults to inline.

Also, labels had drifted into three styles at once: `[ENTER] Log in again`, `Update now [U]`, and a bare `AWS` advertising no key at all.

## Solution

Give user input a vocabulary of intents and derive the layout from it. User promts can be built using one of the new constructors in `internal/output/prompt.go`. The developer uses the appropriate prompt type which automatically applies the design rule :

| Constructor | What it is | Rendering |
| --- | --- | --- |
| `output.Confirm(prompt, DefaultYes\|DefaultNo, ch)` | y/n on an action the user already requested | inline `[y/N]`, capitalized answer is what ENTER picks |
| `output.ActionChoice(prompt, options, ch)` | a choice between distinct outcomes | one selectable row per option |
| `output.Acknowledge(prompt, label, ch)` | a single keypress, no choice | inline |

`UserInputRequestEvent` can't be used directly, but can be easily extended with a new prompt type if needed. See technical details:

<details>
<summary>Technical details</summary>

To enforce the usage of constructors, `UserInputRequestEvent`'s fields are unexported - this is done by renaming variables from uppercase to lowercase and read-only accessors are added. This way a struct literal built anywhere outside `internal/output` does not compile:

```
internal/reset/reset.go:43:36: unknown field Prompt in struct literal of type
output.UserInputRequestEvent, but does have unexported prompt
```

Unexporting only `Vertical` would not have worked — Go lets a keyed literal from another package omit unexported fields, so the bypass would still compile and would still default to the inline layout that caused the bug. Removing every exported field is what closes it. `UserInputRequestEvent{}` remains legal but inert: no prompt, nil channel, nothing settable.

</details>

## Behavior changes

The default has been changed to No for `lstk snapshot remove`. Rationale - other destructive operations, e.g. `lstk reset`, `lstk volume clear`, have No as a default - meaning that accidentally pressing Enter does not perform an operation that cannot be undone. I've made snapshot remove consistent with the rest. See comments in review below: https://github.com/localstack/lstk/pull/459#discussion_r3804630907

## Review

I've reviewed the code and tested the behaviour manually. I've also shortened generated comments and claude instructions.

I've added a self-review with examples of inconsistencies that this PR fixes and prevents in the future.

## Validation

- `go test ./internal/... ./cmd/... -count=1`
- `golangci-lint run ./...` — 0 issues
- `make build`
- Seal proved by adding a raw literal to `internal/reset/reset.go` and confirming `go build ./...` rejects it

Follow-up to #457 and #458.

Closes DEVX-1057


internal/container/start.go
```diff
@@ -1629,15 +1627,14 @@ func (m *startupMonitor) await(ctx context.Context, containerID, healthURL strin
...
-					{Key: "s", Label: "[S] Stop and exit"},
+			m.sink.Emit(output.ActionChoice(
+				"LocalStack is still starting. Check progress with 'lstk logs'.",
+				[]output.InputOption{
+					{Key: "w", Label: "Keep waiting"},
+					{Key: "s", Label: "Stop and exit"},
 				},
-				ResponseCh: responseCh,
-				Vertical:   true,
-			})
```
tiurin, line 1640: This shows well the key benefits of the change.
- name talks about a specific intent
- rendering details that are easy to miss (`Vertical`) are hidden behind `ActionChoice` implementation.
- Key rendering like `[W]` is derived from the key itself, doesn't leave a place for drift

internal/reset/reset.go
```diff
@@ -41,18 +41,11 @@ func Reset(ctx context.Context, rt runtime.Runtime, containers []config.Containe
 
 	if !force {
 		responseCh := make(chan output.InputResponse, 1)
-		sink.Emit(output.UserInputRequestEvent{
-			Prompt: "Reset emulator state? All resources will be lost",
-			Options: []output.InputOption{
-				{Key: "y", Label: "Yes"},
-				{Key: "n", Label: "NO"},
```
tiurin, line 48: Was inconsistent with the rest of y/n prompts: `Yes/NO` here, `y/N` everywhere else.

internal/container/select.go
```diff
@@ -19,12 +19,7 @@ func SelectEmulator(
 	}
 
 	responseCh := make(chan output.InputResponse, 1)
-	sink.Emit(output.UserInputRequestEvent{
-		Prompt:     "Which emulator would you like to use?",
-		Options:    options,
```
tiurin, line 24: Actual shortcut keys to select the emulators were not displayed because `ShortName()` didn't contain the text:

Image

It's hard to understand while writing the code because `InputOption` receives both key and label, so agents often assume label doesn't need to contain the key shortcut prefix in their text.

With the refactoring the rendering is left to the `ActionChoice` implementation and is consistent between usages - shortcut keys are always displayed if provided. prevents the problem and keeps `ShortName()` clear and without a possibility to drift from defined shortcut key to what is typed in the label

internal/update/notify.go
```diff
@@ -75,12 +75,11 @@ func promptAndUpdate(ctx context.Context, sink output.Sink, opts NotifyOptions,
 	sink.Emit(output.MessageEvent{Severity: output.SeveritySecondary, Text: fmt.Sprintf("> Release notes: %s", releaseNotesURL)})
 
 	responseCh := make(chan output.InputResponse, 1)
-	sink.Emit(output.UserInputRequestEvent{
-		Prompt:     "Update lstk to latest version?",
-		Options:    []output.InputOption{{Key: "u", Label: "Update now [U]"}, {Key: "r", Label: "Remind me next time [R]"}, {Key: "s", Label: "Skip this version [S]"}},
```
tiurin, line 80: Here key shortcut was printed as a suffix (`Update now [U]`), while in every other user prompt it is a prefix (e.g. `[R] Re-authenticate`).

Leaving rendering the shortcuts to `ActionChoice` implementation rather than typing it out in a label string keeps the formatting consistent across CLI.

internal/output/events.go
```diff
@@ -271,13 +271,30 @@ type InputResponse struct {
...
 
+// Base struct for all user input requests. Not to be instantiated directly.
+// Build user input requests with typical constructors defined in prompt.go,
+// e.g. Confirm, ActionChoice, or Acknowledge.
 type UserInputRequestEvent struct {
-	Prompt     string
-	Options    []InputOption
-	ResponseCh chan<- InputResponse
-	Vertical   bool
+	prompt     string
```
tiurin, line 278: Keeping fields package-private by using lowercase names. This means `UserInputRequestEvent` can't be meaningfully instantiated directly. `prompt.go` is in the same package so new constructors can set these fields when they build an instance.

internal/snapshot/remove.go
```diff
@@ -49,18 +49,15 @@ func Remove(ctx context.Context, rt runtime.Runtime, containers []config.Contain
...
-			Prompt: fmt.Sprintf("Delete cloud snapshot 'pod:%s'? This operation cannot be undone.", podName),
-			Options: []output.InputOption{
-				{Key: "y", Label: "Y"},
-				{Key: "n", Label: "n"},
-			},
-			ResponseCh: responseCh,
-		})
+		sink.Emit(output.Confirm(
+			fmt.Sprintf("Delete cloud snapshot 'pod:%s'? This operation cannot be undone.", podName),
+			output.DefaultNo,
```
tiurin, line 54: There is a behaviour change here - the default is changed to No. Rationale - other destructive operations, e.g. reset emulator, clear volume, have No as a default - meaning that accidentally pressing Enter does not perform an operation that cannot be undone. I've made snapshot remove consistent with the rest here - @anisaoshafi would love to hear your opinion since you worked on snapshots recently IIRC.

internal/output/events.go
```diff
@@ -271,13 +271,30 @@ type InputResponse struct {
...
+	options    []InputOption
+	responseCh chan<- InputResponse
+	// vertical renders each option as its own selectable row instead of a
+	// trailing "[a/b]" hint.
+	vertical bool
 }
 
+func (e UserInputRequestEvent) Prompt() string { return e.prompt }
```
tiurin, line 286: Adding accessors for fields that are package-private now - they need to be read by rendering code in other packages, e.g. in `app.go`.

tiurin (commented):
I've added examples of inconsistencies that are fixed and should be prevented in the future by this refactoring.

gtsiolis (approved):
Clean refactor moving `UserInputRequestEvent` construction behind three intent constructors (`Confirm`/`ActionChoice`/`Acknowledge`) with private fields, migrating every call site and its tests, and centralizing shortcut derivation in `OptionLabel`.

1. **thought(non-blocking):** on `internal/snapshot/remove.go` — this quietly changes user-facing behavior: the old delete prompt used labels `{Y, n}`, so ENTER confirmed the delete, whereas `Confirm(DefaultNo)` makes ENTER cancel it. That's strictly safer for an irreversible op and is covered by the new `[y/N]` assertion, but a behavior change riding inside a "refactor" PR is easy to miss — worth a line in the description so reviewers know it's intentional. (Same applies to `reset`/`volume clear`, though those already defaulted to No.)
2. **praise:** the coverage migration is careful — the app-level tests that could only be built from hand-rolled option shapes are removed, and their rules (explicit-`enter` priority over an uppercase default, non-letter labels, `any`-key priority) are already exercised directly against `resolveOption` in `TestResolveOption`, so nothing is actually lost.

---

tiurin:
> * **thought(non-blocking):** on `internal/snapshot/remove.go` — this quietly changes user-facing behavior: the old delete prompt used labels `{Y, n}`, so ENTER confirmed the delete, whereas `Confirm(DefaultNo)` makes ENTER cancel it. That's strictly safer for an irreversible op and is covered by the new `[y/N]` assertion, but a behavior change riding inside a "refactor" PR is easy to miss — worth a line in the description so reviewers know it's intentional. (Same applies to `reset`/`volume clear`, though those already defaulted to No.)

👍 This change was already mentioned in the self-review: https://github.com/localstack/lstk/pull/459#discussion_r3804630907. But I agree it is worth highlighting in the PR description. Added.

carole-lavillonniere (approved):
Much cleaner, great work 👏

internal/snapshot/remove.go
anisaoshafi: I think no as default is the right way, plus making it consistent with the rest of commands.
Thanks for correcting that, and for the nice work 👏🏼

PR #464, opened 2026-08-18
Add snowflake-next preview emulator type

hovaesco:
## Motivation

lstk only knows the Python Snowflake emulator: `type = "snowflake"` means `localstack/snowflake`. Running the rewritten emulator meant hand-overriding `image`/`tag` and remapping env — a recipe that lived in a Slack thread and that every tester had to rediscover.

## Solution

Adds `snowflake-next` as a preview emulator type, so `lstk -t snowflake-next` just works and records the choice in config like any other type. It is deliberately absent from the interactive first-run picker — a preview is reachable only by asking for it explicitly, so a fresh install's first choice stays a GA product.

The type is now a plain registry entry with no per-emulator special-casing. Earlier revisions carried adaptation code — rewriting the listen address onto the gateway port, bind-mounting the image's own `VOLUME`, placing `PGDATA` — because the image diverged from every other emulator. localstack/snowflake-rs#2245 made it a container-level drop-in for `localstack/snowflake`: it binds from `GATEWAY_LISTEN`, declares `/var/lib/localstack` as its volume, and picks its data dir from `LOCALSTACK_PERSISTENCE`. lstk's generic start path already covers all three, so that code is gone — which is why this diff is ~250 lines smaller than it was.

Found and fixed along the way: the IaC proxies and the emulator list handed to extensions enumerated only the *selectable* types, so `lstk terraform` against a running preview reported the misleading `LocalStack AWS Emulator is not running`.

Verified against the published image (revision `994f10d`): start answers health on the configured host port, `--persist` survives a container recreation, and `lstk volume clear` still resets state.

<details>
<summary>Docs</summary>

User-facing, worth documenting:

- **New emulator type** `snowflake-next`, a preview of the next Snowflake emulator. Selected with `lstk start --type snowflake-next` / `lstk -t snowflake-next`, or `type = "snowflake-next"` in `config.toml`. Displayed as "LocalStack Snowflake Preview Emulator". Default image `localstack/snowflake-next:latest`.
- **Intentionally absent from the interactive first-run picker** — reachable only by asking for it explicitly. Worth stating plainly, since it is otherwise undiscoverable.
- **Behaves like the other emulators**: `--persist` keeps state across restarts, `lstk volume clear` resets it, `--endpoint-url` can target a remote instance. No type-specific caveats.
- **Migration note for early testers**: the hand-rolled `image`/`tag`/`SNOWFLAKE_LISTEN_ADDR`/`PGDATA` config from the preview period is no longer needed — replace it with `type = "snowflake-next"`. A custom `image` remains a hard error when switching types.
- A remote `snowflake-next` reached over `--endpoint-url` is reported under the GA Snowflake emulator's name; the two are indistinguishable from the health payload.

</details>

## Review

Human review advised — new user-facing emulator type, and it depends on the image contract from localstack/snowflake-rs#2245 holding.

Towards LAV-595


gtsiolis (approved):
This wires the `snowflake-next` preview emulator type end to end — reachable only via `--type`/config (never the first-run picker), with its port remap and PostgreSQL persistence handled on the start path, plus integration tests and every non-compiler-checked map/slice touch-point updated.

1. **suggestion(non-blocking):** on `internal/emulator/snowflake/snowflake.go` / `internal/container/start.go` — the `--persist` uid-1000 caveat on Linux native Docker is well documented in code, but has no user-facing signal: when it bites, the user sees only an opaque PostgreSQL "refuses to start" surfacing as a health-check timeout. A note/warning on `--persist` for `snowflake-next` on Linux pointing at host-directory ownership would save a confusing debugging session. Preview-acceptable to defer.
2. **question(non-blocking):** on `cmd/status.go` — `snowflake-next` maps to `snowflake.NewClient()`, but per the comment you added in `internal/endpoint/target.go` the preview's health payload carries a version and no `services` map. Does `lstk status` render correctly against that payload, or could it under-report? I did not verify this against a running preview image (the licensed image wasn't available to me).
3. **praise:** thorough TDD — the integration tests pin the port remap, both persistence paths, the first-run picker omission, and the IaC-rejection naming, and every entry the compiler can't catch (`knownImages`, `emulatorHealthPaths`, `ContainerPort`, `SelfValidatesLicense`, `emulatorDisplayNames`, the `cmd/status.go` client map, `tipsForType`) is covered.

---

peter-smith-phd (changes requested):
I understand what the change is trying achieve, but could we please have some upfront discussion on how to solve it? (Please forgive me if I missed the Slack message). I want to make sure we have a clean/elegant solution to this, rather than just asking Claude to make a whole ton of changes.

Some important questions I have:
1) Are customers actively using the new Snowflake emulator, or is this for internal use only? What's the timeline for making the new Snowflake emulator the default?
2) This PR touches 21 files and changes 601 lines of code, which seems excessive given that we're essentially making 3 small configuration changes.  When I look at the code changes, I see a lot of places where a list of 3 things becomes a list of 4 things. That implies we should be centralizing those values in a common place, not just repeating the list everywhere.

hovaesco:
1. The Snowflake Rust (Next) emulator is not yet being officially released to the public, but several prospects and customers are waiting to test it. We currently only have internal users and beta testers, and the lstk integration is one of the remaining pieces needed for a public release. The sooner we can launch the preview, the better. We’re tentatively targeting GA for early next year, depending on feedback from the preview phase.

2. To be precise, the 603 added lines include approximately 406 lines of tests and snapshots. The runtime changes also extend beyond three configuration values. The accepted type set is centralized in `KnownEmulatorTypes()`, while the other occurrences represent behavior owned by different subsystems rather than duplicated copies of the same list. I’m open to consolidating more of this into a descriptor registry, but that would be a broader architectural refactor. If that is required for this PR, could you please suggest the specific abstraction you have in mind?

PR #473, opened 2026-08-27
Add S3-specific endpoint to generated AWS profile

anisaoshafi:
## Motivation
`aws --profile localstack` used directly (not through lstk's own proxies) sent S3 requests to the profile's general endpoint, breaking S3's virtual-host-style bucket addressing.

## Solution
`lstk setup aws` now writes an S3 endpoint override in a [services localstack] block, referenced via services = localstack in the profile. It reuses the existing endpoint.S3Addressing helper already used by the Terraform/CDK proxies. This block follows the same idempotency and `--force` overwrite rules as the rest of the profile.

## Review
Small, self-contained, covered by unit and integration tests. Self-merge candidate.

Closes DEVX-755

### DIFF
--- a/internal/awsconfig/awsconfig.go
+++ b/internal/awsconfig/awsconfig.go
@@ -19,9 +19,10 @@ import (
 )
 
 const (
-	ProfileName       = "localstack"
-	configSectionName = "profile localstack" // ~/.aws/config uses "profile <name>" as section header
-	credsSectionName  = "localstack"         // ~/.aws/credentials uses just the profile name
+	ProfileName         = "localstack"
+	configSectionName   = "profile localstack"  // ~/.aws/config uses "profile <name>" as section header
+	credsSectionName    = "localstack"          // ~/.aws/credentials uses just the profile name
+	servicesSectionName = "services localstack" // referenced by the "services" key in configSectionName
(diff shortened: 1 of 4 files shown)


joe4dev (approved):
Good to see S3 compatibility addressed in the AWS profile. That's particularly important to cdk. I really like the non-destructive edit in `~/.aws/profile` 👏 ✨  

A couple of non-blocking questions/suggestions:

❓ Do we want to offer a fallback for customers being blocked by DNS rebind protection, similar to the probing we do in the LocalStack Toolkit https://github.com/localstack/localstack-toolkit-vscode/pull/51?

> This block follows the same idempotency and --force overwrite rules as the rest of the profile.

❓ What is the intended behavior of `--force`?
Given an existing `~/.aws/profile`, both non-force and --force performed a non-destructive edit. Does `--force` simply skip the interactive "[Y/n]" prompt?

nit: The confirmation message "✔︎ Created LocalStack profile in ~/.aws/config" does not distinguish between Create and Update.

## Sidenotes

Not directly related, just FYI:

* The `lstk aws` proxy command does not respect the `--profile localstack`. Is this intentional or is there a workaround?
* The AWS VSCode Extension cannot parse the service-specific section and spams verbose logs (their bug):
    Image

PR #476, opened 2026-08-31
Feed keyboard input into the wrapped-tool PTY

dominikschubert:
## Motivation

When `lstk aws` or `lstk az` pages long output on a terminal, the pager draws its first page and then ignores every keystroke — SPACE, ENTER, and `q` do nothing and the command looks frozen. The inner PTY introduced for output streaming (DEVX-1026/1028) was output-only, while `less` ≥577 reads keyboard input from the terminal attached to its stderr — that PTY, which nothing fed. The result was that the whole terminal ultimately became unresponsive, obviously not a great user experience 😄 

## Solution

When lstk's stdin is a terminal, the inner PTY becomes the wrapped tool's stdin and controlling terminal (`Setsid`+`Setctty`): the user's terminal switches to raw mode and keystrokes are pumped into the PTY, whose line discipline — under the child's control — handles echo, buffering, and signal keys. Ctrl-C reaches the child via the PTY; redirected stdin (`yes | lstk aws ...`) keeps today's byte-exact passthrough. The pump reads through a cancelreader so teardown joins it and restores the terminal deterministically. The accepted trade-off (^Z becomes a no-op, as with `ssh`/`docker run -it`) is pinned in a test.

The full design — two prototyped architectures, measured evidence, and an adopter survey for `terraform`/`cdk`/`sam` — is in `openspec/changes/fix-wrapped-tool-pty-input/design.md`. Tests cover keystroke forwarding, terminal restoration, pump teardown, redirected-stdin passthrough, and end-to-end pager keystrokes for both proxies via a new faketool pager mode.

<details>
<summary>Docs</summary>

Nothing to document: this restores expected pager/keyboard behavior for `lstk aws` and `lstk az` on a terminal — no new commands, flags, or configuration. The only observable trade-off (Ctrl-Z does not suspend an interactive `lstk aws`/`lstk az` run) matches common wrappers like `ssh` and `docker run -it` and is documented in the openspec design.

</details>

## Review

Human review advised — changes interactive terminal semantics (raw mode, new session/controlling terminal, ^Z no-op trade-off), and the openspec design decision is proposed pending team discussion.

## Additional Human Notes 🧔‍♂️ 
This PR includes both a design doc to explicitly decide between two similar architectures as well as an existing implementation of the suggest way forward. The reason for this is that I focused on the fix first, branched off with a different model and found the difference impactful enough to surface.

For that reason, I recommend reviewing the openspec files first. If preferred I can also split the PR if it facilitates easier reviewing.

Closes DEVX-1049


CLAUDE.md
```diff
@@ -176,9 +176,9 @@ lstk supports Git-style extensions: when `lstk <name>` is not a built-in command
...
...e, and the frozen Python aws CLI then block-buffers stdout (8 KB, ignoring `PYTHONUNBUFFERED`) — streaming commands like `aws logs tail --follow` showed nothing until exit (DEVX-1026). The PTY makes the child see a terminal (line-buffered, colored output; stdout/stderr merged) while the master side is copied through the spinner's `StopOnWriteWriter`. When lstk's stdin is also a terminal, the PTY additionally becomes the child's stdin and controlling terminal, with the user's terminal switched to raw mode and its keystrokes pumped into the PTY — without that, the aws CLI's pager read keyboard input from the PTY (less ≥577 reads the terminal attached to its stderr) and looked frozen (DEVX-1049); Ctrl-C then reaches the child as SIGINT via the PTY's line discipline rather than the shared process group (see `wireInteractiveStdin`). With stdin redirected, stdin passes through untouched and no new session is created, so piped data never crosses a line discipline and Ctrl-C still arrives via the process group. Falls back to plain `proc.Run` when no PTY can be allocated (Windows, see `pty_windows.go`), and stays on pipes whenever stdout is redirected so pipelines never receive colors/CRLF.
```
dominikschubert, line 179: I don't think this kind of information should be in the `CLAUDE.md` file as it's against current best practices, but let's keep the existing pattern for now.

peter-smith-phd: I'd agree with you on this. We already have the OpenSpec files for this feature, so it can still be examined if necessary. I believe that the `CLAUDE.md` is scanned unconditionally, so it should only contain very generic information. However, the OpenSpec files are read _conditionally_ based on whether the feature is relevant to the current prompt.

peter-smith-phd (commented):
Thanks @dominikschubert, I tested this and it worked correctly for my test case.

As for the PR review, do you feel this needs a second human to review it? We're trying to move towards not having a second human reviewer, but if you don't feel comfortable with the whole PTY mechanism, then perhaps we need somebody else to look at it (I know, PTYs make my own head spin)

The "Human review advised" comment in the PR description made me wonder if you're asking for somebody, or whether that's left over from Claude's auto-generated PR description.

dominikschubert:
> As for the PR review, do you feel this needs a second human to review it? We're trying to move towards not having a second human reviewer, but if you don't feel comfortable with the whole PTY mechanism, then perhaps we need somebody else to look at it (I know, PTYs make my own head spin)
> 
> The "Human review advised" comment in the PR description made me wonder if you're asking for somebody, or whether that's left over from Claude's auto-generated PR description.

@peter-smith-phd 
It would be good to ultimately get a sign-off on the design/approach, but I can also showcase this internally next week and worst case we pick a different alternative. I'm not looking for review of the implemented solution though. Understanding the overall wrapping approach and PTY handling should be helpful for the team as this is a fairly fundamental feature of lstk. 

Regarding human review: This was definitely left in intentionally by myself but since the change here are generally reversible, we can always change our approach. It also doesn't introduce any new contract with a user.  Good point and thanks for flagging that 👍 

I'll keep the proposal in the openspec changes for now until I've presented the tradeoff and will move ahead with merging the suggested fix itself in the meantime.

PR #477, opened 2026-09-01
Ship LocalStack's bundled extensions in every release package

carillan81:
**NOTE:**
This PR needs to be merged when the extensions are released and never before. The moment this PR is merged, the package will include the extensions in the next release.

### Motivation
LocalStack's own extensions (lstk doctor, lstk deploy) exist but do not ship with lstk. Today someone has to place the files next to the binary by hand. This makes every release carry them, on all three install channels.

### Solution
The extensions ship as one binary, bundled-extensions, next to lstk. lstk tells it which extension to be by launching it under the name lstk-<name>, the same trick busybox and git use. Which commands exist comes from lstk-extensions.toml shipped beside it, so that file is required whenever the binary is present. If it is missing or broken, lstk says so and points at lstk update rather than reporting an unknown command.

The release pipeline downloads the bundle from the private extensions repo, verifies every file against its checksums, and refuses to release if the descriptions file and the binary disagree about which commands exist. The binary is asked directly for its command list (bundled-extensions list), so the toml can never promise a command the binary does not answer to. Goreleaser puts both files in the archives, the Homebrew cask inherits them, and a script copies them into each npm platform package.

A release ships exactly two files per platform: the binary and the descriptions file. There are no lstk-<name> alias links. lstk never used them to resolve anything, and they behaved differently on every channel (npm drops symlinks when packaging, Windows extractors turn them into text files), so dropping them makes all three channels identical. A leftover link from an older install is inert: lstk still takes its command list from the toml and dispatches by argv[0].

### Testing
New bash suites for the release scripts (make test-scripts), unit and integration tests for the dispatch, and a PR check that stops the two halves of the packaging config being merged separately. Verified against a real goreleaser build and a real brew and npm install.

<details> <summary><b>Docs</b></summary>
New docs/extensions-bundling.md: what ships, where the files live per install method, the release pipeline, what the private repo must publish, and how to diagnose a broken install. 

docs/extensions-authoring.md gains a section on how bundled extensions differ for authors. 

CLAUDE.md and the openspec change are updated.

No existing documented command changes, so nothing else needs updating.

</details>


docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+ship **inside every lstk release** as two files placed next to the `lstk`
+binary:
+
+| File | What it is |
+| --- | --- |
+| `bundled-extensions` (`.exe` on Windows) | One multi-call binary providing every bundled extension. It decides which extension to be from the name it is invoked as (`argv[0]`), the way busybox and git do. |
+| `lstk-extensions.toml` | A flat TOML table, `name = "one-line description"`, hand-written in the private extensions repository. For the bundle this file is load-bearing: it is the only record of which commands the binary provides. |
+
+lstk never learns about bundled commands from directory contents. `lstk <name>`
+reads the toml, sees `name` listed, and executes `bundled-extensions` with
```
peter-smith-phd, line 19: Interesting... I didn't know you can do this. I assumed that `argv[0]` was always just the name of the binary on the filesystem.

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+Why one binary rather than a copy per name: copies cost roughly 30 MB per
+extension in every archive and npm package, and symlink aliases do not survive
+the tar/zip extractors or Windows. One binary has none of those problems. The
+trade is that the toml must be present and correct whenever the binary is,
+which the release gate below enforces and the runtime treats as a broken
+install if violated.
+
+## Where the files live on disk
+
+lstk looks in one place: the directory of its own symlink-resolved executable
```
peter-smith-phd, line 57: Question: if `lstk` is on the user's path, then will `bundled-extensions` also be on the user's path? This matters, since the user could easily stumble across the `bundled-extensions` binary and try running it. In this case, we need to get a reasonable error message (e.g. `please run inside lstk`). Alternatively, call it `lstk-bundled-extensions` so it's obvious what it is.

Now that I read further in this doc, I'm not sure if this will be a problem for `npm` or `brew`, which both seem to symlink only this interesting binary (`lstk`).

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+   build to one bundle.
+2. **Fetch and verify.** `scripts/fetch-bundled-extensions.sh` resolves
+   `latest` to a concrete tag **once**, prints it, downloads that tag's release
+   assets with `gh release download`, and verifies every asset against the
+   `checksums.txt` published in the same release. A missing manifest, an
+   unlisted asset or a mismatching hash aborts. It then unpacks each platform
+   archive and stages only two members: the binary as
+   `bundled/<os>_<arch>/bundled-extensions[.exe]` (mode 0755) and the
+   descriptions file as `bundled/lstk-extensions.toml` (taken once; every
+   archive must carry an identical copy). It re-creates one `lstk-<name>`
```
peter-smith-phd: Is this symlink-creation still relevant? I thought you wanted to skip that approach and just set `argv[0]` explicitly.

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+   symlink to the binary per command, except on Windows (see "Running an
+   extension directly" below), and records the names, sorted, in
+   `bundled/bundle-commands.txt`: they are the bundle's own declaration of
```
peter-smith-phd: If we're not creating symlinks any more, what will `bundle-commands.txt` contain? The only way to know what the bundled extensions are is the `lstk-extensions.toml` file.

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+   which commands the binary answers to. An archive with no aliases, or whose
+   list differs from another platform's, aborts. It fails if any of lstk's six target platforms
+   (`linux`/`darwin`/`windows` × `amd64`/`arm64`) has no archive. A platform can be exempted only by listing it in
+   `UNSUPPORTED_PLATFORMS` at the top of the script, so a gap is always a
+   visible choice.
+3. **Gate the pairing.** `scripts/check-descriptions.sh bundled/linux_amd64`
+   reads the command names from the toml (left-hand side only; values are
+   never parsed) and compares them with the binary's own list in
+   `bundle-commands.txt`. It fails if commands are described but there is no
+   binary, if there is a binary but no commands are described, if the command
```
peter-smith-phd: This is only a single binary - `bundled-extensions`, and the only way we know which extensions are supported is to read the `lstk-extensions.toml`, so I'm not clear how this "check" will work.

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+   list is missing or empty, or if the toml describes a command the bundle
+   does not provide (lstk would exec the binary under a name it does not
+   answer to). A command the bundle provides but the toml omits only warns:
+   lstk will not expose it. Descriptions and the list are the same on every
+   platform, so one directory is enough.
+4. **Package.** GoReleaser adds the staged files to each archive at the root.
```
peter-smith-phd, line 109: Am I correct in believing that the output from this step is some type of `npm` or `brew` archive file that we can manually installed to check that it works? If so, I'd definitely like to do that as part of my testing, _before_ we ever publish anything to the real _npm_ or _brew_ repositories.

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+
+## What the private repository must publish
+
+Each tagged release of `localstack/lstk-bundled-extensions` ships:
+
+- one archive per lstk target platform, named
+  `bundled-extensions_<tag>_<os>_<arch>.tar.gz` (`.zip` for Windows),
+  containing at its root the multi-call binary `bundled-extensions` (`.exe` on
+  Windows), `lstk-extensions.toml`, and one `lstk-<name>` alias entry (a
+  symlink on Unix, a copy on Windows) for every command the binary answers
```
peter-smith-phd: Again, I thought the plan was to remove symlinks? Is this file out of date, perhaps?

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+  to. They serve two purposes: they are the binary's own statement of its
+  command list, which the release gate checks the toml against, and they are
+  re-created in the packaged install so a command can be run directly (see
+  below);
+- the same `lstk-extensions.toml` in every archive, hand-authored, describing
```
peter-smith-phd, line 153: For now, let's remove `deploy` from this file. I'm expecting it'll take another 2 weeks before I'm ready to ship `deploy`, and I don't want to block your work on launching `doctor`.

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+or, without access to the private repository, with placeholders:
+
+```bash
+scripts/fetch-bundled-extensions.sh --stub
+goreleaser release --snapshot --clean
+```
+
+`--stub` writes a tiny shell-script stand-in named `bundled-extensions` for
```
peter-smith-phd, line 188: What's the purpose of `--stub`? Do we have a practical reason for testing with stubs, rather than the real binaries? To me, it seems like extra code that could break, or confuse people.

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+`goreleaser check`. It fails if `.goreleaser.yaml` references `bundled/` while
+the release job has no fetch step, or the reverse. `goreleaser check` cannot
+catch this itself because it only validates config syntax and never looks at
+the filesystem. Both halves must land in the same PR.
+
+The bash suites for all four scripts run with `make test-scripts`.
+
+## Updates
+
+**Homebrew and npm** replace the whole package directory on `lstk update`, so
```
peter-smith-phd, line 285: This is good to know, and gives me more confidence that we'll get it right 😄

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+replacement of `lstk` + `bundled-extensions` + `lstk-extensions.toml` is
+pending (section 1 of the `add-bundled-extension-distribution` change). Until
+it lands, the in-the-field updater replaces only `lstk`; the other two files
+must be extracted from the archive by hand. Updating never deletes standalone
+`lstk-<name>` files a user placed next to the binary.
+
+## Running an extension directly
+
+Where the channel allows it, a release also carries one `lstk-<name>` symlink
+to `bundled-extensions` per command, so `lstk-doctor` can be run straight from
```
peter-smith-phd: I think this is a bit misleading, but running `lstk-doctor` from a shell won't give it the necessary extension context data structure it needs. If we're not using symlinks as the official mechanism, I suggest we don't create them at all.

docs/extensions-bundling.md
```diff
@@ -0,0 +1,222 @@
...
+
+`lstk update` on the binary channel does not currently re-create them: its
+extractor skips symlink entries. Whatever a fresh install put there is left
+alone, so an updated install keeps the links it already had.
+
+## Diagnosing a broken install
+
+If `bundled-extensions` is present but `lstk-extensions.toml` is missing,
+unreadable or empty, lstk cannot know which commands the binary provides.
+Running a command the bundle would normally provide reports "bundled
```
peter-smith-phd, line 300: How does `lstk` know that the command is a bundled command (e.g. `doctor`) versus just a completely missing command (e.g. `doesnt-exist`). In other words, how does it know to indicate that "extensions are not usable"?

peter-smith-phd (changes requested):
Thanks @carillan81, I've read through the relevant `.md` files, and have made a few comments. Most notably, I see that we're still talking about the `lstk-<name>` symlink in many places, but I thought we had decided against that idea?

Once we got that sorted out, I'd like to do a lot of manual testing myself (using the code in this PR). That'll convince me whether we're ready to launch this or not. I think we're heading in the right direction with this PR, but I'd just like to double-check everything.

docs/extensions-bundling.md
carillan81: This allows to mock bundled-extensions so a user without access to the private repo can create their own packages.

carillan81:
> Thanks @carillan81, I've read through the relevant `.md` files, and have made a few comments. Most notably, I see that we're still talking about the `lstk-<name>` symlink in many places, but I thought we had decided against that idea?
> 
My understanding was that we don't use symlinks to run extensions from lstk but we keep them as a cosmetic and auxilliary tool (run extensions without lstk providing context manually). Revisiting this, it's really not needed to bundle the symlinks and introduces some drift between installation methods (symlinks work on some, not others). I will then remove everything related to symlinks from the bundling mechanism.
I'll add the changes when they are ready (some hours I hope)

docs/extensions-bundling.md
carillan81: I was using symlinks to declare extensions but it was a weak check. I'm implementing a quick check in `bundled-extensions` binary to return the extensions contained in the package, This will help to check consistency with the toml file and avoid bad releases.

docs/extensions-bundling.md
carillan81: the `bundled-extensions` binary doesn't need to be in the path. i.e. npm and brew link lstk binary, but at runtime it reads the actual path and looks there for the toml and binary.
Agreed that a better message when invoked without the proper context is a good way to help the user. Implementing.

docs/extensions-bundling.md
carillan81: I didn't know either 🤷 TIL

docs/extensions-bundling.md
carillan81: In order to be able to check consistence, I'm creating a way to launch the binary and return the extensions included in the bundle

docs/extensions-bundling.md
carillan81: We remove this

docs/extensions-bundling.md
carillan81: There is a three step flow for this:
1. Look in local arguments (`lstk start`)
2. Check toml file for extensions, if found, run the `bundled-extensions` binary changing argv(0)
3. Look for lst-foo in the path

If extensions are missing the installation will flag itself as broken

docs/extensions-bundling.md
carillan81: ok. Only issue with this is that then the toml file and the binary won't match unless we remove deploy command entirely. For this my proposal is to disable the consistency check until we release deploy. There is only one extension, issues are very very unlike

docs/extensions-bundling.md
carillan81: Yes. A snapshot build produces the same artifacts the real release publishes, so you can install and test all three channels before anything goes out.

I've added a "Testing a release candidate before publishing" section to this doc with the exact commands. I ran all of them locally, so they're verified rather than just plausible.

Two things:

npm: run npm pack first. Installing the directory directly makes npm symlink it, and the launcher then can't find the platform binary — it looks like a packaging bug but isn't.
Homebrew: brew won't install a loose .rb any more, so the cask has to go into a local tap. The sha256 goreleaser generates already matches your local archives; only the URLs need repointing at file://.
One thing worth flagging: use the real bundle for this, not --stub. The stub only proves the packaging path runs and tells you nothing about the extensions — so you'll need access to extensions repo (the secret is already in the repo for CI, you can manage access diferently maybe)

carillan81:
Hey @peter-smith-phd, I pushed some more changes. Those changes are aligned with [this other PR in the `lstk-bundled-extensions` repo.](https://github.com/localstack/lstk-bundled-extensions/pull/14)
In order to be able to use this complete the other PR needs to be merged and a release generated.
I think most of the questions are addressed.

docs/extensions-authoring.md
```diff
@@ -111,6 +112,29 @@ Reuse `machineId` rather than deriving your own: it is already the final hashed
...
+`.exe`, and lstk only ever hands the bundle a name the toml lists, so a lookup
+miss inside the binary means the toml and the binary disagree: report it
+loudly rather than guessing.
+
+The binary is the only copy on disk — there is no `lstk-<name>` file per
+command — so it is reachable only as `lstk <name>`. It nonetheless sits in the
+install directory next to `lstk`, where somebody will eventually find it and
+run it directly. Handle that: when `LSTK_EXT_API_VERSION` is unset there is no
+runtime context to work from, so print a short message saying the binary is
+part of lstk and naming the command to use instead, then exit non-zero. Do not
```
peter-smith-phd, line 134: This is a good approach. I assume there'll be another PR soon, to add this feature to `lstk-bundled-extensions`?

docs/extensions-bundling.md
```diff
@@ -0,0 +1,322 @@
...
+   archive must carry an identical copy, which is what lets the next step
+   verify one file rather than six). Nothing else is staged, and nothing else
+   reaches a released package. It fails if any of lstk's six target platforms
+   (`linux`/`darwin`/`windows` × `amd64`/`arm64`) has no archive. A platform can be exempted only by listing it in
+   `UNSUPPORTED_PLATFORMS` at the top of the script, so a gap is always a
+   visible choice.
+3. **Gate the pairing.** `scripts/check-descriptions.sh bundled/linux_amd64`
+   reads the command names from the toml (left-hand side only; values are
+   never parsed) and compares them with the bundle's own answer: it runs
+   `bundled-extensions list`, which prints one bare command name per line. It
```
peter-smith-phd, line 96: At the moment, I see this output:

```
bundled-extensions is a multiplexed binary; invoke it through an lstk-<name> entrypoint.
registered extensions:
  lstk-deploy
  lstk-doctor
```

Is that the final output you're expecting to see, or will that be changed? I ask this questions because `bundled-extensions` seems to print this output for everything, not just for the `list` sub-command. It also doesn't print "one bare command name per line".

docs/extensions-bundling.md
```diff
@@ -0,0 +1,322 @@
...
+   archive and stages only two members: the binary as
+   `bundled/<os>_<arch>/bundled-extensions[.exe]` (mode 0755) and the
+   descriptions file as `bundled/lstk-extensions.toml` (taken once; every
+   archive must carry an identical copy, which is what lets the next step
+   verify one file rather than six). Nothing else is staged, and nothing else
+   reaches a released package. It fails if any of lstk's six target platforms
+   (`linux`/`darwin`/`windows` × `amd64`/`arm64`) has no archive. A platform can be exempted only by listing it in
+   `UNSUPPORTED_PLATFORMS` at the top of the script, so a gap is always a
+   visible choice.
+3. **Gate the pairing.** `scripts/check-descriptions.sh bundled/linux_amd64`
```
peter-smith-phd: I think the whole premise of this check is that we're concerned that `bundled-extensions` and `lstk-extensions.toml` can get out of sync, and that would be a bad thing.

But, I'm not sure it's bad. For example, when a new extension is being constructed, but not yet ready for release, we'd want it to be available in the `bundled-extensions` binary without being made available to all users. In that case, we'd leave it out of `lstk-extensions.toml`, but developers/beta-testers could manually add it back, to test the non-released command.

The `deploy` command is a great example of this. It's ready for early testing, but just not for general release.

docs/extensions-bundling.md
peter-smith-phd: Thanks, I'll definitely be testing this manually - if not today, by the end of this week 😃

docs/extensions-bundling.md
```diff
@@ -0,0 +1,322 @@
...
+
+`latest` re-resolves on every invocation, so a naive re-run of an
+already-published tag could ship different extension binaries under a version
+already in the wild. The fetch step guards against this: before resolving, it
+reads the existing GitHub release for the lstk tag being built and, if its
+notes already carry a `Bundled extensions:` line, pins the fetch to that tag.
+To do the same by hand, pass the recorded tag explicitly:
+
+```bash
+LSTK_EXTENSIONS_READ_TOKEN=... scripts/fetch-bundled-extensions.sh --tag v0.3.1
```
peter-smith-phd: Not sure if it's relevant or not, but `v0.3.1` isn't a valid tag for `bundled-extensions`. It uses `vYYYY.MM.DD` format.

scripts/test-scripts.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# Runs the bash test suites for the release helper scripts under scripts/.
```
peter-smith-phd: Can I suggest that you move all these new `.sh` files into a sub-directory called `scripts/bundled-extensions/...`?

That'll keep nicely packaged, with their intention being clear to anybody browsing the source code.

test/integration/__snapshots__/extension_bundle_test.snap
```diff
@@ -0,0 +1,52 @@
...
+
+Tools:
+  aws         Run AWS CLI commands against LocalStack
+  az          Run Azure CLI commands against LocalStack
+  cdk         Run AWS CDK against LocalStack
+  sam         Run the AWS SAM CLI against LocalStack
+  terraform   Run Terraform against LocalStack
+
+Extensions:
+  deploy      Deploy to LocalStack
```
peter-smith-phd, line 6: I worry that this snapshot test will fail on a regular basis, not because of the extensions code, but because of all the other things (commands / options) that will be changing, outside of the scope of extensions.

I'm not sure if it's worth addressing this concern now, but we should consider it in the future.

CLAUDE.md
```diff
@@ -170,7 +170,7 @@ Shared plumbing lives in `cmd/proxy.go` (`leadingFlags`, `stripLeadingProxyFlags
...
... entries every bundle archive must carry, and which is never packaged), so a described command the binary cannot dispatch fails the release; the fetch script also re-creates those aliases as relative symlinks beside the staged binary (not on Windows, where extractors materialize them as junk text files) so a packaged install can run `lstk-<name>` directly — goreleaser preserves symlinks in tar.gz, `npm pack` drops them, and lstk itself never resolves them, `.goreleaser.yaml` packages both at the archive root (the cask inherits them; a release-job step copies them into each npm **platform** package), and the resolved bundle tag is recorded in the release notes. `scripts/check-bundled-packaging-sync.sh` runs on every PR to keep the packaging and download halves from merging separately. Bash tests for these scripts: `make test-scripts` (`scripts/tests/`). Set-wise co-update on the binary channel (`internal/update`) is still pending in the `add-bundled-extension-distribution` change. See [extensions-authoring.md](docs/extensions-authoring.md) for the author-facing contract and [extensions-bundling.md](docs/extensions-bundling.md) for the release pipeline and on-disk layout per channel.
```
peter-smith-phd: Let's try and trim down this contribution to `CLAUDE.md`. Given that agents will unconditionally read this file, we want it to contain the most important information, not everything.  Perhaps 5-10 lines maximum?

We already have lots of detail in the `openspec` files, which are _conditionally_ read when needed, rather than read every time.

peter-smith-phd (commented):
Thanks @carillan81, it's looking really good. I've provided more feedback, although it's pretty minor this time.

The only thing I want to do is manually test this on MacOS, Windows, and Linux, for the three different distribution methods.  I got part way through doing this, but ran out of time today. I should be able to get it tested tomorrow.

One more thing - given that we're publicly launching `lstk` on Tuesday, there'll be a lot of people downloading `lstk` next week. I suggest we hold off merging this until at least the following week.

docs/extensions-authoring.md
carillan81: PR is already created: localstack/lstk-bundled-extensions#14

docs/extensions-bundling.md
carillan81: This needs a PR: localstack/lstk-bundled-extensions#14

test/integration/__snapshots__/extension_bundle_test.snap
carillan81: I changed it to just check extensions part

docs/extensions-bundling.md
carillan81: fixed

CLAUDE.md
carillan81: trimmed

peter-smith-phd (approved):
Thanks, I've managed to successfully test on MacOS, Linux, and Windows, using binaries, npm and brew. Everything worked as expected for a fresh install.

I'm now confident this will work, although let's still delay the merge until the end of next week, given that we have the `lstk` marketing launch coming up on Tuesday. We don't want to accidentally break things shortly after the launch.

Thanks for doing this!

docs/extensions-bundling.md
carillan81: I had that in mind:

* If an extension is in the toml file but not in the binary - Error, stop the release
* If an extension is in the binary but not in the toml file - Warning but continue

This is the current situation with deploy. My last PR in `lstk-bundled-extensions` ships deploy in the binary but not in the toml file so it won't be shown in `lstk --help`. It can be run "if you know what you are doing".

PR #478, opened 2026-09-01
Add --json output to lstk start

anisaoshafi:
Added `--json` support to `lstk start` (and bare `lstk`, since it starts the same way).

`--json` always forces non-interactive behavior, even from a real terminal. The TUI never launches when `--json` is passed.

What it returns:
- Success: emulator type, container name, endpoint, version, whether it was already running, persistence.
- Failure: a stable error code (e.g. `LICENSE_INVALID`, `PORT_CONFLICT`, `AUTH_REQUIRED`) instead of just text.

Decisions:
- One JSON object on stdout, nothing else, same as stop/reset/update. No separate progress channel.
- Added two error codes that were missing: `PORT_CONFLICT` and `LICENSE_NOT_COVERED`.
- `container.Start` used to return just a version string, and an empty string meant "already running", an indirect signal that could also just mean the version wasn't known. Now it returns a small struct with a real `AlreadyRunning` field plus the connection details (endpoint, container name, persistence), so callers get that data directly instead of inferring it from a blank string.

| scenario | output |
|----| -----|
| no running emulator | imagen |
| already running emulator | imagen |
| bare `lstk --json` | imagen |
| port conflict error | imagen |
| startup timeout | https-_github com_user-attachments_assets_c28876c1-354b-40e2-8747-6f4b140db1fc |
| bad token | imagen |

<details>
<summary> <h2>JSON response structure</h2></summary>

```json
{
  "schemaVersion": 1,
  "command": "start",
  "status": "ok" or "error",
  "data": { ... },
  "warnings": [ ... ],
  "error": { ... }
}
```

| Field | Meaning |
|---|---|
| `schemaVersion` | A version number for this response format. Only changes if the structure itself changes later. |
| `command` | Which command this came from. Always `"start"` here. |
| `status` | `"ok"` or `"error"`, the first thing to check |
| `data` | Filled in when it works |
| `warnings` | Heads-up notices, not failures |
| `error` | Filled in when it fails |

**When it works, `data` contains:**

| Field | Meaning |
|---|---|
| `emulator` | Which emulator started: `aws`, `snowflake`, or `azure` |
| `container` | The name it's running under |
| `endpoint` | The address to connect to it |
| `version` | The version that's running |
| `alreadyRunning` | `true` if it was already up, `false` if this call started it |
| `persistence` | `true` if it keeps data between restarts |

**When it fails, `error` contains:**

| Field | Meaning |
|---|---|
| `code` | A short, fixed label for the reason (e.g. `PORT_CONFLICT`) |
| `category` | A general group for that reason (see below) |
| `message` | A plain-language explanation |
| `retryable` | `true` if trying again later might just work |
| `actions` | A suggested fix, when there's an obvious one |

**Categories (`category`):**
- `RUNTIME`: Docker or port problem
- `AUTH`: login or license problem
- `EMULATOR`: it failed to start, or something else is already using it
- `CONFIG`: a bad lstk setting
- `INTERNAL`: anything else

</details>

Resolves DEVX-1052


peter-smith-phd (commented):
A few questions/comments:

* For the "port conflict" example where you show the screenshot, why does it return `status: ok` rather than `status: error`?  Any error of this nature should be made visible to a tool (e.g. VS Code) that invokes `lstk`
* What error code is returned if the `--timeout` limit is exceeded? I believe this feature was added recently, and might not have been considered when the error codes were proposed.

Other than those things, it looks good to me.

anisaoshafi:
>  For the "port conflict" example where you show the screenshot, why does it return `status: ok` rather than `status: error`?  Any error of this nature should be made visible to a tool (e.g. VS Code) that invokes `lstk`

My bad @peter-smith-phd, I pasted the wrong screenshot in the PR description, just updated it there. This is the error:
imagen

anisaoshafi:
> What error code is returned if the --timeout limit is exceeded? I believe this feature was added recently, and might not have been considered when the error codes were proposed.

It was already implemented, but I didn't include in description. I've added this scenario in PR description with corresponding screenshot.
This is the payload, error code `EMULATOR_START_FAILED`: 
Screenshot 2026-09-02 at 14 26 58

peter-smith-phd (approved):

PR #479, opened 2026-09-01
Add --json output to lstk status

anisaoshafi:
Add `--json` output to `lstk status`.
`--json` with `--endpoint-url` is also supported.

What it returns:

- Per configured emulator: type, running, health, and (when running) name, version, host, uptime, persistence.
- If the emulator is not running, returns an error (`EMULATOR_NOT_RUNNING`), same as `stop`/`reset`. Not a plain-text error, a proper envelope.

Decisions:
- `no-resource` flag added to avoid from listing resources. listing resources by default can be slow, so this is a way to opt out.
- JSON mode stops at the first not-running configured emulator, same as plain-text status (and same as stop/reset). It does not list every configured emulator.
- An external target (--endpoint-url) always has `running: true`, since resolving the target already confirms it is reachable before status runs. `name`, `uptimeSeconds`, and `persistence` are left out for it, those are Docker facts that do not exist for an emulator lstk did not start.

| scenario | output |
|----|-----|
| not running |  imagen |
| running and healthy | imagen |
| no resources flag | imagen | 
| endpoint URL | imagen |
| not reached | imagen |
| config not found | imagen | 
| docker daemon stopped | imagen |

<details>
<summary> <h2>JSON response structure</h2></summary>

```json
{
  "schemaVersion": 1,
  "command": "status",
  "status": "ok" or "error",
  "data": { ... },
  "warnings": [ ... ],
  "error": { ... }
}
```

| Field | Meaning |
|---|---|
| `schemaVersion` | A version number for this response format. Only changes if the structure itself changes later. |
| `command` | Which command this came from. Always `"status"` here. |
| `status` | `"ok"` or `"error"`, the first thing to check |
| `data` | Filled in when it works |
| `warnings` | Heads up notices, not failures |
| `error` | Filled in when it fails |

**When it works, `data` contains a list of emulators, one entry per configured type that was reached:**

| Field | Meaning |
|---|---|
| `type` | Which emulator this is: `aws`, `snowflake`, or `azure` |
| `running` | Always `true` for an entry that made it into the response. A not-running emulator is an error instead, see below |
| `health` | `"healthy"` or `"unhealthy"` |
| `name` | The container name. Only present for an lstk-managed emulator |
| `version` | The version that's running |
| `host` | The address to connect to it |
| `uptimeSeconds` | How long it has been up. Only present for an lstk-managed emulator |
| `persistence` | `true` if it keeps data between restarts. Only present for an lstk-managed emulator |
| `resourceSummary` | Counts of what's deployed inside. Only present with `--resources` |
| `resources` | The full list of what's deployed inside. Only present with `--resources` |

**When it fails, `error` contains:**

| Field | Meaning |
|---|---|
| `code` | A short, fixed label for the reason (e.g. `EMULATOR_NOT_RUNNING`) |
| `category` | A general group for that reason (see below) |
| `message` | A plain language explanation |
| `retryable` | `true` if trying again later might just work |

**Categories (`category`) seen here:**
- `RUNTIME`: Docker isn't available. Retryable
- `EMULATOR`: the emulator isn't running. Not retryable
- `CONFIG`: a bad lstk setting. Not retryable

</details>

Resolves DEVX-1053

### DIFF
--- a/cmd/json_envelope.go
+++ b/cmd/json_envelope.go
@@ -40,7 +40,8 @@ func jsonAwareSink(cmd *cobra.Command, cfg *env.Env, w io.Writer) output.Sink {
 }
 
 // writeEnvelope marshals envelope as compact JSON and writes it to w, followed
-// by a newline, as the single line of output a JSON-capable command produces.
+// by a newline. SetEscapeHTML(false): this is CLI output, not HTML, so & < >
+// should render literally.
 func writeEnvelope(w io.Writer, envelope output.Envelope) error {
 	enc := json.NewEncoder(w)
 	enc.SetEscapeHTML(false)
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -19,21 +19,44 @@ import (
 )
 
 func newStatusCmd(cfg *env.Env) *cobra.Command {
-	return &cobra.Command{
-		Use:     "status",
-		Short:   "Show emulator status and deployed resources",
-		Long:    "Show the status of a running emulator and its deployed resources",
-		PreRunE: initConfigDeferCreate(nil),
+	cmd := &cobra.Command{
+		Use:         "status",
+		Short:       "Show emulator status and deployed resources",
+		Long:        "Show the status of a running emulator and its deployed resources",
+		PreRunE:     initConfigDeferCreate(nil),
+		Annotations: map[string]string{jsonSupportedAnnotation: "true"},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			target, err := endpoint.Resolve(cmd.Context(), cmd)
-			if err != nil {
-				return err
-			}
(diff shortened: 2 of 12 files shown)


peter-smith-phd (commented):
Thanks for doing this. It generally looks good to me, although I feel we shouldn't mark `--endpoint-url` as being out of scope. Although we don't immediately need it to support VS Code, the `lstk status` command does support this option, we'd only be creating work for ourselves to come back and fix it later (it'll definitely be needed).

anisaoshafi:
Thanks @peter-smith-phd, agree. 
Added support for `--endpoint-url` https://github.com/localstack/lstk/pull/479/commits/109254f7a1dcce8ec5af3472e9ec03d34cfbf4ce. Tested with an ephemeral instance: imagen

peter-smith-phd (approved):

PR #482, opened 2026-09-03
Dpx 693 pr 2 teach lstk update to handle the bundled extensions

carillan81:
## Motivation

Once releases ship the bundled extensions (`bundled-extensions` and `lstk-extensions.toml` next to `lstk`), `lstk update` on the binary channel has to replace all three files together, not just the binary. Today it replaces one file, `lstk`, and ignores everything else in the archive.

That has a second-order problem. The first bundling release is installed by the updater users already have, which only replaces `lstk`. They land on a current binary with no extensions, and because they are already on the newest version, `lstk update` says "already up to date" until the next release ships. Nobody can fix that retroactively, so the new updater has to detect and repair it.

## Solution

**Set-wise replacement (stage-then-commit).** Every member of the archive is copied into the install directory under a `.lstk-new` name; only when all copies succeed is each renamed over its final name, `lstk` last. A file visible under its real name is never half-written, a failure before the last rename leaves a working `lstk` to re-run with, and the updater never deletes an `lstk-*` file it cannot prove it owns (additive-only). An archive containing only `lstk` behaves exactly as before, so pre-bundling releases and rollbacks are unaffected.

~**Repair when current but incomplete.** The release stamps the expected set into the binary (`version.bundledSet`, via ldflags). When versions match but a stamped member is missing or unusable, `lstk update` reinstalls the same release, then re-checks; if the archive still did not deliver the members it fails loudly ("did not restore the bundled extensions") instead of looping. Empty stamp keeps today's pure version comparison. Homebrew and npm never repair; the package manager owns the whole set there. The passive update notice also nudges an incomplete install toward `lstk update`.~

The same-version repair from the first revision was dropped in review. The transition release (old updater, new archive) leaves the bundle uninstalled until the next `lstk update`. Instead of repairing, lstk tells the user: a release build that finds no bundle beside itself adds a reinstall hint to the unknown-command error and warns on an up-to-date `lstk update` (also in the `--json` warnings array). One goreleaser ldflag, one detector, two call sites.

**Hardening.** Windows moves every existing member aside before renaming (a running extension no longer breaks the commit); a squatting symlink or directory at a staging path is refused rather than written through; the install directory is listed literally instead of globbed (a `[` in the path used to fail every update); setuid/setgid bits survive the update; a concurrent update cannot truncate another's staging file.

**Structured output.** `UpdateCheckedEvent` gains `RepairBundled`; the `--json` check shape gains `"repairBundled": true` for a same-version repair (absent otherwise, and scrubbed from the applied shape).

**Dependencies and follow-ups**
- `dpx-692` (PR 1) must merge first or together: it carries the resolver that dispatches bundled commands from the toml. Without it a repaired install has the files but nothing resolves them.
- The `bundledSet` ldflags stamp (tasks.md 5.2) is not yet in `.goreleaser.yaml`; until it lands on the packaging side the repair path stays dormant and this PR changes nothing user-visible.
- Once both branches are in, alias `bundledBinaryBaseName` from `extension.BundledBinaryName` (TODO left on the constant).

**Testing**
- Unit: `internal/update` covers every behaviour above, including the Windows path from any host via a `goos` parameter. All new behaviour was written test-first.
- Integration: transition repair, complete-set no-op, pre-bundling release unaffected, broken-archive fail-loud, leftover cleanup, plus the existing mock-GitHub happy/failure paths.
- Manual testbed against packages built by the dpx-692 pipeline (`goreleaser`, npm publisher): the full transition (published 0.23.0 to bundling release, repair, no-op, bundled-to-bundled upgrade, rollback, forward) passes on the binary channel and npm. Caveat: the currently published bundle predates the `bundled-extensions list` contract, so the release gate was run as a warning for that exercise; Homebrew was not executed.

<details>
<summary>Docs</summary>

~No documentation work needed outside this PR. `docs/structured-output.md` is updated for the new `repairBundled` key, and the update guarantees are documented in the `internal/update` package comment. User-facing bundling docs (`docs/extensions-bundling.md`) live on dpx-692.~

The reinstall hint is new user-facing output; the spec has a scenario for it. Release notes for the first bundling release should carry one line: if a bundled command reports "unknown command" after updating, reinstall lstk once.
</details>

## Review

Human review advisable. This is the update path, the one thing a broken release cannot ship a fix for, and it adds new user-visible behaviour (the repair). Not a self-merge candidate.

Closes DPX-693


peter-smith-phd (changes requested):
Thanks for implementing this, but I'm not sure we need a 2000-line solution to handle the "binary" case where we're moving from just `lstk` to `lstk`/`bundled-extensions`/`lstk-extensions.toml`. I appreciate the need to carefully replace the three files, rather than just one, but this solution feels over complicated. I was expecting closer to 100 lines of code.

In particular, do we really need the "repair" functionality? If it breaks for some reason, we already have error messages that inform the user of the problem. They should instead just uninstall/reinstall `lstk` and the problem will go away. My concern is that the code has become so complicated that we'll have trouble maintaining it in future.

What can we do to make this significantly less complicated, and focus on just replacing three files instead of one?

carillan81:
> Thanks for implementing this, but I'm not sure we need a 2000-line solution to handle the "binary" case where we're moving from just `lstk` to `lstk`/`bundled-extensions`/`lstk-extensions.toml`. I appreciate the need to carefully replace the three files, rather than just one, but this solution feels over complicated. I was expecting closer to 100 lines of code.
>
> In particular, do we really need the "repair" functionality? If it breaks for some reason, we already have error messages that inform the user of the problem. They should instead just uninstall/reinstall `lstk` and the problem will go away. My concern is that the code has become so complicated that we'll have trouble maintaining it in future.
>
> What can we do to make this significantly less complicated, and focus on just replacing three files instead of one?

After some discussion with humans and AI I agree that the repair functionality is an overkill. We are facing a temporal issue while the users update and the repair adds some code that will be stale in a month.

I think we must address user UX so I just implemented a better error message that gives a hint for reinstall (and command) in case the command is not found. It adds some code but barely adds a few tenths of lines.


| Bucket | Code | Comments | Blank | Docs / config | Added | Removed |
| -- | -- | -- | -- | -- | -- | -- |
| Go production | 247 | 82 | 24 |  | 353 | 29 |
| Go tests | 679 | 46 | 62 |  | 787 | 51 |
| Docs (openspec spec + tasks, CLAUDE.md) |  |  |  | 16 | 16 | 16 |
| Build config (`.goreleaser.yaml`) |  |  |  | 1 | 1 | 1 |
| **Total** | **926** | **128** | **86** | **17** | **1157** | **97** |

This looks a much better solution to me.

```
$ lstk deploy
Error: unknown command "deploy" for lstk
  This lstk release ships bundled extensions, but none are installed in <install dir>.
  ==> See help: lstk -h
  ==> Reinstall lstk: download the latest release from https://github.com/localstack/lstk/releases/latest
(exit 1)

$ lstk update
Checking for updates...
> Note: Already up to date (9.9.9-SNAPSHOT-dfd62a7)
> Warning: This lstk release ships bundled extensions, but none are installed in <install dir>. Reinstall lstk: download ...
```

Does this works for you?

peter-smith-phd:
> Does this works for you?

Sorry, I ran out of time to review this today, but will look at it in more detail tomorrow. On the surface, it looks like a good improvement 😄

openspec/changes/add-bundled-extension-distribution/specs/extension-bundling-distribution/spec.md
```diff
@@ -63,7 +63,14 @@ Bundled extensions are payload rather than a precondition in one sense only: **a
 - **WHEN** a user on a pre-bundling lstk runs `lstk update` and the latest release bundles extensions
 - **THEN** the update succeeds using the in-the-field updater (which replaces only the lstk binary and ignores the archive's extra members)
 - **AND** the install is left with an incomplete set, since that updater predates bundling and cannot be made to fail
-- **AND** the incomplete set is repaired by the next `lstk update`, which SHALL NOT wait for a newer release to become available
+- **AND** until the next release's `lstk update` (now the set-wise updater) installs the complete set, lstk points the user at a reinstall: an unknown command and an up-to-date `lstk update` both say the release ships bundled extensions that are not installed, and name the reinstall command for the install method (accepted one-release gap; a same-version repair was considered and dropped as too much machinery in PR #482)
```
peter-smith-phd, line 66: This doesn't feel correct. Will the following behaviour happen?

1) I update next week (using the binary install method) which doesn't (yet) download the `bundled-extensions` binary too.
2) I type `lstk strt` (note the typo)
3) It suggests "you should reinstall lstk"?

Am I understanding that correctly? I would have thought that `lstk` would just say `Error: unknown command "strt" for lstk` as it does now, since reinstalling seems pretty drastic to recover from a typo. The issue of the missing extensions will automatically fix itself in the next update.

In a month or so we can change this behaviour, but it doesn't feel correct in the migration period.

internal/version/version.go
```diff
@@ -4,4 +4,11 @@ package version
 // because the linker can only modify variables at link time.
 var version = "dev"
 
+// bundlesExtensions is "true" on release builds, which ship the
+// bundled-extensions binary and lstk-extensions.toml beside lstk.
+var bundlesExtensions = "false"
```
peter-smith-phd, line 9: Won't this always be `true` from now on? Is there a reason that the `BundlesExtensions` function must be called? Why not just assume that all future releases have bundles installed?

internal/update/reinstall.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	return true
+}
+
+// ReinstallCommand returns the command, or the download page for a plain
+// binary, that installs the complete set for the given install method.
+func ReinstallCommand(m InstallMethod) string {
+	switch m {
+	case InstallHomebrew:
+		return "brew reinstall --cask localstack/tap/lstk"
```
peter-smith-phd: I had the impression that `brew` and `npm` would handle the "set" correctly, and we wouldn't need to detect missing files.

.goreleaser.yaml
```diff
@@ -28,7 +28,7 @@ builds:
     flags:
       - -trimpath
     ldflags:
-      - -s -w -X github.com/localstack/lstk/internal/version.version={{ .Version }}
+      - -s -w -X github.com/localstack/lstk/internal/version.version={{ .Version }} -X github.com/localstack/lstk/internal/version.bundlesExtensions=true
```
peter-smith-phd: As mentioned earlier, I'm unsure why this is necessary. I don't think we plan to stop having bundles, so couldn't the code simply hard-code the fact that all `lstk` releases use bundled extensions.

internal/update/update.go
```diff
@@ -1,3 +1,27 @@
...
+// stage-then-commit scheme (extract.go). The guarantees to preserve:
+//
+//  1. A file under its real name is never truncated or half-written: content
+//     is only ever written to a fresh staging file and renamed into place.
+//  2. An interrupted update is repaired by re-running `lstk update`: nothing
+//     commits until everything is staged, leftovers are cleaned first, and
+//     lstk commits last, so a working lstk always remains. Windows caveat: a
+//     crash between renaming lstk.exe aside and renaming the new one in leaves
+//     no lstk.exe; rename lstk.exe.old back by hand.
+//  3. Nothing is deleted: an lstk-* file absent from the archive is left alone,
```
peter-smith-phd: This third bullet point doesn't seem relevant anymore, given that we don't use `lstk-*` files anymore.

internal/update/update.go
```diff
@@ -1,3 +1,27 @@
...
+//     because the updater cannot tell a dropped extension from a user's file.
+//
+// An archive carrying only lstk installs exactly as before bundling existed.
+//
+// The pre-bundling updater installs a bundling release with only its lstk
```
peter-smith-phd, line 21: Now that I think about it - this "update" process is what should really be failing if the "bundle" isn't downloaded correctly. The whole idea of checking (`DetectMissingBundle`) when the user later runs a command seems a bit late. It's very unlikely that the bundle files will suddenly go missing after they're successfully installed.

internal/update/extract.go
```diff
@@ -10,9 +10,82 @@ import (
...
+					err, rerr, movedAside, m.dest)
+			}
+		}
+		return err
+	}
+	return nil
+}
+
+// extractAndReplace installs every member the archive carries as one unit:
+// lstk, the bundled-extensions binary, any lstk-* binaries, and the
```
peter-smith-phd: Comment is out of date. We no longer have `lstk-*` binaries.

internal/update/extract.go
```diff
@@ -10,9 +10,82 @@ import (
...
+// descriptions file. An archive carrying only lstk is a set of size one. A
+// member that fails to stage or commit fails the whole update, naming it.
 func extractAndReplace(archivePath, exePath, format string) error {
+	return replaceSet(archivePath, exePath, format, goruntime.GOOS)
+}
+
+// replaceSet takes the platform as a parameter so the Windows naming and
+// move-aside rules are testable on any host (unit tests run on Linux in CI).
+func replaceSet(archivePath, exePath, format, goos string) error {
```
peter-smith-phd, line 88: While I was reviewing this, I started to wonder why we need to do this file-by-file. Since the binary archives are literally `tar.gz` or `.zip` files, why couldn't we just extract the full archive and replace the entire directory in a single `mv` command. Unfortunately, I think the answer is that we don't have a well-defined way for binary-downloads to be installed, and we can't possibly know where the user decided to install `lstk` in the first place. This is messy, but that's what we have, so I guess the file-by-file replacement makes sense and partly explains why this PR is longer than I thought it should be 🫨 

But, it did uncover the fact that `completions` and `manpages` directories are never updated. They'll always contain the original content they were downloaded with (at least, that's what Claude told me).

internal/version/version.go
carillan81: this is false until we merge the other PR, just in case we merge this one first (this one could already be merged and released with lstk usual cycle)

internal/update/update.go
carillan81: cleaned up comments

internal/update/extract.go
carillan81: That's correct. The lack of `completions` and `manpages` in the update is a previous issue, not related with extensions. My take is that we create a new issue and fix it in a different PR.

openspec/changes/add-bundled-extension-distribution/specs/extension-bundling-distribution/spec.md
carillan81: changed so we show the reinstall message if the user tries to run `doctor` or `deploy`. This way we keep the good UX of extensions without polluting other regular errors

internal/update/reinstall.go
carillan81: That is correct, we removed the check and error message for those cases

.goreleaser.yaml
carillan81: It's cheap to keep it and allows to properly detect failed updates. This is the way to tell lstk "this bundle should (shouldn't) have extensions".
Changed to false. It needs to be set to true the moment we release an extension.

carillan81:
Refined some workflows after review:

* From a product perspective, the reinstall warning seems to be the best option. I properly narrowed down the hint. Reinstall message only appears if you try to run `doctor` or `deploy` commands, any other typo or unknown command would yield just "unknown command" error.
  I think we should keep the flag, it's a cheap mechanism and enables the reinstall message. We can remove it after the transition if it's really in the way.
* Reinstall message only for binary, there is no way we can have an incomplete install from npm or brew.
* Removed some remnants from lstk-\* symlinks that remained in the code.
* Manpages and completions are not updated in the binary path, but is not something in the scope of this change, we can address it in a different issue.

I think this narrows down the changes for UX and the logic is not complex. We can trim down some of this things when the extensions are stable enough.

internal/update/update.go
carillan81: The issue with missing extensions is not when you use this new `update` method, is because you used the old one that just replaced `lstk`. There is no way to detect that unless you run `lstk` again.

.goreleaser.yaml
peter-smith-phd: OK thanks. I'd really love to see it removed eventually though. This is similar to the classic problem of cleaning up "Feature Flags" after a feature has already been launched to everybody 😄

internal/update/update.go
peter-smith-phd: > There is no way to detect that unless you run lstk again.

I think that's the behaviour I'm expecting to see - you _would_ run it again.

Week 1 - First `lstk` update doesn't support extensions, but does support the ability to download them in future. This doesn't really matter, since customers aren't expecting to see `doctor` or `deploy`, so they won't notice any problems.

Week 2 - Second `lstk` update correctly downloads and allows extensions to be run.

The "added check" that I'm suggesting is for `lstk` to detect/report missing extensions, _only_ after the first update has already taken place. This is when any failed download or failed install/update should be noticed. It also seems appropriate to say: "Sorry, the download failed and your installation is corrupted, please reinstall".

Having said all of this, I think your current solution is probably OK, and no additional code changes are required. Because:

1) You do already have lots of error checking in the install/update code, so we're covered there.
2) You've limited the error reporting to only be for `lstk doctor` and `lstk deploy`, not just any `lstk xxx` command that's missing (although, I still feel this check is unnecessary, it won't do any damage).
3) This only impacts users who have downloaded the binaries directly, and I suspect the majority of customers will have used `brew` or `npm` instead. We might only be worrying about a very small number of users.

internal/update/extract.go
peter-smith-phd: Agreed - not your problem to fix.

This is just a fundamental flaw with the way that "binary download" has been designed. It doesn't account for updating any files except for `lstk`, which means that `manpages` and `completions` get out of date over time.

internal/update/reinstall.go
```diff
@@ -0,0 +1,74 @@
...
+	return detectMissingBundle(info, goruntime.GOOS)
+}
+
+// DetectMissingBundleFor is DetectMissingBundle limited to the commands the
+// bundle provided at the time of the transition, so a typo never earns a
+// reinstall hint. Retire it with the hint once every supported release ships
+// the set-wise updater.
+func DetectMissingBundleFor(command string) (MissingBundle, bool) {
+	switch command {
+	case "deploy", "doctor":
```
peter-smith-phd, line 48: I won't hold up this PR for this, but I still feel this check isn't necessary.

That is, if the bundled extensions aren't installed yet, because the user hasn't updated to the latest release, then those commands simply don't exist yet. They'll get them the following week when they do `lstk update` again. No big deal - it just delays the roll-out by an extra week, but only for users who downloaded the binaries directly (very few people).

On the other hand, if the bundle goes missing in > 2 weeks from now, it can only be because of a corrupt download/install, which should be detected and reported during the `lstk update` process.

Am I missing a case, perhaps? I agree that `lstk update` will need to run twice before the extensions appear, but that's a reasonable trade-off.

In any case, we should plan to remove this code at a later point in time. It'll just clutter up the logic if it stays there longer than a few months 😄

peter-smith-phd (approved):
Thanks, looks good to go 😄 . I added a few comments about the "Missing Bundle" mechanism, but nothing that would block this from being released. However, I do feel we should clean up the code in a few months from now, after bundled extensions are the default behaviour.

As for when to merge, I'd suggest this Friday? There are scheduled releases every Thursday, so this new code will go out on Sept 17th - far enough away from the `lstk 1.0.0` launch that we won't be changing too much in a critical launch week.

joe4dev:
@carillan81 How does the bundled extension update integrate with externally-managed lstk installations such as mise or nix?

Background: See DEVX-1029 and PR https://github.com/localstack/lstk/pull/491

For example, the proposed solution says:

> Every member of the archive is copied into the install directory under a .lstk-new name

How does this work with nix where the install directory is read-only? Have we considered using the writable lstk directory `~/.config/lstk/` to store extensions?

> And with nix, lstk lives in a read-only location, and couldn't upgrade itself anyway.

carillan81:
This file by file update is done only when `lstk` was installed with the binary distribution. Every other current available install method uses the tool command (i.e `brew upgrade lstk`).

I guess if we have any other install mechanism it won't change as long as we use the external tool to update.

joe4dev:
> This file by file update is done only when `lstk` was installed with the binary distribution. Every other current available install method uses the tool command (i.e `brew upgrade lstk`).

Thank you for the clarification.

> I guess if we have any other install mechanism it won't change as long as we use the external tool to update.

So, homebrew, mise, and nix will take care of updating the lstk extensions automatically?
Are extensions always shipped together with lstk / tied to one lstk version (i.e., no version management required) or can they be updated separately?
I found the [extension-bundling-spec](https://github.com/localstack/lstk/pull/482/changes#diff-b533f001fe93f348b44a2bd4a9696b2a32c30738c98aa85ce24aa5cee18a9679L1) quite helpful to learn more about the purpose (e.g., shipping closed-source extension)

carillan81:
Extensions are only updated when `lstk` is updated. They are a package and there is no way to update extensions right now if they are not updated with `lstk`. I don't know about mise and nix but I have a "homemade testbed™️" and properly tested npm and hombrew. They take care of replacing all the bundle.

internal/update/reinstall.go
carillan81: I will keep this for the moment as product wanted a good UX and feedback to the user. I agree we can remove this in a few months, when the adoption of 1.x.x versions is high enough.

PR #490, opened 2026-09-07
Add --containers, --arch and copyto to the EC2 sandbox script

peter-smith-phd:
## Motivation

`scripts/ec2-sandbox.sh` starts up a throwaway Windows or Linux system on EC2 for testing `lstk` or LocalStack Toolkit. This PR adds some important features: run a real Docker Engine on Windows, build on ARM64 CPUs, and get a file onto the box without hand-rolling an `scp` command.

## Solution

Four additions to the sandbox script. 

- **`--containers=docker|podman`** — Windows sandboxes now default to real Docker Engine, accessible via `\\.\pipe\docker_engine`, the same endpoint Docker Desktop uses, so `docker.exe` and lstk both find it with no configuration. `--containers=podman` keeps the previous behaviour of using Podman.

- **`--arch=x64|arm64`**, required for `create` — Linux sandboxes can now be either ARM64 or x64 architecture. Windows is x64-only because ARM64 is not supported by Microsoft on their Windows server images (only on Windows Desktop).

- **`copyto <os> [--arch <a>] <path>`** — copies a file, or a directory and its contents, onto the sandbox's desktop (`User`'s on Windows, `user`'s on Linux). Important for manually installing `.vsix` files or other necessary test files.

- Every sandbox now also gets the AWS CLI, Terraform and the SAM CLI installed system-wide.


anisaoshafi (approved):
Thanks for improving the setup 👏🏼 
I'll be testing it this afternoon.

PR #491, opened 2026-09-08
Allow permanently disabling lstk update checks

joe4dev:
## Motivation

lstk asks to update on every start with no way to turn it off. With weekly releases, "skip this version" buys a few days — the reporter sees the prompt almost daily.

It is also wrong when another tool owns the binary. mise, nix, asdf, scoop and chocolatey manage the version themselves; replacing it desynchronises them, and on nix the store is read-only so the update fails after a full download.

## Solution

`[cli] check_for_update_on_startup`, also `LSTK_CHECK_FOR_UPDATE_ON_STARTUP`:

| setting | self-managed, interactive | externally managed or non-interactive |
| --- | --- | --- |
| `true` (default) | today's blocking prompt | one non-blocking note |
| `false` | no check | no check |

A boolean rather than a `prompt`/`notify`/`off` enum ([review](https://github.com/localstack/lstk/pull/491#discussion_r3989880826)): install detection already decides between a prompt and a note, so the only choice left to the user is whether to check at all. Externally-managed installs get the note and name the manager rather than advising `lstk update`. `lstk update` refuses on them (`--force` overrides); an explicit `lstk update` and `lstk update --check` are never gated by the setting.

### Prompt: before / after

| Before | After |
| --- | --- |
| lstk-upadate-prompt-v1 | lstk-update-prompt-v2 |
| `[S] Skip this version` — per-version, buys days | `[N] Never check again` — persists `check_for_update_on_startup = false` |

`[S]` is removed rather than kept alongside `[N]`: it is the option the ticket says does not solve the problem, it was the third way to say "no" on one prompt, and it was the only per-version persisted state. `cli.update_skipped_version` goes with it; a leftover key is inert.

## Docs

<details>
<summary>Docs needed</summary>

New user-facing surface to document:

- `[cli] check_for_update_on_startup` config key and the `LSTK_CHECK_FOR_UPDATE_ON_STARTUP` environment variable (boolean, default true), and that neither gates an explicit `lstk update`.
- Externally-managed installs (mise, nix, guix, asdf, scoop, chocolatey) default to `notify` and are not updated in place; `lstk update --force` overrides.
- `--json` gains the `UPDATE_EXTERNALLY_MANAGED` error code (`docs/structured-output.md` updated in this PR).
- The update prompt loses `[S] Skip this version` and gains `[N] Never check again`.

</details>

<details>
<summary>Manual testing (using fish shell) — 13 scenarios, all passing</summary>

manual-testing

> [!NOTE]
> Captured before the boolean simplification, so the screenshot shows the old `update_check` key. Behaviour is unchanged apart from the key name and `[N]`'s persisted value; the recipe below is current.

Covered: enabled and disabled; env var overriding config; an invalid value rejected rather than coerced; a mise install naming the manager in its note; `lstk update` refusing on mise and nix (plain and `--json`); a read-only install directory refused; `update --check` still working when the check is disabled; `[N]` persisting the opt-out and the next run staying silent; the first-run prompt omitting `[N]`; and a mise install never prompting.

**To replicate.** Build a version-stamped binary (a `dev` build skips the check entirely) and point Docker at nothing, so `start` emits the update output and then fails at the health check without starting an emulator:

```fish
set SMOKE (mktemp -d)
go build -ldflags "-X github.com/localstack/lstk/internal/version.version=0.0.1" -o $SMOKE/bin/lstk .
set -x DOCKER_HOST tcp://localhost:1
printf '[[containers]]\ntype = "aws"\ntag  = "latest"\nport = "4566"\n\n[cli]\n' > $SMOKE/config.toml
set M $SMOKE/mise/.local/share/mise/installs/github-localstack-lstk/latest
mkdir -p $M; cp $SMOKE/bin/lstk $M/
```

```fish
# enabled / disabled / invalid
HOME=$SMOKE $SMOKE/bin/lstk --config $SMOKE/config.toml start --non-interactive        # note
HOME=$SMOKE LSTK_CHECK_FOR_UPDATE_ON_STARTUP=false $SMOKE/bin/lstk --config $SMOKE/config.toml start --non-interactive  # silent
HOME=$SMOKE LSTK_CHECK_FOR_UPDATE_ON_STARTUP=quiet $SMOKE/bin/lstk --config $SMOKE/config.toml start --non-interactive  # rejected

# externally managed
HOME=$SMOKE $M/lstk --config $SMOKE/config.toml start --non-interactive   # note names mise
HOME=$SMOKE $M/lstk update                                                # refuses
HOME=$SMOKE $M/lstk update --json                                         # UPDATE_EXTERNALLY_MANAGED

# interactive (needs a TTY)
HOME=$SMOKE $SMOKE/bin/lstk --config $SMOKE/config.toml start             # [U] / [R] / [N]
```

Pressing `[U]` performs a real download; set `LSTK_UPDATE_GITHUB_API_ENDPOINT` and `LSTK_UPDATE_GITHUB_DOWNLOAD_ENDPOINT` to a dead address to make it inert.

</details>

## Interaction with #482 (bundled extensions)

#482 landed in parallel and touches the same file.

**A silent one.** `detectMissingBundle` guards on `InstallBinary`, so introducing `InstallExternal` would have quietly removed the missing-bundle hint for mise/nix/asdf installs — they used to classify as `InstallBinary`. It now covers external installs and names the managing tool (`reinstall lstk through mise`) instead of `#482`'s release-download instruction, which would install outside the manager and be overwritten on its next sync.

## Open questions

**Where the update prompt sits in the start flow.** `NotifyUpdate` runs as the first action of the start goroutine, ahead of the Docker health check, auth, and the emulator picker. On a genuine first run `config.toml` does not exist yet — the picker creates it — so `[N]` has nowhere to persist and is omitted rather than offered and silently dropped:

lstk-update-prompt-v2-first-run

The trade-off: moving the notification after the picker would make `[N]` available on every run, but prompting early is worth more — a user on an old or broken CLI should be offered the update before the CLI attempts real work, and a late prompt would be preempted by exactly the Docker failure an update might fix. The gap is narrow, since a first run almost always means a fresh install already on the latest version. Deliberately left as is; flagging it for a second opinion.

## Review

Worth a human review. New user-facing surface (config key, env var, flag, error code) and a changed prompt, plus one behaviour change that is not opt-in: a detected mise/nix/asdf install gets a note instead of the prompt.

Automatic detection of externally managed (e.g., mise package installer) provides convenience by default, but add a bunch of complexity. The alternative would be solely relying on manual configuration.

Openspec preview: https://github.com/localstack/lstk/tree/4330e4d/openspec/changes/add-update-check-config

## Related

#465 was an earlier, independent attempt at the same design, closed 2026-08-25.

https://github.com/localstack/lstk/pull/482 extending updates for extensions

Closes DEVX-1029


openspec/changes/add-update-check-config/proposal.md
```diff
@@ -0,0 +1,36 @@
+## Why
+
+`lstk` has no way to stop asking about updates. The only escape hatch today is "Skip this version" (`cli.update_skipped_version`), which the weekly release cadence invalidates within days — so a user on a fast-moving channel is prompted again almost every day (DEVX-1029, reported from Slack). That option is removed here rather than kept alongside the new one: it bought days, not quiet, and left the prompt with three ways to say "no".
```
joe4dev, line 3: When do we include openspec documentation in a PR (as present in this repo for several changes) and when not? What's your recommendation here?
/cc @anisaoshafi

anisaoshafi: We tend to include openspec documentation when we're implementing a bigger feature/change and where we make decisions (don't just let the agent decide for us) when there's more than one obvious alternative.
For simpler work, it's not needed and doesn't add value. 
This is the convention we've discussed, though it's not enforced. If you think it's a good idea, just add it.

joe4dev: Thanks for sharing.

We could probably skip specs if considering the disable check updates feature individually, but given the context of having multiple implementations (prior persistent skip, version skip) and broader discussion in Slack [here](https://linear.app/localstack/issue/DEVX-1029/allow-permanently-disabling-lstk-update-checks#comment-8278cbac) including 5+ people, it's probably worth adding here. There's also the open question on first start (missing config) affecting the general startup lifecycle.

joe4dev:
@gtsiolis What's your thought regarding the open question (see Section "Open questions")? Do you see a UX issue in having the `[N] Never ask again` option skipped at the first run?

joe4dev:
@carillan81 What's your suggestion on how update notifications for bundled extensions work in this context (e.g., externally-managed lstk)? Do we to treat lstk and lstk extension updates separately? Happy to discuss.
⚠️ I think we should align https://github.com/localstack/lstk/pull/482 and this PR to avoid unintended interactions.

carillan81:
Extensions are only updated when `lstk` is updated. They are a package and there is no way to update extensions right now if they are not updated with `lstk`. That specific PR is specific to include extensions in the update process of a manually installed `lstk`. I don't see any need of notifications.

joe4dev:
> Extensions are only updated when `lstk` is updated. They are a package and there is no way to update extensions right now if they are not updated with `lstk`. That specific PR is specific to include extensions in the update process of a manually installed `lstk`. I don't see any need of notifications.

Great, thank you for the clarification @carillan81  👍 

Then, our PRs complement each other nicely ✨

internal/update/external_install.go
```diff
@@ -0,0 +1,114 @@
...
+		Value: "lstk update --force",
+	}
+}
+
+// blockSelfUpdate reports why an in-place binary replacement must not be
+// attempted, or nil when it may proceed.
+//
+// Homebrew and npm are never blocked: they delegate to `brew upgrade` and
+// `npm install -g`, which own their install directory. An indeterminate probe
+// does not block either — guessing "read-only" from an unrelated I/O error
```
anisaoshafi, line 95: good decision to not block on probes unknown to lstk 💯

anisaoshafi (approved):
I didn't review in detail, but the overall logic looks good ✨ 
Thanks for the comprehensive set of screenshots in `Manual testing` section in PR description.

I also like the tradeoff you took on the open question to drop the option `[N]` when config.toml doesn't exist.

internal/config/default_config.toml
```diff
@@ -61,3 +61,7 @@ port = "4566"    # Host port the emulator will be accessible on
 # [env.ci]
 # SERVICES = "s3,sqs"
 # EAGER_SERVICE_LOADING = "1"
+
+# CLI behavior
+[cli]
+# update_check = "notify"  # Update check on start: "prompt" (default), "notify", "off"
```
gtsiolis: **suggestion**: Could we simplify this to a boolean setting, like `check_for_update_on_startup = false`? When enabled, we could preserve the existing prompt for self-managed interactive installations and show a non-blocking notice for externally managed ones, as discussed in the [relevant discussion](https://linear.app/localstack/issue/DEVX-1029/allow-permanently-disabling-lstk-update-checks#comment-40426154). When disabled, it would skip the check entirely. This could cover the use cases without asking users to choose between prompting and notifying. 💭

Boolean behavior coverage:

| Setting | Self-managed, interactive | Externally managed |
|--------|--------|--------|
| true | Prompt | Non-blocking notice |
| false | No check | No check |

gtsiolis: **praise**: Thanks for adding and persisting this option here in the configuration.

gtsiolis (commented):
Thanks, @joe4dev! Added one suggestion, let me know what you think.

internal/config/default_config.toml
joe4dev: My thoughts:

* Pros: Simpler user configuration with self-describing flag `LSTK_CHECK_FOR_UPDATE_ON_STARTUP=0`; no more magic [N] Never ask again conversion to notify
* Con: Self-managed users (and failed detection of externally-managed) cannot opt into the non-blocking notify-only behavior anymore -> more users disable update check altogether

I think we should go for the simplification if we don't plan to offer a notify-only option.

gtsiolis: I can’t think of a use case or user need for an explicit notify-only preference on self-managed installations. Externally managed installations would already get that behavior automatically.

I’d go with the simpler boolean for now and revisit if user feedback shows a need for more control. 💯

joe4dev:
@carillan81 I discovered and fixed an interaction between the bundled extension (https://github.com/localstack/lstk/pull/482) and this PR (see PR description).

> A silent one. detectMissingBundle guards on InstallBinary, so introducing InstallExternal would have quietly removed the missing-bundle hint for mise/nix/asdf installs — they used to classify as InstallBinary. It now covers external installs and names the managing tool (reinstall lstk through mise) instead of #482's release-download instruction, which would install outside the manager and be overwritten on its next sync.

Could you review this part and validate whether the adjusted message for externally-managed installations is appropriate?

internal/config/default_config.toml
joe4dev: Done ✅ 
I migrated the config to use the simplified boolean approach:
`[cli] check_for_update_on_startup` and `LSTK_CHECK_FOR_UPDATE_ON_STARTUP`

carillan81:
Everything looks good to go, nice job!
nitpick: There is a duplicated text in a message: ´Reinstall lstk: reinstall lstk using XXX´

internal/update/reinstall.go
```diff
@@ -52,14 +52,24 @@ func DetectMissingBundleFor(command string) (MissingBundle, bool) {
...
-	return MissingBundle{Dir: dir, Reinstall: reinstallInstruction}, true
+	return MissingBundle{Dir: dir, Reinstall: reinstallInstructionFor(info)}, true
+}
+
+// reinstallInstructionFor names the tool that owns an externally-managed
+// install. Sending those to a release download would install outside the
+// manager, leaving it to overwrite the result on its next sync.
+func reinstallInstructionFor(info InstallInfo) string {
+	if info.Method == InstallExternal && info.Manager != "" {
+		return "reinstall lstk through " + info.Manager
```
carillan81, line 70: nitpick: ´reinstall lstk´ text is duplicated in the message

carillan81 (approved):

internal/update/reinstall.go
joe4dev: Thanks for raising this @carillan81  🕵️‍♂️ 

What do you think about the following wording? What wording do you suggest?

```
==> Restore extensions: reinstall lstk through mise
==> Restore extensions: download the latest release from https://github.com/localstack/lstk/releases/latest
```

Motivation:
1. Label focusing on the intention `Restore extensions:` because we detected an installation with missing bundled extensions. My understanding is that re-installing the same version restores extensions and upgrading to the latest release is not strictly necessary. Is that correct @carillan81 ?
2. Keep the action in the message. Let me know if we have another convention.

carillan81: I don't have a strong opinion. I would try to be consistent between message and install method. My vote would be 1, but both of them are acceptable.

joe4dev: I'm not sure whether I understand "try to be consistent between message and install method" 🤔 

Given `<label>: <message`, do you suggest to use `reinstall` independent of the install method?

```
# Externally managed
==> Restore extensions: reinstall lstk through mise
# Manually managed
==> Restore extensions: reinstall lstk from https://github.com/localstack/lstk/releases/latest
```

joe4dev:
@gtsiolis Could you re-review this permanent update check from a UX lens before shipping ?

In particular, I tagged you regarding a discussion on how to phrase the request to reinstall lstk in case of broken bundled extensions (interaction with https://github.com/localstack/lstk/pull/482)

internal/update/reinstall.go
joe4dev: What's your suggestion @gtsiolis ?

carillan81: I'd say, if you installed with nise, suggest reinstall with nise. Same suggestion as install method unless we actively want to direct users to an specific install method (which I don't think is the case)

joe4dev: > I'd say, if you installed with nise, suggest reinstall with nise.

Absolutely.

> Same suggestion as install method

Yes, with the slight adaptation that install != reinstall (e.g., mise install vs. mise install --force) and I believe reinstall is required to fix missing bundles.

gtsiolis: Thanks for addressing the duplicated _Reinstall_ wording. The new _Restore_ label reads well, and the managed-install path correctly points users back to mise. 🐭 

For standalone installs, downloading the latest release can be a little ambiguous because downloading alone does not restore the missing bundled files, right? Suggested changes:

```sh
# For standalone installs

  Error: unknown command "doctor" for lstk
    This lstk release ships bundled extensions, but none are installed in /usr/local/bin.
    ==> See help: lstk -h
    ==> Restore extensions: download the latest release from https://github.com/localstack/lstk/releases/latest

# For a mise-managed install (BEFORE)

  Error: unknown command "doctor" for lstk
    This lstk release ships bundled extensions, but none are installed in ~/.local/share/mise/installs/lstk/0.0.2.
    ==> See help: lstk -h
    ==> Restore extensions: reinstall lstk through mise

# For a mise-managed install (AFTER)

  Error: unknown command "doctor" for lstk
    This lstk release ships bundled extensions, but none are installed in /usr/local/bin.
    ==> Restore extensions: reinstall lstk from the latest release archive:
        https://github.com/localstack/lstk/releases/latest
    ==> See help: lstk -h

For `lstk update --check`:

  Checking for updates...
  > Note: Already up to date (0.0.2)
  > Warning: This lstk release ships bundled extensions, but none are installed in /usr/local/bin.
    Restore extensions: reinstall lstk from the latest release archive:
    https://github.com/localstack/lstk/releases/latest
```

Thoughts?
---

> https://github.com/localstack/lstk/releases/latest

Ideally, links like this would be replaced by short URLs, but should be ok to leave out of scope fo now.

joe4dev: > For standalone installs, downloading the latest release can be a little ambiguous because downloading alone does not restore the missing bundled files, right?

Yes, therefore I changed "download" (pure download) to "reinstall" (download and put at the right place with the right permissions):

```
# For standalone installs (BEFORE):
==> Restore extensions: download the latest release from https://github.com/localstack/lstk/releases/latest
# For standalone installs (V1):
==> Restore extensions: reinstall lstk from https://github.com/localstack/lstk/releases/latest
# For standalone installs (AFTER – refined wording based on your suggestion and link on new line):
==> Restore extensions: reinstall lstk from the latest release archive:
    https://github.com/localstack/lstk/releases/latest
```

@gtsiolis I don't understand the mise example from above. Why do you want to replace a managed-install with standalone instructions?

```
# For a mise-managed install (BEFORE)
==> Restore extensions: reinstall lstk through mise
# For a mise-managed install (AFTER)
==> Restore extensions: reinstall lstk from the latest release archive:
    https://github.com/localstack/lstk/releases/latest
```

In my opinon, reinstalling through mise should use `mise install lstk --force` because we never want to manually interfere with managed packages (e.g., nix with root).


sidenote: why does the example change the order of `==> See help: lstk -h` (inconsistent managed vs. standalone)?

joe4dev: For reference, that's how the standalone messages now look like (after adjusting wording, newlines, and formatting to your example):

```
lstk deploy, exit 1:
Error: unknown command "deploy" for lstk
  This lstk release ships bundled extensions, but none are installed in /private/tmp/claude-501/-Users-joe-Projects-LocalStack-appinspector-workspace-lstk/0765a94a-24b7-4bdb-b2b0-66cd4f80b493/scratchpad/render.
  ==> See help: lstk -h
  ==> Restore extensions: reinstall lstk from the latest release archive:
      https://github.com/localstack/lstk/releases/latest

lstk update --check, plain mode, exit 0:
Checking for updates...
> Note: Already up to date (0.0.2)
> Warning: This lstk release ships bundled extensions, but none are installed in /private/tmp/claude-501/-Users-joe-Projects-LocalStack-appinspector-workspace-lstk/0765a94a-24b7-4bdb-b2b0-66cd4f80b493/scratchpad/render.
  Restore extensions: reinstall lstk from the latest release archive:
  https://github.com/localstack/lstk/releases/latest

lstk update --check, interactive UI (90 columns wide, colors removed), exit 0:
⣽ Checking for updates
⣻ Checking for updates
⢿ Checking for updates
⡿ Checking for updates
> Note: Already up to date (0.0.2)
> Warning: This lstk release ships bundled extensions, but none are installed in
/private/tmp/claude-501/-Users-joe-Projects-LocalStack-appinspector-workspace-l
           stk/0765a94a-24b7-4bdb-b2b0-66cd4f80b493/scratchpad/render.
  Restore extensions: reinstall lstk from the latest release archive:
  https://github.com/localstack/lstk/releases/latest
```

gtsiolis: @joe4dev You're right, probably a copy-paste error. It was intended to show the standalone-install state.

gtsiolis (approved):

PR #494, opened 2026-09-08
Skip release and integration workflows in forks

joe4dev:
## Motivation

A fork with Actions enabled runs our tag and release workflows against its own repository. They fail on empty secrets — but a `v*.*.*` tag pushed in a fork gets far enough for goreleaser to cut a real GitHub release there, since a fork's `GITHUB_TOKEN` is populated and only the Homebrew tap push fails. The integration suite can likewise only ever fail in a fork, because `LOCALSTACK_AUTH_TOKEN` is never present.

## Solution

`if: github.repository == 'localstack/lstk'` on the five jobs that tag, publish, or run the integration suite. The guard is constant-true in this repo on every trigger — including the `workflow_call` from `automated-release.yml` — so upstream CI is unchanged; only a fork's own runs skip.

Repository-scoped rather than the `head.repo.fork` guard the SARIF (Static Analysis Results Interchange Format) uploads use: `test-integration`'s per-OS job names are required status checks, and skipping a matrix job may never report them, which would leave a fork PR unmergeable rather than merely red. Fork PRs into this repo are unaffected — `github.repository` is the base repo there, so the suite still runs and still fails on the missing token.

<details>
<summary>Docs</summary>

Nothing user-facing — CI configuration only. One paragraph added to CLAUDE.md's Release Process section so a future tag/publish workflow picks up the guard.

</details>

## Review

Human review advised — touches four release-pipeline workflows. Verified by inspection and `actionlint`; not exercised against a real fork (requires merging into main first).

## Related

Resolves DEVX-1100

### DIFF
--- a/.github/workflows/automated-release.yml
+++ b/.github/workflows/automated-release.yml
@@ -18,6 +18,9 @@ concurrency:
 jobs:
   check-changes:
     name: Check for changes since last release
+    # Root of the release graph: every other job needs it, so one guard keeps a
+    # fork's scheduled or dispatched run from tagging or releasing.
+    if: github.repository == 'localstack/lstk'
     runs-on: ubuntu-latest
     outputs:
       has_changes: ${{ steps.check.outputs.has_changes }}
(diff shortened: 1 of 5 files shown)


anisaoshafi (approved):
Thanks for fixing this 🎉

PR #495, opened 2026-09-08
Show one tip on first start, document completion setup in the CLI

joe4dev:
## Motivation

#484 introduced a UX regression by rendering two tips on the first lstk startup (shell completion + rotation tip). Furthermore, the review also flagged the long docs URL as poor terminal UX.

## Solution

**One tip per run, enforced structurally rather than by convention.**

- `selectTip` (new `internal/container/tips.go`) is the only place that decides which tip to show, and it returns one string, not a list. The first-run completion tip outranks the rotating per-emulator tips: first run happens once per install, while the rotating tips come back on every later start.
- `container.Start` is the only place that emits it. Its body moved to an unexported `start`, leaving the public entry point a thin wrapper around the single emit. `emitPostStartPointers` no longer emits a tip, and `internal/ui` no longer emits one either — `RunOptions.CompletionTip` is replaced by `StartOptions.FirstRun`.

Adding a tip now means ranking it inside `selectTip`; there is nowhere else to put it. The rule is written up under "Post-start tips" in `CLAUDE.md`.

**The tip names a command instead of linking to docs.**

```
> Tip: Set up tab completion: lstk completion
```

A URL in terminal output can't be clicked, rots, and drifts from an already-shipped binary. So `lstk completion` now carries the setup instructions itself; the per-shell subcommands forward to it, and `lstk docs` renders it, which makes the CLI the source and the docs site the follower.

User-visible change: a first interactive start shows the completion tip and nothing else. Every other run is unchanged, rotation included.

## Shell completion

Bare `lstk completion` (and `lstk completion --help`) now prints:

```
Generate shell completion scripts for lstk.

To load completions:

Bash:

  # Load in current session
  eval "$(lstk completion bash)"

  # Load in new sessions (Linux)
  echo 'eval "$(lstk completion bash)"' >> ~/.bashrc

  # Load in new sessions (macOS)
  echo 'eval "$(lstk completion bash)"' >> ~/.bash_profile

Zsh:

  # Load in current session
  autoload -Uz compinit && compinit
  source <(lstk completion zsh)

  # Load in new sessions (Linux, macOS)
  echo 'autoload -Uz compinit && compinit' >> ~/.zshrc
  echo 'source <(lstk completion zsh)' >> ~/.zshrc

Fish:

  # Load in current session
  lstk completion fish | source

  # Load in new sessions (Linux, macOS)
  lstk completion fish > ~/.config/fish/completions/lstk.fish

PowerShell:

  # Load in current session
  lstk completion powershell | Out-String | Invoke-Expression

  # Load in new sessions (Windows, Linux, macOS)
  if (!(Test-Path $PROFILE)) { New-Item -ItemType File -Path $PROFILE -Force }
  lstk completion powershell | Out-File -Append -Encoding utf8 $PROFILE
```

Every recipe is the same shape: the current-session command, or that same command appended to the shell's startup file.

### Optional alternatives considered and dropped

Verified on macOS 15 / Apple Silicon.

| Dropped option | Why |
| --- | --- |
| Zsh: `lstk completion zsh > "${fpath[1]}/_lstk"` | `fpath[1]` is root-owned on stock macOS (`/usr/local/share/zsh/site-functions`), so a normal user gets `permission denied`. It doesn't exist on stock macOS yielding `zsh: no such file or directory`. Under oh-my-zsh it is a plugin directory. On my private mac, it points to `~/.docker/completions` |
| Zsh: `lstk completion zsh > $(brew --prefix)/share/zsh/site-functions/_lstk` | Silently does nothing on macOS's built-in zsh — that directory is not on the default `$fpath` (default `fpath[1]` is the *Intel* prefix while Homebrew sits at `/opt/homebrew`). It needs Homebrew's own zsh, or `FPATH="$(brew --prefix)/share/zsh/site-functions:${FPATH}"` before `compinit`. |
| Bash: `lstk completion bash > /etc/bash_completion.d/lstk` and the `$(brew --prefix)/etc/bash_completion.d/lstk` variant | Both need the bash-completion package, which the `eval` form does not — lstk's script ships its own `_get_comp_words_by_ref` fallback (DEVX-950). They only buy lazy loading on first Tab instead of sourcing at every shell start. |
| Guarding the startup-file lines, e.g. `[[ $commands[lstk] ]] && source <(lstk completion zsh)` | Would avoid `command not found: lstk` in every new shell after an uninstall, but roughly doubles the length of each line. Happy to add it to bash and zsh if reviewers prefer. |

Two things that look optional but are not:

- The zsh `compinit` line is required. Cobra's generated script calls `compdef` on line 2, and `compdef` does not exist until `compinit` has run — without it you get `compdef: command not found` and nothing registers. A duplicate `compinit` (oh-my-zsh already runs one) is harmless. Sourcing costs ~8 ms per shell start.
- PowerShell uses `Out-File -Encoding utf8` rather than `>>`, which writes UTF-16LE on PowerShell 5.1 and leaves a profile the shell cannot parse.

<details>
<summary>Docs</summary>

The CLI is now the source of truth for shell-completion setup, and the docs site should follow it rather than the reverse.

Handled in localstack/localstack-docs#937, which rewrites the [Shell completions](https://docs.localstack.cloud/aws/developer-tools/running-localstack/lstk/#shell-completions) section to match `lstk completion` — the page currently recommends `${fpath[1]}` and the Homebrew `site-functions` path for zsh (neither works on stock macOS — see the table above), omits the required `compinit` step, and has no PowerShell instructions. The same section is duplicated on the Azure page; that PR updates both.

No new command, flag, or env var. The first-start tip no longer links to the docs, so the `#shell-completions` anchor is no longer referenced from the CLI.

</details>

### First startup tip

No more docs link and adjusted message:
Screenshot 2026-09-10 at 19 00 11

### Shell Completions

- [x] @joe4dev I tested Bash, Zsh, Fish, PowerShell on macOS and PowerShell on Windows
- [x] @dfangl Bash and Zsh on Linux (Fish and PS should be the same as macOS)

## Review

Human review advised — it changes user-facing output and help text on the shared start path, and @gtsiolis raised the original concern on #484, so it's worth confirming this is the resolution he had in mind.

## Related

Towards DEVX-821

Docs PR with synchronized shell completion changes: https://github.com/localstack/localstack-docs/pull/937
* Preview: https://devx-821-lstk-shell-completi.localstack-docs.pages.dev/aws/developer-tools/running-localstack/lstk/#shell-completions


internal/container/tips.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/localstack/lstk/internal/config"
+	"github.com/localstack/lstk/internal/output"
+)
+
+// completionTip fires on first run, not install: no install path has a usable
+// hook (npm's package.json is generated, binaries have none). Must stay a plain
+// MessageEvent — ui.Run renders no DeferredOutput, so a deferred event is lost.
+const completionTip = "> Tip: Enable tab completion for your shell: lstk completion [bash|zsh|fish|powershell] " +
+	"See https://docs.localstack.cloud/aws/developer-tools/running-localstack/lstk/#shell-completions"
```
gtsiolis: **suggestion**: I'd skip the long URL all together, and only consider adding one when we add a short URL as described in https://github.com/localstack/lstk/pull/484#issuecomment-5585430187. A good short URL for this CLI UX case could be `lstk.dev/shell-completion`.

internal/container/tips.go
```diff
@@ -0,0 +1,64 @@
...
+	// Nothing came up (no containers configured), so nothing to tip about.
+	if emulatorType == "" {
+		return
+	}
+	if tip := selectTip(emulatorType, firstRun, interactive); tip != "" {
+		sink.Emit(output.MessageEvent{Severity: output.SeveritySecondary, Text: tip})
+	}
+}
+
+// selectTip returns the one tip to show after a start, or "" for none.
```
gtsiolis, line 30: **praise**:  The one-tip change addresses the original concern. 🙏

internal/container/tips.go
```diff
@@ -0,0 +1,64 @@
...
+	"math/rand/v2"
+
+	"github.com/localstack/lstk/internal/config"
+	"github.com/localstack/lstk/internal/output"
+)
+
+// completionTip fires on first run, not install: no install path has a usable
+// hook (npm's package.json is generated, binaries have none). Must stay a plain
+// MessageEvent — ui.Run renders no DeferredOutput, so a deferred event is lost.
+const completionTip = "> Tip: Enable tab completion for your shell: lstk completion [bash|zsh|fish|powershell] " +
```
gtsiolis: **suggestion**: Until we have a better domain we can use for links, and regardless that a good practice I've seen as alternative to interactive prompts to set up completion is to show instructions on `completion` command at the top.

So we can show this as the tip.

```suggestion
const completionTip = "> Tip: Set up tab completion: lstk completion
```

Then, we can prefix the compltion command output above with...

```
 % lstk completion

Generate the autocompletion script for lstk for the specified shell.

To load completions:

Bash:
  # Verbose specific instructions [...]
  $ source <(lstk completion bash)

  # To load completions for every new session, add to ~/.bashrc:
  eval "$(lstk completion bash)"

  [...] {Any more instructions needed}

Zsh:
  # Verbose specific instructions [...]
  [...]

Usage: lstk completion [flags]

Commands: [...]

[...]
```

gtsiolis (commented):
Thanks for tackling this, @joe4dev! Left some comments, let me know what you think!

internal/container/tips.go
joe4dev: We can do that if `lstk completion` is self-descriptive (your suggestion [below](https://github.com/localstack/lstk/pull/495#discussion_r3978095294))

joe4dev: What are the next steps regarding short URLs (the migration link is another [recent example](https://github.com/localstack/localstack-cli-standalone/pull/40#discussion_r3960578972))?
`lstk.dev/shell-completion` sounds good if we can have product-specific domains or `ls.cloud` as a generic domain. Maybe, platform can help here?
cc @jbeemster

internal/container/tips.go
joe4dev: I like the idea of self-describing commands. Main challenges:
* Maintenance: We need to keep the instructions in sync with the docs and test properly
* Verbosity: Bash, Zsh, Powershell, Fish

Our docs miss instructions for powershell and macOS (bash, zsh)

~~For example, your suggestion for bash already deviates from the docs and `~/.bashrc` is a not the right place because it also loads for non-interactive shells.~~
EDIT: [bash-completion](https://github.com/scop/bash-completion#macos-os-x) recommends using `~/.bashrc` but loading it from `~/bash_profile` might not be configured everywhere

The aws CLI [completion instructions](https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-completion.html) summarize where it can go:

```
Bash– .bash_profile, .profile, or .bash_login
Zsh– .zshrc
Tcsh– .tcshrc, .cshrc, or .login
```

Do you think we should fix/test these properly and keep in-sync?

internal/container/tips.go
gtsiolis: I think choosing the domain should remain a product decision, but open to discuss any option and definitely consider non-`.cloud` domains. I consider `lstk.dev` to be a product-specific domain but I see it's taken. ➕ 

In the past I hijacked[[1]](https://localstack-cloud.slack.com/archives/C07LMD90LTF/p1730197122958529) `localstack.dev` for product marketing[[2](https://localstack-cloud.slack.com/archives/C07JJQVFY86/p1740740843752499)][[3](https://localstack-cloud.slack.com/archives/C07JJQVFY86/p1738336611222479)][...] but someone broke all links[[4](https://localstack-cloud.slack.com/archives/C07LMD90LTF/p1773246085874359)]. 😭 

Now we recently started using `localstack.dev` for sandboxes[[5](https://linear.app/localstack/issue/DEVX-1041/set-up-localstackdev-as-the-sandbox-base-domain)], but we need a shorter domain anyway.

For this PR, I'd skip the domain or short URLs set up.

But longer-term introducing short URLs can be helpful for many use cases, including the migration link you mentioned above. We just have to make sure no one breaks the links again.

internal/container/tips.go
gtsiolis: On challenges, I'd not sync anything (⚠️). Here's what I'd do:
1. Delete the duplicate information on docs and leave CLI carry the single source of truth. Then docs can only reference the command and rely on the CLI help output on specifics. IMO, keeping the steps to enable shell completion in docs is a sign our docs can be better.
2. Focus only on `bash`, `zsh`, `fish`, and `powershell`. Some verbosity is ok here as a tradeoff.

joe4dev: 1. I noticed that the CLI docs are even duplicated within the docs between AWS and Azure. I like the idea of a self-documenting product, though I'm concerned we're loosing visibility and discoverability as the docs formatting with the tabs-per-shell  is much nicer than the verbose in-terminal plain text (it's one of the only docs section I used personally). I created a docs PR to update it consistently https://github.com/localstack/localstack-docs/pull/937 What about adding a unit test here that detects de-sync regressions? Or do you prefer removing the docs altogether?
2. Done and tested ✅

internal/container/tips.go
joe4dev: Removed the docs link for now until we have a short-URL solution

internal/container/tips.go
gtsiolis: On [1] I'd generally aim for maintaining a [single source of truth](https://en.wikipedia.org/wiki/Single_source_of_truth) (SSOT) whenever possible which means never ever having to sync or check for de-sync regressions, but only reference links. In other words, I would delete the duplicate shell completion instructions in the docs and only mention the `completion` command.

cmd/completion.go
```diff
@@ -132,26 +130,85 @@ func selfContainBashCompletion(completionCmd *cobra.Command) {
...
+// It lives on the `completion` command alone — the first-start tip points at
+// bare `lstk completion` (PR #495 review), so that command has to answer the
+// question with no URL to follow, and the per-shell subcommands forward here
+// rather than keeping a second copy. `lstk docs` renders it too, so the docs
+// site follows the CLI instead of being kept in step by hand.
+//
+// %[1]s is the binary name. Command lines are indented so wrapText
+// (cmd/help.go) leaves them intact — it reflows unindented prose to the
+// terminal width, which would break a command mid-word on a narrow one.
+const completionSetup = `Generate a tab-completion script for your shell. If you installed via Homebrew, completions are set up automatically.
```
gtsiolis: > If you installed via Homebrew, completions are set up automatically.

**question**: Could we skip this, even if true? It's also breaking when resizing terminal window.


```suggestion
const completionSetup = `Generate shell completion scripts for lstk.
```

cmd/completion.go
```diff
@@ -132,26 +130,85 @@ func selfContainBashCompletion(completionCmd *cobra.Command) {
...
+
+To load completions temporarily or permanently:
```
gtsiolis: **question**: Could we safely skip the temporarily or permanently part?

```suggestion
To load completions:
```

cmd/completion.go
```diff
@@ -132,26 +130,85 @@ func selfContainBashCompletion(completionCmd *cobra.Command) {
...
+
+Bash:
```
gtsiolis, line 149: **praise**: Amazing, thanks for caring about the details here. ❤️

gtsiolis (approved):
Nearly perfect, adding a latest screenshot of the current state for reference. 🏁 

Image

cmd/completion.go
joe4dev: Yes, Homebrew users hopefully discover automatically that it already works before trying to configure twice. I'll mention it in the docs for reference.

internal/container/tips.go
joe4dev: I removed the instructions in the updated docs PR ready for review: https://github.com/localstack/localstack-docs/pull/937

I kinda liked the pretty instructions online, but also appreciate easier maintenance with SSOT. Let's try it out 🚀

internal/container/tips.go
```diff
@@ -0,0 +1,66 @@
...
+	"math/rand/v2"
+
+	"github.com/localstack/lstk/internal/config"
+	"github.com/localstack/lstk/internal/output"
+)
+
+// completionTip fires on first run, not install: no install path has a usable
+// hook (npm's package.json is generated, binaries have none). It names the bare
+// command rather than a URL — `lstk completion` carries the per-shell setup
+// itself (completionShells in cmd/completion.go), so there is nothing to follow
```
anisaoshafi: `completionShells` is a leftover. I think the comments are a bit too verbose, which also increases the chances of them becoming outdated.

anisaoshafi (approved):
Looks good, thanks for tackling this in a consistent way ✨ 

non-blocker: I have a problem reading the comments, they continue to be quite verbose. Not a problem specific just to this PR, but lstk PRs in general. We need to continue tweaking the agent skills/instructions.

anisaoshafi:
appreciate the thorough testing across different shells 👏🏼

joe4dev:
> Looks good, thanks for tackling this in a consistent way ✨
> 
> non-blocker: I have a problem reading the comments, they continue to be quite verbose. Not a problem specific just to this PR, but lstk PRs in general. We need to continue tweaking the agent skills/instructions.

thanks for the reminder. Agreed that the current guidance is not working:

> - Keep comments short: 1–3 lines is the norm, ~6 is a ceiling that needs a real reason (a subtle invariant, upstream behaviour that will surprise the reader, a rejected alternative someone will otherwise re-attempt). Long comments are harder to read and go stale — they are a maintenance burden, not extra rigor.

It started adding multiple blocks to overcome the 6-line limit 🤦 

Any ideas how to guide the agents?

Could we learn from https://github.com/softaworks/agent-toolkit/blob/main/skills/writing-clearly-and-concisely/README.md , which I discovered from this blog post https://medium.com/engineering-in-the-age-of-ai/how-to-use-ai-to-generate-texts-that-sound-like-a-human-would-actually-write-them-c7eef78e0b42
I like the example:

Before:

> // Note that it is important to initialize the authentication module
// before the router in order to ensure that route guards function correctly.

After:

> // don't reorder: auth must init before the router or the guards won't fire

internal/container/tips.go
joe4dev: ups, removed in [`a02aba5` (this PR)](https://github.com/localstack/lstk/pull/495/changes/a02aba518dbfdcbc7cc3c5c2d1e95d5346ae8f7f) and shortened comments in general.

anisaoshafi:
> > Looks good, thanks for tackling this in a consistent way ✨
> > non-blocker: I have a problem reading the comments, they continue to be quite verbose. Not a problem specific just to this PR, but lstk PRs in general. We need to continue tweaking the agent skills/instructions.
> 
> thanks for the reminder. Agreed that the current guidance is not working:
> 
> > * Keep comments short: 1–3 lines is the norm, ~6 is a ceiling that needs a real reason (a subtle invariant, upstream behaviour that will surprise the reader, a rejected alternative someone will otherwise re-attempt). Long comments are harder to read and go stale — they are a maintenance burden, not extra rigor.
> 
> It started adding multiple blocks to overcome the 6-line limit 🤦
> 
> Any ideas how to guide the agents?
> 
> Could we learn from https://github.com/softaworks/agent-toolkit/blob/main/skills/writing-clearly-and-concisely/README.md , which I discovered from this blog post https://medium.com/engineering-in-the-age-of-ai/how-to-use-ai-to-generate-texts-that-sound-like-a-human-would-actually-write-them-c7eef78e0b42 I like the example:
> 
> Before:
> 
> > // Note that it is important to initialize the authentication module
> > // before the router in order to ensure that route guards function correctly.
> 
> After:
> 
> > // don't reorder: auth must init before the router or the guards won't fire

I like the repo you linked, the examples there look much neater and simpler. We incorporate them to our existing skills and instructions instead of adding them as a separate skill.
I want to follow up on this later, likely next week🤞🏼

PR #499, opened 2026-09-09
Distinguish proxied tool exits from lstk errors in command telemetry

joe4dev:
## Motivation

`lstk aws` fails for two very different reasons: an lstk preflight error, or the AWS CLI itself rejecting the user's command. Telemetry recorded both identically, so the error dashboard ranks users' own CLI mistakes as lstk errors — that is how "exit status 252" became the no. 1 lstk error (DEVX-1003).

## Solution

`lstk_command` events gain `result.proxy_error`, true when the failure came from a wrapped tool the user asked for. It describes the error's origin, not whether a tool was proxied — a successful `lstk aws s3 ls` is `false`.

The discriminator is **not** an `*exec.ExitError` in the chain, as the ticket proposed: lstk shells out for its own purposes too, and those exits are lstk's failures. Instead each proxy exec site declares ownership via `proc.MarkUserToolExit`. Forgetting it on a future proxy only loses a data point; marking one of lstk's own execs would hide an lstk bug behind the user's name.

That forced one split: `azurecli.Exec` served both the `lstk az` passthrough and lstk's own `setup azure`/interception calls. It is now `Exec` (the user's, marks) and `Run` (lstk's, does not) over a shared `execAz`.

`proxy_error` is deliberately not `omitempty`, so absence means only "emitted before this field existed".

### The user's tool failed → `proxy_error: true`

| Command | `exit_code` | Why |
|---|---|---|
| `lstk aws s3 lss` | 252 | AWS CLI usage error — not lstk's fault |
| `lstk aws s3 ls` against a service error | the CLI's | Emulator rejected it, lstk only forwarded |
| `lstk terraform apply` on a bad config | terraform's | terraform rejected the user's HCL |
| `lstk <extension> …` exiting non-zero | the tool's | Extension's own outcome |

The exit code is always the wrapped tool's own, unchanged.

### lstk failed → `proxy_error: false`

| Command | `exit_code` | Why |
|---|---|---|
| `lstk aws s3 ls`, Docker down | 1 | `runtime not healthy` — never reached the tool |
| `lstk aws s3 ls`, emulator not running | 1 | lstk preflight |
| `lstk aws s3 ls`, success | 0 | No failure to attribute |

### lstk shells out, but the failure is still lstk's → `proxy_error: false`

These are the cases the rejected `*exec.ExitError` discriminator would have got wrong. The user never typed these invocations and cannot fix them by changing their command line.

| Command | Shells out to | Resulting `error_msg` |
|---|---|---|
| `lstk setup azure` | `az cloud list` | `could not list Azure clouds: az […]: exit status 1` |
| `lstk az stop-interception` | `az cloud show` | `could not determine the active Azure cloud: az […]: exit status 1` |
| `lstk update` (Homebrew install) | `brew upgrade` | `update failed: exit status 1` |
| `lstk terraform init` (S3 backend) | `aws s3api create-bucket` | `creating state bucket …: exit status N` |

Note the contrast: `lstk aws` with the AWS CLI exiting 252 gives `error='exit status 252'` and `proxy_error: true`, while `lstk update` with `brew` exiting 1 gives `error='update failed: exit status 1'` and `proxy_error: false`. Both carry an `*exec.ExitError` and both read "exit status N" — only the call site knows which is the user's.

### For the analytics side

- **The field is `proxy_error`, not `proxied` as the ticket specifies.** `proxied` reads as "this invocation was proxied", which is false for a successful `lstk aws s3 ls` that the field would still mark `false`. Renaming needs a nod from whoever builds the dashboard panel.
- A signal-terminated tool (Ctrl-C on `lstk terraform apply`, CI `timeout`) records `proxy_error: true, exit_code: -1`. Not an lstk error, but not a usage error either — exclude `exit_code == -1`.
- The error-rate *denominator* (proxy invocations including successes) is not derivable from `proxy_error` alone, since the field is failure-only per the ticket. It still needs the proxy command-name list.

### Also

The integration env inherited an ambient `LOCALSTACK_DISABLE_EVENTS=1`, silently disabling every telemetry assertion. 43 integration tests that fail locally for developers who export it now pass. Separate commit.

<details>
<summary>Docs</summary>

Nothing user-facing. No command, flag, output, env var, or documented behavior changes — this adds a field to the internal `lstk_command` analytics payload, which no file under `docs/` describes. The consumer is the analytics dashboard, not the CLI user.

</details>

## Review

Human review advised — it changes what the analytics error ranking means, and the `azurecli` split touches `lstk setup azure` and interception, whose integration tests skip without a working Azure CLI.

## Related

Resolves DEVX-1004

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -182,6 +182,11 @@ When lstk's stdout and stderr are both terminals, `lstk aws` runs the child via 
 
 `lstk az` has the identical wrapper pattern and gets the identical treatment (DEVX-1028, and the DEVX-1049 pager-input fix arrives through the shared `proc.RunInPTY`): `azurecli.Exec` takes the same `usePTY` parameter under the same both-streams-are-terminals condition, and `azurecli`'s own `execEnv` injects `PYTHONUNBUFFERED=1` (leaving a user-set value alone). Unlike the frozen aws v2 binary, every `az` distribution runs a real Python interpreter, so that env var is effective there — the PTY additionally makes az see a terminal, which is what its own terminal-gated output depends on. `azurecli.Run` (short captured-output execs behind `setup azure`/interception) always passes `usePTY: false`.
 
+# Attributing Wrapped-Tool Exits
+
+Separately from how a tool is *run*, each proxy exec site declares *whose* failure a non-zero exit is by passing the result through `proc.MarkUserToolExit`: `lstk aws`, `terraform`, `cdk`, `sam`, the `lstk az` passthrough (`azurecli.Exec` — **not** `azurecli.Run`, which is lstk's own orchestration), and extensions. Telemetry reads it back via `proc.IsUserToolExit` for `result.proxy_error` on `lstk_command` events, which keeps users' own CLI usage errors out of the lstk error ranking (DEVX-1004).
+
+Do not infer this from the runner. `proc.Run` promises signal handling only, and lstk shells out for its own purposes too (`brew upgrade` behind `lstk update`, the `aws` calls behind terraform backend provisioning). A new proxy that forgets the mark loses a data point; marking one of lstk's own execs hides an lstk bug behind the user's name — so leave it off when unsure.
 
 # Snapshots
 
--- a/cmd/extension.go
+++ b/cmd/extension.go
@@ -15,6 +15,7 @@ import (
 	"github.com/localstack/lstk/internal/extension"
 	"github.com/localstack/lstk/internal/log"
 	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/proc"
 	"github.com/localstack/lstk/internal/runtime"
 	"github.com/localstack/lstk/internal/telemetry"
 	"github.com/spf13/cobra"
@@ -82,7 +83,7 @@ func dispatchExtension(ctx context.Context, cfg *env.Env, tel *telemetry.Client,
 	if runErr != nil {
 		errorMsg = runErr.Error()
 	}
-	tel.EmitCommand(ctx, "ext:"+name, "", nil, time.Since(start).Milliseconds(), exitCode, errorMsg)
+	tel.EmitCommand(ctx, "ext:"+name, "", nil, time.Since(start).Milliseconds(), exitCode, errorMsg, proc.IsUserToolExit(runErr))
 
 	return runErr
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -23,6 +23,7 @@ import (
 	"github.com/localstack/lstk/internal/env"
 	"github.com/localstack/lstk/internal/log"
 	"github.com/localstack/lstk/internal/output"
+	"github.com/localstack/lstk/internal/proc"
 	"github.com/localstack/lstk/internal/runtime"
 	"github.com/localstack/lstk/internal/telemetry"
 	"github.com/localstack/lstk/internal/tracing"
@@ -581,7 +582,7 @@ func instrumentCommands(cmd *cobra.Command, tel *telemetry.Client) {
 				errorMsg = runErr.Error()
 			}
 
-			tel.EmitCommand(c.Context(), commandDisplayName(c), subcommand, flags, time.Since(startTime).Milliseconds(), exitCode, errorMsg)
+			tel.EmitCommand(c.Context(), commandDisplayName(c), subcommand, flags, time.Since(startTime).Milliseconds(), exitCode, errorMsg, proc.IsUserToolExit(runErr))
 
 			return runErr
 		}
--- a/internal/awscli/exec.go
+++ b/internal/awscli/exec.go
@@ -108,7 +108,7 @@ func Exec(ctx context.Context, opts ExecOptions, stdout, stderr io.Writer, args 
 		runErr = proc.Run(cmd)
 	}
 
-	if err := runErr; err != nil {
+	if err := proc.MarkUserToolExit(runErr); err != nil {
 		var exitErr *exec.ExitError
 		if errors.As(err, &exitErr) {
 			span.SetAttributes(attribute.Int("aws.exit_code", exitErr.ExitCode()))
--- a/internal/azurecli/exec.go
+++ b/internal/azurecli/exec.go
@@ -31,14 +31,21 @@ func CheckInstalled() error {
 	return nil
 }
 
-// Exec runs `az <args...>`. extraEnv is appended to the inherited process environment
-// (later entries win), letting callers inject AZURE_CONFIG_DIR, proxy, and CA settings
-// without mutating the user's global Azure CLI configuration.
+// Exec runs `az <args...>` on the user's behalf — the `lstk az` passthrough.
+// lstk's own az calls go through Run instead; only Exec attributes a non-zero
+// exit to the user. extraEnv is appended to the inherited process environment
+// (later entries win), letting callers inject AZURE_CONFIG_DIR, proxy, and CA
+// settings without mutating the user's global Azure CLI configuration.
 //
 // When usePTY is true (lstk's stdout and stderr are both terminals), the child's
 // output goes through a pseudo-terminal merged into stdout — see proc.RunInPTY
 // for why; otherwise stdout/stderr are wired as given.
 func Exec(ctx context.Context, extraEnv []string, usePTY bool, stdin io.Reader, stdout, stderr io.Writer, args ...string) error {
+	return proc.MarkUserToolExit(execAz(ctx, extraEnv, usePTY, stdin, stdout, stderr, args...))
+}
+
+// execAz is the shared body of Exec and Run, and claims neither's ownership.
+func execAz(ctx context.Context, extraEnv []string, usePTY bool, stdin io.Reader, stdout, stderr io.Writer, args ...string) error {
 	ctx, span := otel.Tracer("github.com/localstack/lstk/internal/azurecli").Start(ctx, "az cli")
 	defer span.End()
 
@@ -111,11 +118,13 @@ func setIfAbsent(env *[]string, key, value string) {
 	*env = append(*env, prefix+value)
 }
 
-// Run executes `az <args...>` with extraEnv and returns the captured stdout, stderr,
-// and any error. On non-zero exit, the error wraps stderr to aid debugging.
+// Run executes an `az <args...>` call lstk composed itself — the short
+// captured-output execs behind `setup azure` and interception — and returns the
+// captured stdout, stderr, and any error. On non-zero exit, the error wraps
+// stderr to aid debugging. A failure here is lstk's own, not the user's.
 func Run(ctx context.Context, extraEnv []string, args ...string) (stdout, stderr string, err error) {
 	var outBuf, errBuf bytes.Buffer
-	runErr := Exec(ctx, extraEnv, false, nil, &outBuf, &errBuf, args...)
+	runErr := execAz(ctx, extraEnv, false, nil, &outBuf, &errBuf, args...)
 	stdout = outBuf.String()
 	stderr = errBuf.String()
 	if runErr != nil {
--- a/internal/extension/exec.go
+++ b/internal/extension/exec.go
@@ -52,7 +52,7 @@ func Invoke(ctx context.Context, ext *Extension, args []string, runCtx Context) 
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 
-	if err := proc.Run(cmd); err != nil {
+	if err := proc.MarkUserToolExit(proc.Run(cmd)); err != nil {
 		var exitErr *exec.ExitError
 		if errors.As(err, &exitErr) {
 			span.SetAttributes(attribute.Int("extension.exit_code", exitErr.ExitCode()))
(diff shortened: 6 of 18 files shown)


joe4dev:
## Manual test results

Ran 40 scenarios against this branch's `bin/lstk` with a local telemetry sink (`make telemetry-sink`), the real `localstack-aws` and Azure emulators, the installed aws/terraform/sam/cdk/az CLIs, and shell-script fakes where a specific exit or signal behaviour was needed. Every scenario matched the design except the three below. Full table under the toggle.

### Top 3 problematic results

**1. Interactive Ctrl-C on `lstk aws` is recorded as a proxy failure, not a cancellation.** With stdin, stdout and stderr all terminals, lstk runs the tool in a PTY and pumps Ctrl-C as a byte into it, so only the child receives SIGINT. The event reads `proxy_exit_code: 130, cancelled: false`, and the pipe files it under `proxy`. The same interruption sent non-interactively (`kill -INT <lstk>`, stdout piped) records `cancelled: true` correctly. The design's claim that lstk's own signal context sees every interruption does not hold on the PTY path. Proposed fix: have the keystroke pump in `proc.RunInPTY` record that it forwarded an interrupt byte and let `commandResult` read it. Small and exact; I'd do it in this PR.

**2. User-input and environment errors on proxy commands land in the lstk bucket.** `lstk aws --account 123 s3 ls` (`--account must be a 12-digit AWS account id, got "123"`), `lstk aws` against an Azure endpoint, and `aws` missing from PATH all record `proxied: true`, no `proxy_exit_code`, so `error_source = 'lstk'`. That is a true statement about who failed to complete the invocation, but not what a reliability panel wants. Separating them needs the "why" axis. Quick win: `emitValidationError` (`cmd/iac.go`) emits its `ErrorEvent` with no `Code`, so even `--json` reports `INTERNAL_ERROR`; setting `Code: output.ErrValidationError` there is one line. The rest is the `output.ErrorCode` telemetry plumbing already listed as a follow-up.

**3. `LSTK_TF_DRY_RUN=1 lstk terraform plan` records `proxy_exit_code: 0` although terraform never ran.** A phantom tool success, from the "nil return means the tool exited 0" inference. Documented as the accepted exception, and the knob is developer-only, so low priority; noting it here because it is the one place the inference is visibly wrong.

Two earlier suspects turned out to be harness artefacts, not defects: Ctrl-C early in a TUI command does emit an event once the fake terminal answers Bubble Tea's colour and cursor queries (Ctrl-C at 0.3s, at 3s and `q` during `lstk start` all record `cancelled: true`), and `lstk logs` without `--follow` exits 0 on its own before any key is pressed.

<details>
<summary>Full results (40 scenarios)</summary>

Every run: `make telemetry-sink` on 8089, `LSTK_ANALYTICS_ENDPOINT=http://127.0.0.1:8089`, `LOCALSTACK_DISABLE_EVENTS` unset. "Real" = the running emulator and the installed CLI. "Fake X" = a shell script named `X` first on PATH. "Isolated HOME" = temp HOME with a one-block `config.toml`. Ctrl-C cases run in a PTY that answers Bubble Tea's terminal queries. ✓ = observed event matches the design.

| # | Setup | Command | proxied | proxy_exit_code | cancelled | exit_code | |
|---|---|---|---|---|---|---|---|
| 1 | real | `lstk aws s3 lss` | true | 252 | false | 252 | ✓ |
| 2 | real | `lstk aws s3 ls` | true | 0 | false | 0 | ✓ |
| 3 | real | `lstk aws s3 ls s3://missing` | true | 254 | false | 254 | ✓ |
| 4 | real | `lstk tf version` → recorded `terraform` | true | 0 | false | 0 | ✓ |
| 5 | real, dir with invalid HCL | `lstk terraform validate` | true | 1 | false | 1 | ✓ |
| 6 | Azure emulator (`lstk start --type azure`), `lstk setup azure` | `lstk az group list` / `az group lst` / `az account show` | true | 0 / 2 / 0 | false | 0 / 2 / 0 | ✓ |
| 7 | real, empty dir | `lstk sam validate`, `lstk cdk ls` | true | 1 | false | 1 | ✓ |
| 8 | fake aws traps INT and exits 130; PTY | Ctrl-C while `lstk aws s3 ls` runs | true | 130 | **false** | 130 | ✗ #1 |
| 9 | same fake aws, stdout piped | `kill -TERM <lstk>` and `kill -INT <lstk>` | true | 130 | true | 130 | ✓ |
| 10 | fake aws that sleeps | `kill -9 <aws child>` | true | -1 | false | -1 (process 255) | ✓ documented sentinel |
| 12 | `DOCKER_HOST=tcp://localhost:1` | `lstk aws s3 ls` | true | absent | false | 1 | ✓ |
| 13 | emulator stopped | `lstk aws s3 ls` | true | absent | false | 1 | ✓ |
| 14 | empty PATH, `--endpoint-url` to a fake health server | `lstk aws s3 ls` | true | absent | false | 1 | ✓ shape, see #2 |
| 15 | as 14 | `lstk aws --account 123 s3 ls` | true | absent | false | 1 | ✓ shape, see #2 |
| 16 | `--endpoint-url http://127.0.0.1:1` | `lstk aws s3 ls` | true | absent | false | 1 | ✓ |
| 17 | `--endpoint-url` to an Azure-shaped health server | `lstk aws s3 ls` | true | absent | false | 1 | ✓ shape, see #2 |
| 18 | real; project with `backend "s3"`; fake aws exits 5 | `lstk terraform init` | true | absent | false | 5 | ✓ lstk's own child code propagates, as documented |
| 19 | real; project with aws provider initialised | `LSTK_TF_DRY_RUN=1 lstk terraform plan` | true | 0 | false | 0 | ✗ #3 |
| 21 | emulator stopped, no token → login prompt; PTY | `lstk start` + Ctrl-C at 0.3s, at 3s, and `q` | false | absent | true | 1 | ✓ |
| 22 | real; PTY | `lstk logs -f` + `q`, and + Ctrl-C | false | absent | true | 1 | ✓ |
| 23 | fresh HOME, Docker down | `lstk start --json --non-interactive` | false | absent | false | 1 | ✓ |
| 25 | fake az exits 1, Azure config | `lstk setup azure --non-interactive` → `setup azure` | false | absent | false | 1 | ✓ |
| 26 | as 25 | `lstk az stop-interception` → `az stop-interception` | false | absent | false | 1 | ✓ |
| 27 | isolated HOME | `lstk completion bash` | false | absent | false | 0 | ✓ |
| 28 | isolated HOME | `lstk config path` | false | absent | false | 0 | ✓ |
| 29 | `lstk-ref` on PATH | `lstk ref exit 7` → `ext:ref` | false | absent | false | 7 | ✓ |
| 30 | as 29, stdout piped | `lstk ref signal-wait` + `kill -TERM <lstk>` | false | absent | true | 41 | ✓ |
| 31 | `lstk-ref` as `bin/bundled-extensions` + `lstk-extensions.toml` | `lstk ref exit 7` → `ext:ref` | false | absent | false | 7 | ✓ |
| 32–35, 37 | unknown command; `--config /nonexistent`; `--break-me aws`; malformed config; `LOCALSTACK_DISABLE_EVENTS=1` | various | no event | | | | ✓ predicted (pre-existing denominator holes) |

Not run: 11 (tool exits 0 then lstk fails; not triggerable), 24 (`lstk update` with brew failing; needs a Homebrew install), and the pipe-side checks 38–41.

</details>

joe4dev:
Follow-up to the manual test results above, in three commits:

- **#1 fixed** (`Record a Ctrl-C forwarded through the proxy PTY as cancelled`): `proc.RunInPTY` watches the pumped keystrokes for ETX and marks the returned error; `commandResult` reads it via `proc.WasInterrupted`. Integration test drives `lstk aws` in a PTY against a fake `aws` that traps SIGINT and exits 130, and asserts `cancelled: true`.
- **#2 plumbing** (`Carry the shown error code on SilentError and emit it on command telemetry`): `SilentError.Code`, `output.Fail`, `output.ErrorCodeOf`, and `result.error_code` / `result.error_category` on the event, absent when unclassified.
- **#2 emitting** (`Classify preflight and validation failures with error codes for telemetry`): `emitValidationError` now sets `VALIDATION_ERROR`; the 15 `runtime not healthy` preflights go through `runtime.UnhealthyError` (`RUNTIME_UNAVAILABLE`); `HandleNoRunningContainer` sets `EMULATOR_NOT_RUNNING`; not-installed and wrong-emulator preflights and every site that already set a `Code` now carry it on the returned error. `lstk aws --account 123 s3 ls` records `error_code: VALIDATION_ERROR, error_category: USAGE`.

#3 (`LSTK_TF_DRY_RUN`) is left as the documented exception.

joe4dev:
Second re-test round (all 40 scenarios, plus a controlled re-run of the rows that need the emulator). P1 and P2 confirmed fixed; three new findings, two fixed in this push:

- **F1 fixed** (`Classify endpoint, Azure and terraform provisioning failures`): an unreachable `--endpoint-url` is now `NETWORK_ERROR`/RUNTIME and a wrong emulator type behind it `EMULATOR_WRONG_TYPE`/EMULATOR; only argument checks stay `VALIDATION_ERROR`/USAGE.
- **F2 fixed** (same commit): terraform S3-backend provisioning failures carry `IAC_DEPLOY_FAILED`/IAC; `setup azure` and interception failures carry `EMULATOR_NOT_RUNNING`, `INTERNAL_ERROR` or `VALIDATION_ERROR` via the new non-silent `output.WithCode`, so classifying them did not change who renders them.
- **Ratchet** (`Ratchet ErrorEvent literals that set no error code`): `TestEveryNewErrorEventSetsACode` parses `cmd/` and `internal/` and fails when a file gains an `ErrorEvent` literal without a `Code`, or when one is classified without lowering the baseline (40 literals in 18 files today).
- **F3 investigated, not fixed** (pre-existing, unrelated to this PR): Bubble Tea v1.3.10's `tea_init.go` calls `lipgloss.HasDarkBackground()` in a package `init`, which sends the OSC 11 background-colour query and waits up to termenv's 5s `OSCTimeout` — before `main` installs `signal.NotifyContext`. In a terminal that does not answer, every lstk command (even `--version`) stalls 5s, and a Ctrl-C or SIGINT in that window kills lstk with no telemetry. Real terminals answer within milliseconds, so this only bites dumb PTYs and some CI. Options are in the design notes; the clean one is the Bubble Tea v2 migration, which drops that init workaround.

joe4dev:
`Classify the remaining unambiguous ErrorEvent sites`: 33 of the 40 unclassified `ErrorEvent` literals now carry a code (30 converted to `output.Fail`, 3 code-only where the literal is not the returned error; the TUI's generic error display passes through `ErrorCodeOf`). Ratchet baseline is down to 7 literals in 5 files, each a taxonomy judgment call rather than an omission — listed in `tasks.md` 4b.8 pending a decision. Two new integration tests cover `INTEGRATION_NOT_SET_UP` (`lstk az` before `setup azure`) and `AUTH_REQUIRED` (`snapshot show` without a token).

joe4dev:
`Classify the last ErrorEvent sites and widen three code definitions`: the seven judgment-call sites are classified as agreed — cdk/sam version-check failures `DEPENDENCY_MISSING`, terraform provider not installed `IAC_FILE_NOT_FOUND`, invalid or incompatible snapshot file `SNAPSHOT_INVALID_REF`, leftover-container conflicts on start `EMULATOR_START_FAILED` (the code also rides the start error). The three widened definitions are in `docs/structured-output.md`, so the label moved to `docs: needed`. The ratchet baseline is empty: every `ErrorEvent` in the tree sets a `Code`, and a new one without it fails `TestEveryNewErrorEventSetsACode`.

joe4dev:
Final manual round on `97e23b2`: all 40 scenarios match the design, with the real AWS and Azure emulators and the installed aws/terraform/sam/cdk/az CLIs.

| Group | Rows | Result |
|---|---|---|
| Tool ran (proxied, `proxy_exit_code` set, no code) | 1–7, 6a–c, 19, 29–31 | ✓ codes 252/0/254/1/2/7 preserved; extensions not proxied |
| Interruptions | 8, 9, 21 (Ctrl-C at 0.3s, 3s, `q`), 22 (`logs -f` `q` and Ctrl-C), 30 | ✓ `cancelled: true` everywhere, incl. interactive PTY Ctrl-C |
| Tool never ran, classified | 12–17, 17b, 18, 19b, 23, 25, 26, 28c, 28d | ✓ `RUNTIME_UNAVAILABLE`, `EMULATOR_NOT_RUNNING`, `DEPENDENCY_MISSING`, `VALIDATION_ERROR`, `NETWORK_ERROR`, `EMULATOR_WRONG_TYPE`, `INTEGRATION_NOT_SET_UP`, `IAC_DEPLOY_FAILED`, `IAC_FILE_NOT_FOUND`, `INTERNAL_ERROR`, `AUTH_REQUIRED`, `SNAPSHOT_NOT_FOUND` |
| Signal sentinel | 10 (`kill -9` child) | ✓ `-1`, not cancelled, documented |
| Dry run | 19 | phantom `proxy_exit_code: 0`, documented exception |
| No event | 32–35, 37 | ✓ as predicted (pre-existing `PreRunE` holes) |

Two test-harness hazards surfaced while running this, both pre-existing and outside this PR:
- `test/integration/main_test.go`'s keyring helpers fall back to the **real system keychain** when `LSTK_KEYRING=file` is unset and a developer is logged in; `logout_test.go` and the Docker cleanup then overwrite and delete the developer's `lstk` login. `make test-integration` guards it, a bare `go test` does not.
- The Docker tests' `cleanup()` stops whatever `localstack-aws` container is running, including one the developer started.

